PR #13, opened 2019-02-24
Added code instrumentation capability:

pramodk:
- caliper based instrumentation added
  - add command line option to select instrumentation type


ohm314:
Should this be closed for now?

PR #196, opened 2019-05-09
Codegen compatibility

### LINKED ISSUE #187: Code generation compatibility checker
NMODL lexer, parser capabilities are complete but not code generation backends. For example, certain solver methods or NMODL constructs not used are not implemented yet in current code generation backend.

To avoid generating incorrect code or runtime errors for users, we need a codegen compatibility visitor to print "friendly" messages rather than runtime exceptions.

The constructs that are not supported at the moment are:

* SolveBlock supported methods : cnexp, euler, derivimplicit, sparse (otherwise error). Note that empty method is allowed in which case cnexp is used
* Unsupported constructs
  *  TerminalBlock, StepBlock, PartialBlock, MatchBlock, BABlock, ConstantBlock, ConstructorBlock, DestructorBlock, DiscreteBlock, FunctionTableBlock, IndependentBlock
* GLOBAL variables shouldn't have write count > 0 (need to check in symbol table)
* POINTER variable should be converted to BBCOREPOINTER (for CoreNEURON)
* If BBCOREPOINTER variable exist in the AST
  * there should be VERBATIM block in the MOD file where `bbcore_read` and `bbcore_write` methods should have been defined

Some notes:
* This compatibility visitor should be called before codegen backend is invoked
* There should be command line flag in codegen section to force code generation even if above pass returns compatibility as false

iomaganaris:
Added CodegenCompatibilityVisitor to check if there are blocks in the AST that are not compatible with the NMODL code generation and print related error message with the location of the code if there are

- Fixes #187 

Example:
```
[NMODL] [info] :: Processing /home/magkanar/bbp/nmodldb/models/db/modeldb/64229/mod/intf.mod
[NMODL] [info] :: Running symtab visitor
[NMODL] [warning] :: SYMTAB :: t [Argument] in wrecordOLD shadows <extern_neuron_var> definition in NMODL_GLOBAL
[NMODL] [info] :: Running code compatibility checker
[NMODL] [error] :: Code incompatibility detected
[NMODL] [error] :: Cannot translate mod file to .cpp file
[NMODL] [error] :: Fix the following errors and try again
[NMODL] [error] :: Code Incompatibility :: "sop" POINTER found at [91.11-13] should be defined as BBCOREPOINTER to use it in CoreNeuron
[NMODL] [error] :: Code Incompatibility :: "RES" variable found at [93.26-28] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations
[NMODL] [error] :: Code Incompatibility :: "ESIN" variable found at [93.30-33] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations
[NMODL] [error] :: Code Incompatibility :: "Bb" variable found at [93.35-36] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations
[NMODL] [error] :: Code Incompatibility :: "Psk" variable found at [93.38-40] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations
[NMODL] [error] :: Code Incompatibility :: "Gn" variable found at [94.29-30] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations
[NMODL] [error] :: Code Incompatibility :: CONSTRUCTOR construct found at [166.1-180.1] is not supported
[NMODL] [error] :: Code Incompatibility :: DESTRUCTOR construct found at [182.1-187.1] is not supported
```


src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                                       AstNodeType::CONSTRUCTOR_BLOCK,
+                                                       AstNodeType::DESTRUCTOR_BLOCK,
+                                                       AstNodeType::DISCRETE_BLOCK,
+                                                       AstNodeType::FUNCTION_TABLE_BLOCK,
+                                                       AstNodeType::INDEPENDENT_BLOCK,
+                                                       AstNodeType::GLOBAL_VAR,
+                                                       AstNodeType::POINTER_VAR,
+                                                       AstNodeType::BBCORE_POINTER_VAR};
+    incompatible_ast_nodes = AstLookupVisitor().lookup(node, incompatible_ast_types);
+
```
pramodk: I prefer small functions instead of large logic in single function. For example, see below associated comments :

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if (it->is_solve_block()) {
+            auto solve_block = dynamic_cast<ast::SolveBlock*>(it.get());
+            auto method = solve_block->get_method();
+            if (method != nullptr && method->get_node_name() != "cnexp" &&
+                method->get_node_name() != "euler" && method->get_node_name() != "derivimplicit" &&
+                method->get_node_name() != "sparse") {
+                ss << "\"" << method->get_node_name() << "\" solving method used at [";
+                ss << method->get_token()->position() << "] not supported. ";
+                ss << "Supported methods are cnexp, euler, derivimplicit and sparse\n";
+            }
```
pramodk: this could be small function to check if solver method is supported or not.

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                                       AstNodeType::POINTER_VAR,
+                                                       AstNodeType::BBCORE_POINTER_VAR};
+    incompatible_ast_nodes = AstLookupVisitor().lookup(node, incompatible_ast_types);
+
+    std::stringstream ss;
+    for (auto it: incompatible_ast_nodes) {
+        if (it->is_solve_block()) {
+            auto solve_block = dynamic_cast<ast::SolveBlock*>(it.get());
+            auto method = solve_block->get_method();
+            if (method != nullptr && method->get_node_name() != "cnexp" &&
```
pramodk: names line `cnexp` and others should come from `codegen::naming::CNEXP_METHOD`.

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+        } else if (it->is_discrete_block() || it->is_partial_block()) {
+            std::string node_type_name = it->get_node_type_name();
+            // remove "Block" substring
+            node_type_name.erase(node_type_name.end() - 5, node_type_name.end());
+            // transform name of node type to capital letters, as it is in mod files
+            std::transform(node_type_name.begin(),
+                           node_type_name.end(),
+                           node_type_name.begin(),
+                           ::toupper);
+
```
pramodk: Name of the node not "necessarily" same as AST node name. Hence it would be better to have an extra method in the ast node like `get_nmodl_name` to return corresponding NMODL name. This method will be generated by jinja template if there is `nmodl` property exist in nmodl.yaml. 

@ohm314  can help if needed.

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+                   it->is_constructor_block() || it->is_destructor_block() ||
+                   it->is_independent_block()) {
+            std::string node_type_name = it->get_node_type_name();
+            // remove "Block" substring
+            node_type_name.erase(node_type_name.end() - 5, node_type_name.end());
+            // transform name of node type to capital letters, as it is in mod files
+            std::transform(node_type_name.begin(),
+                           node_type_name.end(),
+                           node_type_name.begin(),
+                           ::toupper);
```
pramodk: In general, if something is repeated, it should either go to utils or local lambda function. But again, this should be replaced by get_nmodl_name().

src/parser/nmodl.yy
```diff
@@ -2424,11 +2454,15 @@ global_var_list:   NAME_PTR
...
                         $$ = ast::PointerVarVector();
-                        $$.emplace_back(new ast::PointerVar($1));
+                        auto new_PointerVar = new ast::PointerVar($1);
+                        new_PointerVar->set_token(*($1->get_token()));
+                        $$.emplace_back(new_PointerVar);
                     }
                 |   pointer_var_list "," NAME_PTR
                     {
-                        $1.emplace_back(new ast::PointerVar($3));
+                        auto new_PointerVar = new ast::PointerVar($3);
```
pramodk: minor : new_PointerVar -> variable naming scheme :)

src/parser/nmodl.yy
```diff
@@ -2407,11 +2433,15 @@ range_var_list  :   NAME_PTR
 global_var_list:   NAME_PTR
                     {
                         $$ = ast::GlobalVarVector();
-                        $$.emplace_back(new ast::GlobalVar($1));
+                        auto new_GlobalVar = new ast::GlobalVar($1);
```
pramodk: minor : new_GlobalVar -> variable naming scheme :) same below

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+                }
+            }
+            if (!found_bbcore_read) {
+                ss << "\"bbcore_read\" function not defined in any VERBATIM block\n";
+            }
+            if (!found_bbcore_write) {
+                ss << "\"bbcore_write\" function not defined in any VERBATIM block\n";
+            }
+        }
+    }
```
pramodk: Looking at how we construct error messages, I think it would be cleaner to define error codes now. I was thinking about this for quite some time but didnt get chance to implement it.

Overall idea would be : we have defined error codes with their messages. In the visitor we just collect error code and associated node. And then some `print_error` function take care of printing them in nice format.

In this function, the stream manipulation and other logic gets mixed up.

test/modtoken/modtoken.cpp
```diff
@@ -68,4 +68,28 @@ TEST_CASE("NMODL Lexer returning valid ModToken object", "[token][modtoken]") {
...
+            symbol_type("text", value);
+            ast::Name value1 = value;
+            ss << *(value.get_token());
+            ss << " + ";
+            symbol_type("\ntext", value);
+            ast::Name value2 = value;
+            ss << *(value.get_token());
+            ModToken adder1 = *(value1.get_token());
+            ModToken adder2 = *(value2.get_token());
+            ModToken sum = adder1 + adder2;
```
pramodk: minor : sum = *(value1.get_token()) + *(value2.get_token())

test/modtoken/modtoken.cpp
```diff
@@ -68,4 +68,28 @@ TEST_CASE("NMODL Lexer returning valid ModToken object", "[token][modtoken]") {
...
+            ss << " = " << sum;
+            REQUIRE(ss.str() ==
+                    "           text at [1.1-4] type 357 +            text at [2.1-4] type 357 =   "
+                    "         text at [1.1-2.4] type 357");
+        }
```
pramodk: would be helpful to have test in parser where we parse block like:

```
NEURON {

}
```
and check if NeuronBlock has correct token information.

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+using visitor::AstLookupVisitor;
+
+/**
+ * \details Checks all the ast::AstNodeTypes that are not compatible with NMODL code
+ * generation and prints related messages. If there is some kind of incompatibility
+ * stop NMODL code generation.
+ */
+bool CodegenCompatibilityVisitor::find_incompatible_ast_nodes(Ast* node) {
+    std::vector<AstNodeType> incompatible_ast_types = {AstNodeType::SOLVE_BLOCK,
```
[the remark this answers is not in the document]
iomaganaris: You are right, maybe `unhandled_ast_types` sounds better

src/codegen/codegen_compatibility_visitor.cpp
iomaganaris: I implemented this function using the `ast.hpp` template but I was wondering if all the nodes should have it, as there as some types that don't have `nmodl_name` at all or they have `None` as `nmodl_name`. I think that only the ones that have a useful `nmodl_name` should have them but I don't know if all the nodes should have this method for the sake of uniformity.

src/codegen/codegen_compatibility_visitor.cpp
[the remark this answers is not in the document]
iomaganaris: Good point to move it to the header file, however it needs to be a vector since it is used by `AstLookupVisitor::lookup(Ast* node, std::vector<AstNodeType>& _types)` that takes as second argument a `vector` of `AstNodeType`s

src/codegen/codegen_compatibility_visitor.cpp
ohm314: I don't think they all need it. If it doesn't make sense, it should maybe not be there at all instead of returning placeholder values.

src/codegen/codegen_compatibility_visitor.cpp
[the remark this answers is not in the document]
ohm314: Yes, sorry! my bad. I realized later but forgot to update my comment.

src/language/parser.py
```diff
@@ -30,14 +30,16 @@ def __init__(self, filename, debug=False):
...
+        All classes have the private member of type ModToken token and
+        functions to set and get it, set_token() and get_token()
+        respectively.
+
+        \TODO check why BeforeBlock and AfterBlock do not have token
         """
-        return (name in node_info.LEXER_DATA_TYPES or
-                name in node_info.SYMBOL_BLOCK_TYPES or
-                name in node_info.ADDITIONAL_TOKEN_BLOCKS)
+        return True
```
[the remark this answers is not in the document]
iomaganaris: You are right.I removed it and simply set `has_token` to `True`, where this function was called :+1:

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+            // transform name of node type to capital letters, as it is in mod files
+            std::transform(node_type_name.begin(),
+                           node_type_name.end(),
+                           node_type_name.begin(),
+                           ::toupper);
+
+            ss << "\"" << it->get_node_name() << "\" " << node_type_name << " construct found at [";
+            ss << it->get_token()->position() << "] is not supported\n";
+        } else if (it->is_ba_block()) {
+            ss << "BEFORE/AFTER construct found at [" << it->get_token()->position();
```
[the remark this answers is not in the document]
pramodk: yeah, we definitely need `get_nmodl_name()` or something similar. I think mentioned somewhere in previous comments.

src/language/templates/ast/ast.hpp
```diff
@@ -168,6 +168,23 @@ namespace ast {
...
+        * statement can be returned as a std::string for printing to
+        * text/json form.
+        *
+        * @return name of the statement as a string i.e. "{{ node.nmodl_name }}"
+        *
+        * \sa Ast::get_nmodl_name
+        */
+        std::string get_nmodl_name() {
+            return "{{ node.nmodl_name }}";
+        }
```
pramodk: Ahh I see that you have already implemented it here. Few comments:

* Not every AST node has associated nmodl name. In this case this is going to return empty string (?).
* I would change that behaviour a bit :  add this `get_nmodl_name` method in top level AST class (see ast_common.hpp, similar to `get_node_name`,
* the nodes which has nmodl_name non empty, they only implement this method

This way the code will throw an error if we try to use `get_nmodl_name` for the node where nmodl name property is not exist in yaml specification.

test/modtoken/modtoken.cpp
```diff
@@ -68,4 +71,26 @@ TEST_CASE("NMODL Lexer returning valid ModToken object", "[token][modtoken]") {
...
+            symbol_type("text", value);
+            ast::Name value1 = value;
+            ss << *(value.get_token());
+            ss << " + ";
+            symbol_type("\ntext", value);
+            ast::Name value2 = value;
+            ss << *(value.get_token());
+            ModToken sum = *(value1.get_token()) + *(value2.get_token());
+            ss << " = " << sum;
+            REQUIRE(ss.str() ==
```
pramodk: @iomaganaris : for this unit test, can't we just instantiate ModToken objects and sum them? (i.e. without going through lexer to get tokens)

iomaganaris: You are right. This is possible and now done :+1:

src/language/parser.py
```diff
@@ -162,9 +152,8 @@ def parse_yaml_rules(self, nodelist, base_class=None):
             else:
                 args.base_class = base_class if base_class else 'Ast'
 
-                # check if we need token for the node
-                # todo : we will have token for every node
-                args.has_token = LanguageParser.is_token(class_name)
+                # store token in every node
+                args.has_token = True
```
iomaganaris: Remove has_token from everywhere and simply add token to every node

iomaganaris: It is fine like this at this point, as only the base classes do not get a token member. These classes are not abstract, so they can be instantiated by the user. In the future, a token member can be added to them to be able to be used by the user.

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,142 @@
...
+    incompatible_ast_nodes = AstLookupVisitor().lookup(node, incompatible_ast_types);
+
+    std::stringstream ss;
+    for (auto it: incompatible_ast_nodes) {
+        if (it->is_solve_block()) {
+            auto solve_block = dynamic_cast<ast::SolveBlock*>(it.get());
+            auto method = solve_block->get_method();
+            if (method != nullptr && method->get_node_name() != "cnexp" &&
+                method->get_node_name() != "euler" && method->get_node_name() != "derivimplicit" &&
+                method->get_node_name() != "sparse") {
```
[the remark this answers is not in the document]
pramodk: as it doesnt belong to language specification (but implementation), I would keep that outside.

pramodk:
Adding here as a reminder:

```
/Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/src/codegen/codegen_compatibility_visitor.cpp:127:17: warning: 125 enumeration values not handled in switch: 'NODE', 'STATEMENT', 'EXPRESSION'... [-Wswitch]
        switch (node_type) {
                ^
/Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/src/codegen/codegen_compatibility_visitor.cpp:127:17: note: add missing switch cases
        switch (node_type) {
                ^
```

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,193 @@
...
+            break;
+        case AstNodeType::GLOBAL_VAR: {
+            auto it_global_var = std::dynamic_pointer_cast<ast::GlobalVar>(it);
+            ss << return_error_global_var(node, it_global_var);
+        } break;
+        case AstNodeType::POINTER_VAR: {
+            auto it_pointer_var = std::dynamic_pointer_cast<ast::PointerVar>(it);
+            ss << return_error_pointer(it_pointer_var);
+        } break;
+        case AstNodeType::BBCORE_POINTER_VAR:
```
ohm314: would it not be better to have vectors/sets of AstNodeType the ones that shoudl be printed with name, the ones without and then iterate over those two separately? This big switch case doesn't look very nice in the end and seems a bit too verbose?

src/codegen/codegen_compatibility_visitor.cpp
ohm314: I agree, but maybe we can defer this for later?

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -0,0 +1,193 @@
...
+std::string CodegenCompatibilityVisitor::return_error_if_no_bbcore_read_write(Ast* node) {
+    std::stringstream error_message_no_bbcore_read_write;
+    auto verbatim_nodes = AstLookupVisitor().lookup(node, AstNodeType::VERBATIM);
+    auto found_bbcore_read = false;
+    auto found_bbcore_write = false;
+    for (const auto& it: verbatim_nodes) {
+        auto verbatim = std::dynamic_pointer_cast<ast::Verbatim>(it);
+
+        auto verbatim_statement = verbatim->get_statement();
+        auto verbatim_statement_string = verbatim_statement->get_value();
```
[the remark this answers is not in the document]
iomaganaris: You are right! It is better like this

src/codegen/codegen_compatibility_visitor.cpp
iomaganaris: It's true that it doesn't look very nice but I had a hard time to make it less verbose. I have used templates to call the `return_error` functions to avoid declaring lot's of different functions that do the same thing, so for every different type of `node` there must be a switch case that calls the corresponding function with the correct template argument. What could be done is divide the `AstNodeType`s to those that should be printed with name and without as you suggested and then divide also the large switch case in smaller parts or functions. Would this be enough? I also tried to not overcomplicate this part, so that in the future it might be easier to add more cases if needed in the same place.

ohm314: (discussed offline a possible solution)

iomaganaris:
please retest

iomaganaris:
please retest

PR #210, opened 2019-05-28
Fix mofidication of const ion variables

### LINKED ISSUE #209: C code generation: attempted modification of const ion variables when copies are not being optimized
In the case where `CodegenCVisitor::optimize_ion_variable_copies()` returns `false`, I get a compilation error when trying to compile coreneuron.
Here is an example with Ca_HVA2:
```
error #137: expression must be a modifiable lvalue
                  inst->eca[id] = inst->ion_eca[indexes[0*pnodecount+id]];
```

The error arises because ` inst->eca` was marked `const`, which is obviously incorrect. 

PS I'll submit a PR in a second to fix this.

sharkovsky:
Fixes #209.

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -729,9 +729,6 @@ bool CodegenCVisitor::is_constant_variable(std::string name) {
     auto symbol = program_symtab->lookup_in_scope(name);
     bool is_constant = false;
     if (symbol != nullptr) {
-        if (symbol->has_any_property(NmodlType::read_ion_var) && optimize_ion_variable_copies()) {
-            is_constant = true;
-        }
         if (symbol->has_any_property(NmodlType::param_assign) && symbol->get_write_count() == 0) {
             is_constant = true;
         }
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -729,7 +729,7 @@ bool CodegenCVisitor::is_constant_variable(std::string name) {
     auto symbol = program_symtab->lookup_in_scope(name);
     bool is_constant = false;
     if (symbol != nullptr) {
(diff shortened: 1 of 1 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -729,7 +729,7 @@ bool CodegenCVisitor::is_constant_variable(std::string name) {
     auto symbol = program_symtab->lookup_in_scope(name);
     bool is_constant = false;
     if (symbol != nullptr) {
-        if (symbol->has_any_property(NmodlType::read_ion_var)) {
+        if (symbol->has_any_property(NmodlType::read_ion_var) && optimize_ion_variable_copies()) {
```
pramodk: @sharkovsky : `read_ion_var` here is an internal copy of the mechanism. So this can't be const anyway because we have to make copy. So I would just remove this if block.

ohm314: Actually i discussed this with @sharkovsky and he has a point. In his case he did explicitely want `optimize_ion_variable_copies` to be `false` and so the check makes sense. The rationale is that in some derived classes this could be set to `false` too.

sharkovsky: @pramodk if I understand correctly, you're saying that _regardless_ of whether `optimize_ion_variable_copies` returns true or false, we should not mark the symbol constant when `symbol->has_any_property(NmodlType::read_ion_var)` evaluates to true. 
So you're saying the we should remove the `if` statement altogether?

Question: if we have to make copies anyway, why did the error arise only when I set `optimize_ion_variable_copies` to false?

pramodk: > Question: if we have to make copies anyway, why did the error arise only when I set optimize_ion_variable_copies to false?

as discussed offline, when `optimize_ion_variable_copies` is true, those variables are not used and hence error is not generated. Making variables const was bug in the first place.

PR #221, opened 2019-06-17
Added a nmodl dockerfile

ohm314:
The generated image provides a Jupyter notebook environment with nmodl setup.
Added a brief subsection on testing NMODL from docker into the README.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+FROM alpine:latest AS builder
+
+WORKDIR /nmodl/src
+
+RUN apk add --update build-base gcc g++ make cmake flex flex-dev bison git python3-dev
+
+RUN pip3 install --trusted-host pypi.python.org jinja2 pyyaml pytest sympy
+
+
+RUN git clone --recursive https://github.com/BlueBrain/nmodl.git && \
+    cd nmodl && \
+    git checkout 0.2
+
+WORKDIR /nmodl/src/nmodl
+
+RUN pip3 install .
+
+
+FROM alpine:latest
+
+
+RUN apk add --no-cache --update python3 libgfortran libstdc++ openblas && \
+    apk add --no-cache --virtual build-dependencies \
+            build-base linux-headers openblas-dev freetype-dev \
+            pkgconfig gfortran python3-dev && \
+    pip3 install --no-cache-dir --trusted-host pypi.python.org \
+                 jinja2 pyyaml pytest sympy numpy matplotlib jupyter && \
+    apk del build-dependencies && \
+    rm -rf /var/cache/apk/*
+    
+WORKDIR /usr/lib/python3.6/site-packages
+
+COPY --from=builder /usr/lib/python3.6/site-packages/nmodl .
+
+EXPOSE 8888
+WORKDIR /nmodl
+
+CMD ["jupyter", "notebook", "--port=8888", "--no-browser", "--ip=0.0.0.0", "--allow-root"]
+
+
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ fully functional Jupyter notebook environment.
 After installing docker you can pull and run the NMODL image from your terminal:
 
 ```sh
-docker run -it -p 8888:8888 -v ~/yourpath/to/nmodl/notebooks:/nmodl/notebooks ohm314/nmodl:test
-
+wget "https://raw.githubusercontent.com/BlueBrain/nmodl/master/docker/docker-compose.yml"
+DUID=$(id -u) DGID=$(id -g) HOSTNAME=$(hostname) docker-compose up
 ```
 
 If all goes well you should see at the end status messages similar to these:
@@ -85,6 +85,9 @@ If all goes well you should see at the end status messages similar to these:
 
 Based on the example above you should then open your browser and navigate to the URL http://127.0.0.1:8888/?token=a7902983bad430a1193542945f6a64ae303f559343fbb0dc
 
+You can open and run all example notebooks provided in the `examples` folder. You can also create
+new notebooks in `my_notebooks`, which will be stored in a subfolder `notebooks` at your current
+working directory.
 
 ### Using the Python API
 
--- a/README.md
+++ b/README.md
@@ -54,19 +54,19 @@ NET_RECEIVE(weight (uS)) {
 
 See [INSTALL.md](https://github.com/BlueBrain/nmodl/blob/master/INSTALL.md) for detailed instructions to build the NMODL from source.
 
-#### Taking NMODL for a spin
+### Try NMODL with Docker
 
 To quickly test the NMODL Framework's analysis capabilities we provide a
 [docker](https://www.docker.com) image, which includes the NMODL Framework python library and a
-fully functional Jupyter notebook environment.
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -54,6 +54,38 @@ NET_RECEIVE(weight (uS)) {
...
+### Taking NMODL for a spin
+
+To quickly test the NMODL Framework's analysis capabilities we provide a
+[docker](https://www.docker.com) image, which includes the NMODL Framework python library and a
+fully functional Jupyter notebook environment.
+
+After installing docker you can pull and run the NMODL image from your terminal:
+
+```sh
+docker run -it -p 8888:8888 -v ~/yourpath/to/nmodl/notebooks:/nmodl/notebooks ohm314/nmodl:test
```
tristan0x: With the current command, if the user creates a Jupyter notebook from the container, the file will belong to *root* on the container.
You can quickly review the 2nd tech-talk I made about Docker.
https://github.com/CNS-OIST/STEPS_Docker/ provide a possible solution for this issue.

Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM alpine:latest AS builder
```
tristan0x: I would use a fixed version of alpine instead of `latest` to prevent possible breaks. For instance, when `latest` will use Python 3.7, `/usr/lib/python3.6/` used in the recipe won't exist.

README.md
ohm314: thanks, I improved that a bit :)

Dockerfile
ohm314: Yeah, was thinking whether I should fix the version or not. I think you're right, it's better to fix. Done along with a couple of other improvements. Please let me know what you think about it now, when you have time.

README.md
```diff
@@ -54,6 +54,41 @@ NET_RECEIVE(weight (uS)) {
 
 See [INSTALL.md](https://github.com/BlueBrain/nmodl/blob/master/INSTALL.md) for detailed instructions to build the NMODL from source.
 
+### Taking NMODL for a spin
+
+To quickly test the NMODL Framework's analysis capabilities we provide a
+[docker](https://www.docker.com) image, which includes the NMODL Framework python library and a
+fully functional Jupyter notebook environment.
+
+After installing docker you can pull and run the NMODL image from your terminal:
```
tristan0x:
```suggestion
After installing docker and docker-compose you can pull and run the NMODL image from your terminal:
```

docker/recipe/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+ADD entrypoint /usr/bin/
+ENTRYPOINT ["/usr/bin/entrypoint"]
+
+EXPOSE 8888
+WORKDIR /nmodl/notebooks
+
+COPY --from=builder /nmodl/src/nmodl/docs/notebooks ./examples
+
+CMD ["jupyter", "notebook", "--port=8888", "--no-browser", "--ip=0.0.0.0", "--allow-root", "--debug", "--notebook-dir=/nmodl/notebooks"]
```
tristan0x: I don't think option "--debug" is required.

docker/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    hostname: ${HOSTNAME}
+    ports:
+    - "8888:8888"
+    environment:
+    - USER_LOGIN=${USER}
+    - USER_ID=${DUID}
+    - GROUP_ID=${DGID}
+    volumes:
+    - $PWD/notebooks:/nmodl/notebooks/my_notebooks
+    command:
```
tristan0x: The `command` argument is not required as already specified in the docker image with the `CMD` directive.

docker/recipe/entrypoint
```diff
@@ -0,0 +1,41 @@
...
+
+# Run the given command as root if bash or sh,
+# the fake user otherwise.
+case "$1" in
+    sh|bash)
+        exec $@
+        ;;
+    *)
+        cd /nmodl/notebooks
+        gosu dummy "$@"
```
tristan0x: *gosu* utility will be missing and the container will crash if it is started with bash as command, for instance:

docker run osh314/nmodl bash

I suggest to install gosu in the Docker image, like it is done in STEPS_Docker:
https://github.com/CNS-OIST/STEPS_Docker/blob/master/recipe/Dockerfile#L17-L22

ohm314: Hu, but I did add the ‘gosu’ package in Dockerfile:23 or should this be done differently?

docker/docker-compose.yml
ohm314: I was wondering about that. But maybe I’ll leave it for clarity. The end-user can easily see what is run without having to look into the Dockerfile

docker/recipe/Dockerfile
ohm314: You’re right

docker/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: '3.7'
+services:
+  notebook:
+    image: ohm314/nmodl:0.2
```
pramodk: could you push this under bluebrain?

docker/recipe/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+WORKDIR /nmodl/src
+
+RUN apk add --update build-base gcc g++ make cmake flex flex-dev bison git python3-dev
+
+RUN pip3 install --trusted-host pypi.python.org jinja2 pyyaml pytest sympy
+
+
+RUN git clone --recursive https://github.com/BlueBrain/nmodl.git && \
+    cd nmodl && \
+    git checkout 0.2
```
pramodk: If someone rebuilt from Dockerfile, I would prefer to build master. what you think?

ohm314: yes, probably the right way would be to build from master. Once I understand how the automatic building on dockerhub works, I will fix this accordingly

docker/docker-compose.yml
ohm314: I don't know how

docker/recipe/Dockerfile
pramodk: I mean, why can't we remove `checkout 0.2` from here?

pramodk:
@ohm314 : note that I am push forced to re-build on master!

docker/recipe/Dockerfile
tristan0x: I suggest to make the version configurable at `docker build` time.

In Dockerfile:
```
ARG NMODL_VERSION=master
[...]
git checkout ${NMODL_VERSION}
```

Then you can specify another value when building the image: `docker build --build-arg NMODL_VERSION=0.2`

pramodk: that's also good idea. I will update that.

pramodk:
I made following changes:
* pushed images to bluebrain/nmodl on docker hub
* added latest tag along with 0.2
* removed build step from docker-compose.yml (otherwise it was failing)
* added *share* directory so that standalone image can be use interactively
* Dockerfile builds master by default and version can be specified from command line
* updated README for the same

@tristan0x : I haven't used `gosu` before. Would be ok if that requested change is done later?

ohm314:
Thanks @pramodk - very cool! Unless I'm missing something, the package is already being installed (see Dockerfile) and used in the entry_point script.

pramodk:
Ok. I don't what `gosu` is, so I might have missed it. @tristan0x : could you approve?

PR #223, opened 2019-06-21
Fixes issue with Eigen:Matrix named X clashing with code variable X

### LINKED ISSUE #217: invalid types ‘double[int]’ for array subscript
@nrnhines reported the following compilation error with [`release.mod`](https://github.com/neuronsimulator/nrn/blob/master/share/demo/release/release.mod) and other files from [`neurondemo`](https://github.com/neuronsimulator/nrn/tree/master/share/demo/release) :
```
release.cpp:292:26: error: invalid types ‘double[int]’ for array subscript
             inst->X[id][0] = inst->Ves[id];
```
A file with all the patches made for the `neurondemo` can be found attached:
[neurondemo-mod.zip](https://github.com/BlueBrain/nmodl/files/3281302/neurondemo-mod.zip)

iomaganaris:
- Changed name of "X" Eigen::Matrix to "X_operator" in sympy_solver_visitor
- Made needed changes to filter the name of the X variable coming from sympy
  solution to X_operator to match the codegen_c_visitor
- Fixes #217

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -1627,11 +1627,11 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(ast::EigenNewtonSolverBloc
     printer->add_newline();
     auto float_type = default_float_data_type();
     int N = node->get_n_state_vars()->get_value();
-    printer->add_line("Eigen::Matrix<{}, {}, 1> X;"_format(float_type, N));
+    printer->add_line("Eigen::Matrix<{}, {}, 1> X_operator;"_format(float_type, N));
 
     print_statement_block(node->get_setup_x_block().get(), false, false);
 
-    // functor that evaluates F(X) and J(X) for Newton solver
+    // functor that evaluates F(X_operator) and J(X_operator) for Newton solver
     printer->start_block("struct functor");
     printer->add_line("NrnThread* nt;");
     printer->add_line("{0}* inst;"_format(instance_struct()));
@@ -1655,7 +1655,7 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(ast::EigenNewtonSolverBloc
 
     printer->add_indent();
     printer->add_text(
-        "void operator()(const Eigen::Matrix<{0}, {1}, 1>& X, Eigen::Matrix<{0}, {1}, "
+        "void operator()(const Eigen::Matrix<{0}, {1}, 1>& X_operator, Eigen::Matrix<{0}, {1}, "
         "1>& F, "
         "Eigen::Matrix<{0}, {1}, {1}>& Jm) const"_format(float_type, N));
     printer->start_block();
@@ -1663,7 +1663,7 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(ast::EigenNewtonSolverBloc
     print_statement_block(node->get_functor_block().get(), false, false);
     printer->end_block(2);
 
-    // assign newton solver results in matrix X to state vars
+    // assign newton solver results in matrix X_operator to state vars
     printer->start_block("void finalize()");
     print_statement_block(node->get_finalize_block().get(), false, false);
     printer->end_block(1);
@@ -1672,13 +1672,14 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(ast::EigenNewtonSolverBloc
     printer->add_text(";");
     printer->add_newline();
 
-    // call newton solver with functor and X matrix that contains state vars
+    // call newton solver with functor and X_operator matrix that contains state vars
     printer->add_line("// call newton solver");
     printer->add_line("functor newton_functor(nt, inst, id, pnodecount, v, indexes);");
     printer->add_line("newton_functor.initialize();");
-    printer->add_line("int newton_iterations = nmodl::newton::newton_solver(X, newton_functor);");
+    printer->add_line(
+        "int newton_iterations = nmodl::newton::newton_solver(X_operator, newton_functor);");
 
-    // assign newton solver results in matrix X to state vars
+    // assign newton solver results in matrix X_operator to state vars
     print_statement_block(node->get_update_states_block().get(), false, false);
     printer->add_line("newton_functor.finalize();");
 }
@@ -1687,7 +1688,7 @@ void CodegenCVisitor::visit_eigen_linear_solver_block(ast::EigenLinearSolverBloc
     printer->add_newline();
     std::string float_type = default_float_data_type();
     int N = node->get_n_state_vars()->get_value();
-    printer->add_line("Eigen::Matrix<{0}, {1}, 1> X, F;"_format(float_type, N));
+    printer->add_line("Eigen::Matrix<{0}, {1}, 1> X_operator, F;"_format(float_type, N));
     printer->add_line("Eigen::Matrix<{0}, {1}, {1}> Jm;"_format(float_type, N));
     printer->add_line("{}* J = Jm.data();"_format(float_type));
     print_statement_block(node->get_variable_block().get(), false, false);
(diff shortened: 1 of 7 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -1627,11 +1627,11 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(ast::EigenNewtonSolverBloc
     printer->add_newline();
     auto float_type = default_float_data_type();
     int N = node->get_n_state_vars()->get_value();
-    printer->add_line("Eigen::Matrix<{}, {}, 1> X;"_format(float_type, N));
+    printer->add_line("Eigen::Matrix<{}, {}, 1> X_operator;"_format(float_type, N));
```
ohm314: I know I'm nitpicking here.. but why did you decide to call it "X_operator"? it's not really an operator, is it?

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1696,7 +1697,7 @@ void CodegenCVisitor::visit_eigen_linear_solver_block(ast::EigenLinearSolverBloc
 
     printer->add_newline();
     printer->add_line(
-        "X = Eigen::PartialPivLU<Eigen::Ref<Eigen::Matrix<{0}, {1}, {1}>>>(Jm).solve(F);"_format(
+        "X_operator = Eigen::PartialPivLU<Eigen::Ref<Eigen::Matrix<{0}, {1}, {1}>>>(Jm).solve(F);"_format(
```
ohm314: also, that doesn't really fix the problem, does it? What if "X_operator" is already in use? should you not make it dynamic by choosing a variable name that is not found in the symbol table? I would try "X", if found, then do "X_[four random chars].

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -186,26 +208,28 @@ void SympySolverVisitor::construct_eigen_solver_block(
...
+                                                                          setup_x_operator_block,
                                                                           update_state_block,
                                                                           finalize_block);
         /// replace statement block with solver block as it contains all statements
         ast::StatementVector solver_block_statements{
             std::make_shared<ast::ExpressionStatement>(solver_block)};
         block_with_expression_statements->set_statements(std::move(solver_block_statements));
     } else {
+        /// filter solutions for variable named "X" and change it to "X_operator"
+        auto solutions_filtered = filter_X(solutions);
```
ohm314: somehow this also doesn't do what the issue proposes, no? You always replace X by X_operator, rather than finding an alternative only when X is taken...

iomaganaris:
> This works in principle, but could be made robust.
> I think the renaming mechanism could be improved.

You are right, it can be improved a lot 👍 This was just a test to check what needs fixing and to get some feedback on that. I am going to work on that when I come back

pramodk:
@iomaganaris : I am rebasing and push forcing this to solve travis build issue.

pramodk:
@iomaganaris : after looking at the changes, I wonder if it would have been simple to just use variable names with `_`. This is because there won't be any variable names with `_` (NMODL language doesn't allow). 

I don't remember very well our last conversation. Lets sync tomorrow.

ohm314:
Ah, @pramodk - well that is indeed much simpler, but neither @iomaganaris nor I were aware of this little detail...

src/utils/common_utils.hpp
```diff
@@ -51,6 +54,48 @@ bool make_path(const std::string& path);
...
+  public:
+    SingletonRandomString(SingletonRandomString const&) = delete;
+    SingletonRandomString& operator=(SingletonRandomString const&) = delete;
+
+    static std::shared_ptr<SingletonRandomString> instance() {
+        static std::shared_ptr<SingletonRandomString> s{new SingletonRandomString};
+        return s;
+    }
+
+    bool random_string_exists(const std::string& var_name) {
```
ohm314:
```suggestion
    bool random_string_exists(const std::string& var_name) const {
```

src/utils/common_utils.hpp
```diff
@@ -51,6 +54,48 @@ bool make_path(const std::string& path);
...
+        return (random_strings.find(var_name) != random_strings.end());
+    }
+
+    std::string get_random_string(const std::string& var_name) {
```
ohm314:
```suggestion
    std::string get_random_string(const std::string& var_name) const {
```

src/utils/common_utils.hpp
ohm314: Also, to avoid having to later do doxygen campaigns, let's as a rule briefly document all public interfaces at least.

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -129,10 +129,63 @@ static bool is_local_statement(std::shared_ptr<ast::Statement> statement) {
...
+        context.replace(foundHere, from.size(), to);
+        lookHere = foundHere + to.size();
+    }
+    return context;
+}
+
+std::vector<std::string> SympySolverVisitor::filter_string_vector(
+    const std::vector<std::string>& original_vector,
+    const std::string& original_string,
+    const std::string& substitution_string) {
```
ohm314:
```suggestion
    const std::string& substitution_string) const {
```

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -129,10 +129,63 @@ static bool is_local_statement(std::shared_ptr<ast::Statement> statement) {
...
+    std::vector<std::string> filtered_vector;
+    for (auto element: original_vector) {
+        std::string filtered_element = replaceAll(element, original_string, substitution_string);
+        filtered_vector.push_back(filtered_element);
+    }
+    return filtered_vector;
+}
+
+std::string SympySolverVisitor::suffix_random_string(const std::string& original_string) {
```
ohm314:
```suggestion
std::string SympySolverVisitor::suffix_random_string(const std::string& original_string) const {
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1621,17 +1621,36 @@ void CodegenCVisitor::print_function(ast::FunctionBlock* node) {
     codegen = false;
 }
 
+std::string CodegenCVisitor::find_var_unique_name(const std::string& original_name) {
```
ohm314:
```suggestion
std::string CodegenCVisitor::find_var_unique_name(const std::string& original_name) const {
```

src/utils/common_utils.cpp
```diff
@@ -55,5 +56,20 @@ bool make_path(const std::string& path) {
     }
 }
 
+std::string generate_random_string(const int len) {
+    std::string s;
```
ohm314:
```suggestion
    std::string s(len, 0);
```

src/utils/common_utils.cpp
```diff
@@ -55,5 +56,20 @@ bool make_path(const std::string& path) {
...
+    static const char alphanum[] =
+        "0123456789"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz";
+    std::random_device dev;
+    std::mt19937 rng(dev());
+    std::uniform_int_distribution<std::mt19937::result_type> dist(0, (sizeof(alphanum) - 1));
+    for (int i = 0; i < len; ++i) {
+        s.push_back(alphanum[dist(rng)]);
```
ohm314:
```suggestion
        s[i] = alphanum[dist(rng)];
```
just to avoid those `push_back`s

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -129,10 +129,63 @@ static bool is_local_statement(std::shared_ptr<ast::Statement> statement) {
     return false;
 }
 
+std::string& SympySolverVisitor::replaceAll(std::string& context,
+                                            const std::string& from,
+                                            const std::string& to) {
```
ohm314:
```suggestion
                                            const std::string& to) const {
```

iomaganaris:
> @iomaganaris : after looking at the changes, I wonder if it would have been simple to just use variable names with `_`. This is because there won't be any variable names with `_` (NMODL language doesn't allow).
> 
> I don't remember very well our last conversation. Lets sync tomorrow.

As @ohm314 said I didn't know that in any part of the variable name the underscore is not supported. In the beginning we said to prefix an `_` which was not supported in NMODL language but when trying this there were some weird compilation issues and I didn't know that underscore wasn't supported in any part of the name. :/

src/utils/common_utils.hpp
iomaganaris: You are right :+1: I will add documentation in all public interfaces from now on

src/utils/common_utils.cpp
iomaganaris: Yeah, good catch :+1:

PR #230, opened 2019-07-10
Change GLOBAL to RANGE variables if they are written

### LINKED ISSUE #218: GLOBAL variables need to become RANGE if the are written
@nrnhines reported issue with `GLOBAL` variables that need to be defined as `RANGE` to enable the codegen backend transformation, if they are written somewhere in the mod file

iomaganaris:
- Added visitor that checks if a GLOBAL variable is written and turns it into RANGE
- Visit ast with GlobalToRangeVisitor after the last symtab creation in main and after running PerfVisitor to calculate the write_count
- Modified SymTab visitor, so that when symtab is updated and the visitor encounters a global_var, it checks if the property of  the variable is range_var in the symtab and it keeps it that way if it is
- Fixes #218


src/nmodl/main.cpp
```diff
@@ -329,6 +330,17 @@ int main(int argc, const char* argv[]) {
             UnitsVisitor(units_dir).visit_program(ast.get());
         }
 
+        /// GLOBAL to RANGE rename visitor
+        {
+            // make sure to run perf visitor because code generator
+            // looks for read/write counts const/non-const declaration
+            PerfVisitor().visit_program(ast.get());
```
ohm314: I see that now the `PerfVisitor` is being run 3 to 4 times.. do we really have to do this?

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * so that they can be handled by Codegen as RANGE variables
+ */
+
+namespace nmodl {
+namespace visitor {
+
+using symtab::syminfo::NmodlType;
+
+void GlobalToRangeVisitor::visit_global_var(ast::GlobalVar* node) {
+    if (ast->get_symbol_table()->lookup(node->get_node_name())->get_write_count() > 0) {
```
ohm314: factor out `node->get_node_name()` into a local var, if for nothing else, for readability...

src/nmodl/main.cpp
iomaganaris: Well, it was needed before the `CodegenCompatibilityVisitor` to check the symtab write count for `GLOBAL` variables and decide if there must be generated an error about turning them to `RANGE`. If we want to delete this case from the `CodegenCompatibilityVisitor` we can remove this `PerfVisitor` pass. 
Then it is needed for sure before the `GlobalToRangeVisitor`. 
I am not sure if the last call before the code generation can be removed due to the `SymtabVisitor(update_symtab).visit_program(ast.get());` calls that are happening before it. Based on the tests I've made, the last pass of the `PerfVisitor` can be removed and still get the same code generated.

src/visitors/global_var_visitor.cpp
iomaganaris: Good point :+1:

src/nmodl/main.cpp
ohm314: @pramodk can you way in on that? Do we need all those calls to that visitor?

pramodk: > I see that now the PerfVisitor is being run 3 to 4 times.. do we really have to do this?

Let's address first question : when + why PerfVisitor needs to be run?
-> When perfvisitor runs, it update all operations count associated with symbols. If particular pass is dependent on such ops-count, perfvisitor must be run. In the current scenario, by default, PerfVisitor is run twice : codegen compatibility and code generation. 

Why we need to run it multiple times?
-> Passes like Inlining and localiser change the AST and hence symbol table needs to be updated with associated ops-count. And hence they need to be re-run.

Actually, PerfVisitor should be run by the associated pass but lets do that re-factorign in separate ticket.

iomaganaris: @pramodk Okay. So the `PerfVisitor` pass before the codegen should stay. 
Another question I have is if the Global to Range error in the `CodegenCompatibilityVisitor` should stay, since you told me that there will be some inconsistency between `Neuron` and `CoreNeuron` data with this transformation.

pramodk: > Another question I have is if the Global to Range error in the CodegenCompatibilityVisitor should stay, since you told me that there will be some inconsistency between Neuron and CoreNeuron data with this transformation.

yes, that's true.

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,35 @@
...
+        ast->get_symbol_table()
+            ->lookup(node->get_node_name())
+            ->remove_property(NmodlType::global_var);
+        ast->get_symbol_table()->lookup(node->get_node_name())->add_property(NmodlType::range_var);
```
pramodk: I think you two days ago you asked half question : `can we change node type...`. Is it related to here? i.e. Can we change the type of node in the AST?

Answer is yes. In above case you have merely changed the property of node in the AST (underlying node is still GLOBAL). 

What we have to do is, we have to remove variables from GLOBAL node and add them to RANGE node. (if RANGE node doesn't exist we have to create one).

iomaganaris:
> Wow! couldn't believe that this was not fixed! :|

Same for me 😛  I saw some super nasty errors in the CI due to this.
If (hopefully) everything in the CI passes, I think that this is ready to merge after reviewing it

iomaganaris:
The PR has (finally) passed the CI and it's ready for review 🙂

src/visitors/global_var_visitor.hpp
```diff
@@ -0,0 +1,70 @@
...
+  private:
+    /// ast::Ast* node
+    std::shared_ptr<ast::Program> ast;
+
+  public:
+    /// \name Ctor & dtor
+    /// \{
+
+    /// Default UnitsVisitor constructor
+    GlobalToRangeVisitor() = default;
```
pramodk, line 50: do we need this? Or do you wanted to delete this?

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * \file
+ * \brief AST Visitor to change ast::GlobalVar to ast::Range
+ * variables in nmodl::symtab to nmodl::symtab::syminfo::NmodlType::range_var
+ * so that they can be handled by Codegen as RANGE variables
+ */
+
+namespace nmodl {
+namespace visitor {
+
+void GlobalToRangeVisitor::visit_statement_block(ast::StatementBlock& node) {
```
pramodk, line 26: typically there are lots of statement blocks and global variables are only defined in neuron block. So I am wondering why we need to visit all statement blocks? Visiting neuron block will suffice ?

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,60 @@
...
+    ast::RangeVarVector range_variables;
+    std::vector<std::string> global_vars_to_remove;
+    for (const auto& statement: statements) {
+        if (statement->is_global()) {
+            const auto& global_variables =
+                std::static_pointer_cast<ast::Global>(statement)->get_variables();
+            global_vars_to_remove.clear();
+            for (const auto& global_variable: global_variables) {
+                auto variable_name = global_variable->get_node_name();
+                if (ast->get_symbol_table()->lookup(variable_name)->get_write_count() > 0) {
```
pramodk, line 37: you should take symbol table in some variable outside the for loop.

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+namespace nmodl {
+namespace visitor {
+
+void GlobalToRangeVisitor::visit_statement_block(ast::StatementBlock& node) {
+    const auto& statements = node.get_statements();
+    ast::RangeVarVector range_variables;
+    std::vector<std::string> global_vars_to_remove;
+    for (const auto& statement: statements) {
```
pramodk, line 30: `const` is good but there are some unnecessary ones here? `const vector` provides access to elements as if they were const.

by the way, all getters now return `const &`.

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,60 @@
...
+        if (statement->is_global()) {
+            const auto& global_variables =
+                std::static_pointer_cast<ast::Global>(statement)->get_variables();
+            global_vars_to_remove.clear();
+            for (const auto& global_variable: global_variables) {
```
pramodk, line 35: same here

src/codegen/codegen_compatibility_visitor.hpp
```diff
@@ -66,6 +66,14 @@ class CodegenCompatibilityVisitor: public visitor::AstVisitor {
     /// Default CodegenCompatibilityVisitor constructor
     CodegenCompatibilityVisitor() = default;
 
+    /// CodegenCompatibilityVisitor constructor that removes the error for
+    /// the global variables if the global-to-range pass is enabled
+    explicit CodegenCompatibilityVisitor(bool enable_global_to_range) {
+        if (enable_global_to_range) {
+            unhandled_ast_types_func.erase(AstNodeType::GLOBAL_VAR);
+        }
+    }
```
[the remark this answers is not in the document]
iomaganaris: I thought that removing the error would be okay, since even if there is there is this error in the original mod file, then with the global to range pass the variables will be translated, resulting in a correct mod file representation. Otherwise, if we keep the error in every case, then to generate the code when there is this issue we would have to pass the `codegen --force` flag, which would generate the code even if there are more errors present. Do you have a better idea on how to handle this?

src/language/code_generator.cmake
```diff
@@ -43,6 +43,22 @@ set(CODE_GENERATOR_YAML_FILES
...
+    ${PROJECT_BINARY_DIR}/src/visitors/json_visitor.cpp
+    ${PROJECT_BINARY_DIR}/src/visitors/json_visitor.hpp
+    ${PROJECT_BINARY_DIR}/src/visitors/lookup_visitor.cpp
+    ${PROJECT_BINARY_DIR}/src/visitors/lookup_visitor.hpp
+    ${PROJECT_BINARY_DIR}/src/visitors/nmodl_visitor.cpp
+    ${PROJECT_BINARY_DIR}/src/visitors/nmodl_visitor.hpp
+    ${PROJECT_BINARY_DIR}/src/visitors/symtab_visitor.cpp
+    ${PROJECT_BINARY_DIR}/src/visitors/symtab_visitor.hpp
+    ${PROJECT_BINARY_DIR}/src/visitors/visitor.hpp
+)
```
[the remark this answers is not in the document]
iomaganaris: It was moved when I run `make clang-format` (using `clang-format 9` from `bb5`). I have no idea why it did it

src/nmodl/main.cpp
```diff
@@ -307,7 +314,8 @@ int main(int argc, const char* argv[]) {
             // run perfvisitor to update read/wrie counts
             PerfVisitor().visit_program(*ast);
             // If there is an incompatible construct and code generation is not forced exit NMODL
-            if (CodegenCompatibilityVisitor().find_unhandled_ast_nodes(*ast) && !force_codegen) {
+            if (CodegenCompatibilityVisitor(nmodl_global_to_range).find_unhandled_ast_nodes(*ast) &&
+                !force_codegen) {
```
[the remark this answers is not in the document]
iomaganaris: You're right 👍  It would be even better probably if we first run the `GlobalToRange` visitor and then pass a flag to enable the error, given the logic I described in above comment

src/visitors/global_var_visitor.hpp
```diff
@@ -0,0 +1,70 @@
...
+ * \file
+ * \brief \copybrief nmodl::visitor::GlobalToRangeVisitor
+ */
+
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "ast/ast.hpp"
```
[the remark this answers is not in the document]
iomaganaris: I merged with `master` but forgot to update this file since there was to complain about it during compilation. I will fix it accordingly

test/visitor/global_to_range.cpp
```diff
@@ -0,0 +1,96 @@
...
+            PROCEDURE rates(v(mV)) {
+            UNITSOFF
+                mtau = mtau+1
+                minf = minf+1
+                htau = htau+1
+                ntau = ntau+1
+                ninf = ninf+1
+            }
+            UNITSON
+        )";
```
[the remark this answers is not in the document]
iomaganaris: You're right, we could use only 2 variables. One which is going to become `RANGE` and one that stays `GLOBAL`

src/codegen/codegen_compatibility_visitor.hpp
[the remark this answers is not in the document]
pramodk: Correct me if I am missing something :
* global to range transformation pass will remove problematic GLOBAL variables
* so even if we have this check in codegen_compatibility_visitor.hpp, it won't or shouldn't raise an error. Similarly, we won't need to pass `--force` for codegen because offending ones are already removed.
* my intention of keeping this check is, for some reasons if global-to-range is skipped, codegen compatibility pass will raise the error.

test/visitor/global_to_range.cpp
```diff
@@ -0,0 +1,96 @@
...
+        std::string output_range = "RANGE minf, ninf, mtau, htau, ntau";
+        std::string output_global = "GLOBAL hinf";
+        THEN("GLOBAL variables that are written are turned to RANGE") {
+            auto result = run_global_to_var_visitor(input_nmodl);
+            REQUIRE(result.find(output_range) != std::string::npos);
+            REQUIRE(result.find(output_global) != std::string::npos);
+        }
```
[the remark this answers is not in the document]
iomaganaris: Probably you're right. In the beginning the test checked this but I altered it to check if the `RANGE` variables are there in the generated mod file, since I thought that this encapsulates everything else. I will also add the `symtab` check to be sure

src/visitors/global_var_visitor.hpp
iomaganaris: Actually this is not needed, so it can be removed

src/visitors/global_var_visitor.cpp
iomaganaris: You're right. Visiting only the `NeuronBlock` should be enough. I will fix this

src/visitors/global_var_visitor.cpp
iomaganaris: Maybe there are some unnecessary ones. I will check again to make sure 👍

src/codegen/codegen_compatibility_visitor.hpp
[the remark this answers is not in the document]
iomaganaris: You're completely right. We should just run this pass before the compatibility checker

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "visitors/global_var_visitor.hpp"
+
+namespace nmodl {
+namespace visitor {
+
+void GlobalToRangeVisitor::visit_neuron_block(ast::NeuronBlock& node) {
+    ast::RangeVarVector range_variables;
+    std::vector<std::string> global_variables_to_remove;
+
+    auto& statement_block = node.get_statement_block();
```
ohm314, line 25: I'm wondering why you are doing everything ont he level of the neuron_block, rather than using the visitor to descend into the tree and do the work at the nodes themselves? It looks like you're not really using the visitor pattern..

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,65 @@
...
+            /// remove offending global variables
+            for (auto& global_var_to_remove: global_variables_to_remove) {
+                const auto& global_var_iter =
+                    find_if(global_variables.begin(),
+                            global_variables.end(),
+                            [&](const std::shared_ptr<ast::GlobalVar>& var) {
+                                return var->get_node_name() == global_var_to_remove;
+                            });
+                std::static_pointer_cast<ast::Global>(statement)->erase_global_var(global_var_iter);
+            }
```
ohm314, line 53:
```suggestion
            for (auto& global_var_to_remove: global_variables_to_remove) {
                std::remove_if(global_variables.begin(), global_variables.end(),
                    [&](const std::shared_ptr<ast::GlobalVar>& var) {
                                return var->get_node_name() == global_var_to_remove;
                            });
            }
```
Wouldn't this be simpler and cleaner?

src/nmodl/main.cpp
```diff
@@ -301,6 +308,19 @@ int main(int argc, const char* argv[]) {
             SymtabVisitor(update_symtab).visit_program(*ast);
         }
 
+        /// GLOBAL to RANGE rename visitor
+        if (nmodl_global_to_range) {
+            // make sure to run perf visitor because code generator
+            // looks for read/write counts const/non-const declaration
+            PerfVisitor().visit_program(*ast);
```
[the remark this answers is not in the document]
pramodk: (I can answer this as I briefly discussed this in the past with Ioannis).
If the AST has been changed (e.g. previous optimisation pass like Inlining, localiser), we want to get new/correct count of `ops` and hence need to re-run this visitor.

.travis.yml
```diff
@@ -78,6 +78,7 @@ before_install:
         export PATH=/usr/local/opt/flex/bin:/usr/local/opt/bison/bin:$PATH;
         brew unlink python@2;
         brew link --overwrite python;
+        export SDKROOT="$(xcrun --show-sdk-path)";
```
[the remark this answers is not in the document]
iomaganaris: This is needed to avoid getting errors with `stdio.h` and `math.h` not found when we build `coreneuron` in `macos`

src/visitors/global_var_visitor.cpp
iomaganaris: It's true that it would be nicer in this case to directly remove the variable from the `globar_variables` but the `nmodl` API doesn't support this. This is something I also discussed with @cattabiani and he pointed me to the `erase_<node>` function to do this task, which is better to avoid deleting stuff without updating the parent

src/visitors/global_var_visitor.cpp
iomaganaris: I am visiting the `NeuronBlock` instead of the `Global` or `GlobalVar` nodes since some of them would need to be deleted and this would become more complicated in this case, needing probably another pass on the AST the way I think about it. We can discuss maybe if you have a better idea on how this could be done

ohm314: I'm not insisting to change - we can keep it "for now" but I think that with the recently introduced `parent` pointers, this should be easier.

src/visitors/global_var_visitor.cpp
ohm314: Since you're simply deleting the global vars, do you need to update the parent pointer? or am I missing something?

pramodk: `parent pointer` is up-link. When you remove the node itself, you don't need to update? (cc: @cattabiani )

src/visitors/global_var_visitor.cpp
iomaganaris: Given your comment I started thinking this a bit more and added also the deletion of the `Global` statements if they are empty here https://github.com/BlueBrain/nmodl/pull/230/commits/5715a027f8ded426f7eb8e2d92e656ff3d2ce451. I believe this is the only issue with visiting the `Global` statements instead of the `NeuronBlock`, which is something that cannot be done easily I believe..

src/visitors/global_var_visitor.cpp
iomaganaris: I checked again and `erase_global_var` simply erases the variable from the `GlobalVarVector`, however in `std::static_pointer_cast<ast::Global>(statement)->get_variables()` `get_variables()` is a `const`, so I believe that it's not possible to delete any variable outside `erase_global_var`

cattabiani: So, about erase there is the routine to erase stuff: erase_*. It was difficult to block people from operating directly on the children (adding a child requires an update of the parent pointer) without preventing them to also erase the node. I agree that is erase is safe. However, I do not see an easy way to allow direct access to the node ONLY for deletion.

cattabiani: I do not really understand your point @iomaganaris . The idea is to do:
- A = get_variables()
- loop over A filling a container with the stuff that you want to erase
- call erase_global_var to erase them all at once

PS: get_variables returns a const& so it should be always up to date

azure-pipelines.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       export PATH=/usr/local/opt/flex/bin:/usr/local/opt/bison/bin:$PATH;
       mkdir -p $(Build.Repository.LocalPath)/build
       cd $(Build.Repository.LocalPath)/build
-      cmake .. -DPYTHON_EXECUTABLE=$(which python3) -DCMAKE_INSTALL_PREFIX=$HOME/nmodl -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      cmake .. -DPYTHON_EXECUTABLE=$(which python3) -DCMAKE_INSTALL_PREFIX=$HOME/nmodl -Dnmodl_PYTHONPATH=$HOME/nmodl -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
cattabiani, line 54: why `nmodl_PYTHONPATH` does not have nmodl with capital letters?

src/visitors/global_var_visitor.cpp
cattabiani: > I'm wondering why you are doing everything ont he level of the neuron_block, rather than using the visitor to descend into the tree and do the work at the nodes themselves? It looks like you're not really using the visitor pattern..

to me it does not make much sense to do the job in the global nodes. It is difficult to know if you need to be deleted or not. It is your parent that has a clear view from up there and can decide who should live and who should convert. I  think @iomaganaris 's solution is fine

cattabiani: > I'm not insisting to change - we can keep it "for now" but I think that with the recently introduced `parent` pointers, this should be easier.

even with parent is a lot of back and forth: you go down to the global node, we come back to delete myself, then, where do I go? I see a lot of pitfalls in deleting nodes when we are visiting them

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,77 @@
...
+    ast::RangeVarVector range_variables;
+    std::vector<std::string> global_variables_to_remove;
+    std::vector<ast::StatementVector::const_iterator> global_statements_to_remove;
+
+    auto& statement_block = node.get_statement_block();
+    auto& statements = (*statement_block).get_statements();
+    const auto& symbol_table = ast->get_symbol_table();
+
+    for (auto statements_it = statements.begin(); statements_it < statements.end();
+         statements_it++) {
```
cattabiani, line 31: it could be better to do ++statements_it

[doc](http://www.cs.technion.ac.il/users/yechiel/c++-faq/increment-pre-post-speed.html)

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,77 @@
...
+void GlobalToRangeVisitor::visit_neuron_block(ast::NeuronBlock& node) {
+    ast::RangeVarVector range_variables;
+    std::vector<std::string> global_variables_to_remove;
+    std::vector<ast::StatementVector::const_iterator> global_statements_to_remove;
+
+    auto& statement_block = node.get_statement_block();
+    auto& statements = (*statement_block).get_statements();
+    const auto& symbol_table = ast->get_symbol_table();
+
+    for (auto statements_it = statements.begin(); statements_it < statements.end();
```
cattabiani, line 30: I think this is a copy. Isn't it better const auto&

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,77 @@
...
+namespace nmodl {
+namespace visitor {
+
+void GlobalToRangeVisitor::visit_neuron_block(ast::NeuronBlock& node) {
+    ast::RangeVarVector range_variables;
+    std::vector<std::string> global_variables_to_remove;
+    std::vector<ast::StatementVector::const_iterator> global_statements_to_remove;
+
+    auto& statement_block = node.get_statement_block();
+    auto& statements = (*statement_block).get_statements();
```
cattabiani, line 27: I think statements is const (and hidden in auto). It would be better to explicitly state that (like const auto&)

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,77 @@
...
+    const auto& symbol_table = ast->get_symbol_table();
+
+    for (auto statements_it = statements.begin(); statements_it < statements.end();
+         statements_it++) {
+        auto& statement = *statements_it;
```
cattabiani, line 32: const auto& as before

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,77 @@
...
+        /// only process global statements
+        if (statement->is_global()) {
+            const auto& global_variables =
+                std::static_pointer_cast<ast::Global>(statement)->get_variables();
+            global_variables_to_remove.clear();
+            for (auto& global_variable: global_variables) {
```
cattabiani, line 38: const auto& as before

src/visitors/global_var_visitor.cpp
```diff
@@ -0,0 +1,77 @@
...
+                if (symbol_table->lookup(variable_name)->get_write_count() > 0) {
+                    range_variables.emplace_back(new ast::RangeVar(global_variable->get_name()));
+                    global_variables_to_remove.emplace_back(variable_name);
+                }
+            }
+
+            /// remove offending global variables
+            for (auto& global_var_to_remove: global_variables_to_remove) {
+                const auto& global_var_iter =
+                    find_if(global_variables.begin(),
```
cattabiani, line 50: uh uh. Cool function that I did not know (find_if)

src/visitors/global_var_visitor.cpp
cattabiani: Ok, sorry but I read things in the wrong order. @ohm314 's solution is good but it cannot work here because stuff is const &. However...

You could take a look to:

https://github.com/BlueBrain/nmodl/blob/7c2c6903a829489148b414a5121c06dc4c338e06/src/visitors/visitor_utils.cpp#L126 

it does exactly the thing you want to do for statements. You can easily adapt it to globalvar. Even better, we could create a template that does this job.

src/visitors/global_var_visitor.cpp
pramodk: Just to be fair with Ioannis :smiley: I asked him to remove some [extra ones here](https://github.com/BlueBrain/nmodl/pull/230#discussion_r429069086).

azure-pipelines.yml
iomaganaris: This comes from [here](https://github.com/BlueBrain/CoreNeuron/blob/6da0f49114c79d26b422ff897ed936af235e4811/CMake/packages/Findnmodl.cmake#L36) and this is something we have to fix in `CoreNEURON` if the path has changed I believe

src/visitors/global_var_visitor.cpp
iomaganaris: Apparently I am an `old line C` programmer 😛 
```
Old line C programmers are used to writing i++ instead of ++i.
```
I will change this though from now on, even if the gains if any are very small 👍

src/visitors/global_var_visitor.cpp
pramodk: Based on above conversation, I am marking this as resolve then.

ohm314: After reconsidering this, I agree that for deletion it's better to delete from above, it does make sense.

PR #231, opened 2019-07-10
Fix issue with installing extra resources

ohm314:
We were using a python configuration file generated via CMake to find the extra resources, such as MOD file examples and javascript visualization utilities. That method worked well enough when using the CMake installation method. Unfortunately this failed when using python distutils (`setup.py`). Especially when installing with `pip`. The python package ended up in a broken state (The config does not copied to the final installation folder and even worse, contains wrong paths).

This patch replaces this mechanism with the proper file placement and installation via the `MANIFEST.in` file. Furthermore we use now distutils `pkg_resources`. Which is the recommended, portable way to access extra resources.


ohm314:
Never mind.. I completely dropped the ball here... see #232 instead.

PR #232, opened 2019-07-10
Fix issue with installing extra resources (take 2)

ohm314:
We were using a python configuration file generated via CMake to find the extra resources, such as MOD file examples and javascript visualization utilities. That method worked well enough when using the CMake installation method. Unfortunately this failed when using python distutils (setup.py). Especially when installing with pip. The python package ended up in a broken state (The config does not copied to the final installation folder and even worse, contains wrong paths).

This patch replaces this mechanism with the proper file placement and installation via the MANIFEST.in file. Furthermore we use now distutils pkg_resources. Which is the recommended, portable way to access extra resources.


README.md
```diff
@@ -108,11 +108,11 @@ Once the NMODL Framework is installed, you can use the Python parsing API to loa
...
 ```python
 from nmodl import dsl
-import os
 
-expsyn = os.path.join(dsl.example_dir(), "expsyn.mod")
+examples = dsl.list_examples() 
+expsyn = dsl.load_example(examples[-1])
 driver = dsl.NmodlDriver()
-modast = driver.parse_file(expsyn)
+modast = driver.parse_string(expsyn)
```
pramodk: do we need `load_example`? that term sound bit heavy to me. I would expect `list_examples` to return full path of models (mod files) and then just use `parse_file`?

nmodl/dsl.py
```diff
@@ -1,21 +1,38 @@
...
-    if os.path.exists(installed_example):
-        return installed_example
+    if resource_exists(__name__, RESOURCE_DIR) and resource_isdir(__name__, RESOURCE_DIR):
+        return resource_listdir(__name__, RESOURCE_DIR)
     else:
-        return os.path.join(PROJECT_SOURCE_DIR, "share", "example")
+        raise FileNotFoundError("Could nto find sample directory")
+
+
+def load_example(example):
```
pramodk: as mentioned previous, if we provide full path, it would be sufficient.

README.md
ohm314: The user should not rely on a full path / absolute location of package resources because they could come packaged from a zip file. `pkg_resource` allows dealing with exactly those cases.

pramodk: Agree...but the full path is returned by the package after installation. Am I missing something?

ohm314: Depends. Usually it's in the filesystem and yes there is a function to get/construct the full path. But this method is more portable. Do you see any downside to directly providing the content of the modfile?

nmodl/ast.py
```diff
@@ -22,16 +22,16 @@ def view(nmodl_ast):
...
 
-    installed_viz_tool = os.path.join(PROJECT_INSTALL_DIR, "share", "viz")
-    if os.path.exists(installed_viz_tool):
-        viz_tool_dir = installed_viz_tool
+    resource = "ext/viz"
+    if resource_exists(__name__, resource) and resource_isdir(__name__, resource):
+        installed_viz_tool = resource_filename(__name__, resource)
     else:
-        viz_tool_dir = os.path.join(PROJECT_SOURCE_DIR, "share", "viz")
+        raise FileNotFoundError("Could nto find sample mod files")
```
iomaganaris: Minor misspelling
```suggestion
raise FileNotFoundError("Could not find sample mod files")
```

nmodl/dsl.py
```diff
@@ -1,21 +1,38 @@
...
     """
-    import os
-    installed_example = os.path.join(PROJECT_INSTALL_DIR, "share", "example")
-    if os.path.exists(installed_example):
-        return installed_example
+    if resource_exists(__name__, RESOURCE_DIR) and resource_isdir(__name__, RESOURCE_DIR):
+        return resource_listdir(__name__, RESOURCE_DIR)
     else:
-        return os.path.join(PROJECT_SOURCE_DIR, "share", "example")
+        raise FileNotFoundError("Could nto find sample directory")
```
iomaganaris: Same as above
```suggestion
raise FileNotFoundError("Could not find sample mod files")
```

nmodl/dsl.py
```diff
@@ -1,21 +1,38 @@
...
+    Args:
+        example: Filename of an example as provided by `list_examples()`
+    Returns:
+        List of available examples
+    """
+    resource =  osp.join(RESOURCE_DIR, example)
+    if resource_exists(__name__, resource):
+        return resource_string(__name__, resource)
+    else:
+        raise FileNotFoundError("Could nto find sample mod files")
```
iomaganaris: Same as above
```suggestion
raise FileNotFoundError("Could not find sample mod files")

File: README.md (hunk 1 above)
Comment (pramodk) [reply README.md]: If user have mod file locally and want to try it, `parse_file` example shown in README is straightforward instead of going via load_examples (I thought ;) )

It's ok, this is minor thing. I thought we could avoid that extra function if not needed.

File: README.md [hunk 6]
```diff
@@ -108,11 +108,11 @@ Once the NMODL Framework is installed, you can use the Python parsing API to loa
 
 ```python
 from nmodl import dsl
-import os
 
-expsyn = os.path.join(dsl.example_dir(), "expsyn.mod")
+examples = dsl.list_examples() 
+expsyn = dsl.load_example(examples[-1])
```
pramodk:
```suggestion
nmodl_string = dsl.load_example(examples[-1])
```

README.md
pramodk:
```suggestion
modast = driver.parse_string(nmodl_string)
```

PR #233, opened 2019-07-10
Changed --force_codegen flag to --force subcommand of codegen

iomaganaris:
- Fixes issue that the --force_codegen flag could not be used with CoreNeuron


st4rl3ss:
LGTM

PR #243, opened 2019-09-24
Replace raw pointers by references

ohm314:
This is a first attempt to make the use of pointers and references more consistent and clean. For now we replace some raw pointer access with references, especially when working with the visitors.
In a future PR we should address the many unnecessary shared_ptr and replace them with unique_ptr and consider accessing elements only by reference when the receiver does not participate in the object's lifetime.


ohm314:
please retest

ohm314:
please retest

PR #248, opened 2019-10-29
Use the newly introduced function in coreneuron to set net_receive

ohm314:
This is necessary because we want to for now keep generated code free of
the more complex changes of data encapsulation in coreneuron.

This PR follows
https://github.com/BlueBrain/CoreNeuron/pull/203

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -2656,7 +2656,7 @@ void CodegenCVisitor::print_mechanism_register() {
             method_name("net_buf_receive")));
     }
     if (info.num_net_receive_parameters != 0) {
-        auto net_recv_init_arg = "NULL";
+        auto net_recv_init_arg = "nullptr";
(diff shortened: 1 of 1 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -2656,11 +2656,13 @@ void CodegenCVisitor::print_mechanism_register() {
             method_name("net_buf_receive")));
     }
     if (info.num_net_receive_parameters != 0) {
-        printer->add_line("pnt_receive[mech_type] = {};"_format(method_name("net_receive")));
-        printer->add_line("pnt_receive_size[mech_type] = num_net_receive_args();");
+        auto net_recv_init_arg = "NULL";
```
iomaganaris: From my understanding `net_recv_init_arg/pnt_receive_init` is a pointer to a function name so it could also be like that?
```suggestion
        auto net_recv_init_arg = "nullptr";
```

PR #252, opened 2020-01-14
Changes tracking refactoring in coreneuron

ohm314:
Mostly updating include paths as headers are moved around/changed in coreneuron.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,33 +47,34 @@ find_package(BISON 3.0 REQUIRED)
 # =============================================================================
 # HPC Coding Conventions
 # =============================================================================
-set(NMODL_ClangFormat_EXCLUDES_RE ".*/ext/.*$$"
-    CACHE STRING "list of regular expressions to exclude C/C++ files from formatting"
-    FORCE)
-set(NMODL_CMakeFormat_EXCLUDES_RE ".*/ext/.*$$"
-    CACHE STRING "list of regular expressions to exclude CMake files from formatting"
-    FORCE)
-set(NMODL_ClangFormat_DEPENDENCIES pyastgen parser-gen
-    CACHE STRING "list of CMake targets to build before formatting C++ code"
-    FORCE)
+set(NMODL_ClangFormat_EXCLUDES_RE
+    ".*/ext/.*$$"
+    CACHE STRING "list of regular expressions to exclude C/C++ files from formatting" FORCE)
+set(NMODL_CMakeFormat_EXCLUDES_RE
+    ".*/ext/.*$$"
+    CACHE STRING "list of regular expressions to exclude CMake files from formatting" FORCE)
+set(NMODL_ClangFormat_DEPENDENCIES
+    pyastgen parser-gen
+    CACHE STRING "list of CMake targets to build before formatting C++ code" FORCE)
 add_subdirectory(cmake/hpc-coding-conventions/cpp)
 
 # =============================================================================
 # Format & execute ipynb notebooks in place (pip install nbconvert clean-ipynb)
 # =============================================================================
-add_custom_target(nb-format
-                  jupyter
-                  nbconvert
-                  --to
-                  notebook
-                  --execute
-                  --inplace
-                  --ExecutePreprocessor.timeout=360
-                  "${CMAKE_SOURCE_DIR}/docs/notebooks/*.ipynb"
-                  &&
-                  clean_ipynb
-                  --keep-output
-                  "${CMAKE_SOURCE_DIR}/docs/notebooks/*.ipynb")
+add_custom_target(
+  nb-format
+  jupyter
+  nbconvert
+  --to
+  notebook
+  --execute
+  --inplace
+  --ExecutePreprocessor.timeout=360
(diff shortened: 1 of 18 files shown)


src/utils/string_utils.hpp
```diff
@@ -34,14 +36,22 @@ enum class text_alignment { left, right, center };
 /// Trim from start
 static inline std::string& ltrim(std::string& s) {
     s.erase(s.begin(),
-            std::find_if(s.begin(), s.end(), std::not1(std::ptr_fun<int, int>(std::isspace))));
+            std::find_if(s.begin(), s.end(), std::not1(
+                std::function<int(int)>(
```
mgeplf: I think using a lambda instead of `std::function`, as it's easier to read, and has less overhead (and I think the cast isn't required?)

mgeplf: Couldn't stop myself, had to look up what the overhead is.

Top is w/ std::function, bottom is w/ lambda (note - my impl. might be wrong, I haven't tested it).
https://godbolt.org/z/EpF2tH

I doubt for this application that overhead is that important (although, I suppose you marked it as `inline`, so maybe it is :)), but the difference surprised me.  I'd even guess that b/c the function is so long the inliner would't follow the suggestion.

ohm314: Thanks @mgeplf :)
Indeed the lambda translates much nicer.
Actually I had just stumbled over this code because `std::ptr_fun` is depricated since C++14 and removed in C++17 and since I had changed compiler flags I started getting those ugly warnings, so thought to fix this up. Without thinking too much I replaced it by `std::function` and was astonished that the compiler was not able to resolve the correct `std::isspace` without the nasty static cast. So... either way, I prefer the lambda implementation.

mgeplf: Cool.  The nagging feeling I had when I saw `std::function` probably has a direct lineage to `std::ptr_fun`, now that I think about it :)

CMakeLists.txt
```diff
@@ -99,10 +100,8 @@ find_python_module(sympy 1.2 REQUIRED)
 find_python_module(textwrap 0.9 REQUIRED)
 find_python_module(yaml 3.12 REQUIRED)
 
-include_directories(${NMODL_PROJECT_SOURCE_DIR}
-                    ${NMODL_PROJECT_SOURCE_DIR}/src
-                    ${PROJECT_BINARY_DIR}/src
-                    ${NMODL_PROJECT_SOURCE_DIR}/ext)
+include_directories(${NMODL_PROJECT_SOURCE_DIR} ${NMODL_PROJECT_SOURCE_DIR}/src
+                    ${PROJECT_BINARY_DIR}/src ${NMODL_PROJECT_SOURCE_DIR}/ext)
```
pramodk: @tristan0x  : It's not relevant here but just to mention that the previous formatting was looking better (at least to me).

src/lexer/CMakeLists.txt
```diff
@@ -62,134 +62,122 @@ set(LEXER_SOURCE_FILES
 # =============================================================================
 # Directories for parsers (as they need to be in separate directories)
 # =============================================================================
-file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/src/parser/nmodl
-                    ${PROJECT_BINARY_DIR}/src/parser/diffeq
-                    ${PROJECT_BINARY_DIR}/src/parser/c
-                    ${PROJECT_BINARY_DIR}/src/parser/unit)
+file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/src/parser/nmodl ${PROJECT_BINARY_DIR}/src/parser/diffeq
+     ${PROJECT_BINARY_DIR}/src/parser/c ${PROJECT_BINARY_DIR}/src/parser/unit)
```
pramodk: same here

src/printer/CMakeLists.txt
```diff
@@ -2,12 +2,9 @@
...
 set(PRINTER_SOURCE_FILES
-    ${CMAKE_CURRENT_SOURCE_DIR}/json_printer.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/json_printer.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/nmodl_printer.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/nmodl_printer.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/code_printer.hpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/code_printer.cpp)
+    ${CMAKE_CURRENT_SOURCE_DIR}/json_printer.hpp ${CMAKE_CURRENT_SOURCE_DIR}/json_printer.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/nmodl_printer.hpp ${CMAKE_CURRENT_SOURCE_DIR}/nmodl_printer.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/code_printer.hpp ${CMAKE_CURRENT_SOURCE_DIR}/code_printer.cpp)
```
pramodk: this as well, my eyes are hurting now ;)

PR #253, opened 2020-01-24
Remove unused variables

alkino:


src/codegen/codegen_c_visitor.cpp
```diff
@@ -2584,7 +2583,7 @@ void CodegenCVisitor::print_mechanism_register() {
     auto args = register_mechanism_arguments();
     auto nobjects = num_thread_objects();
     if (info.point_process) {
-        printer->add_line("point_register_mech({}, NULL, NULL, {});"_format(args, nobjects));
+        printer->add_line("point_register_mech({}, {});"_format(args, nobjects));
     } else {
```
pramodk: I have few comments about the change in API:
* If we look into CoreNEURON source code [here](https://github.com/BlueBrain/CoreNeuron/blob/439cfab870ce01638109cfc1f1a2f5e92f4b0ed5/coreneuron/nrnoc/register_mech.cpp#L375), we don't see usage of these parameters.
* But if we see the origin of corresponding CoreNEURON code in [NEURON here](https://github.com/neuronsimulator/nrn/blob/1dae4848fc14a4b84175a8192d96b4a4360e8cdc/src/nrnoc/init.c#L739), we can see how these params were used. Specifically:

> printer->add_line("point_register_mech({}, {});"_format(args, nobjects));

These two params were Constructor/Destructor functions created form NMODL file. As mod file object is created in NEURON, these functions are not explicitly called again by CoreNEURON. But in future, we might use these functions. 

And hence I am thinking keep current API / unused parameters. (same for nrnpointerindex).

PR #254, opened 2020-01-24
Fixing bad submodule pulls in #252

ohm314:
In PR #252 I mistakenly did a `pull --recurse-submodules` and commited
the updated submodule into this repo pulling in changes from various
submodules. Thiw as not intended and a mistake. Undoing it here. One
small adjustment was to checkout a commit from hpc-coding-conventions
that is one before the cmake-format version change but still much newer
than the previously checked out version.


ohm314:
please retest

PR #255, opened 2020-01-31
CLI11: bump to v1.9.0

alkino:


pramodk:
Note that BBP Jenkins job was not starting because [we have list of usernames](https://bbpcode.epfl.ch/ci/view/hpc/job/hpc.nmodl/configure) that can only trigger job. @alkino  was missing in the list.

image

PR #265, opened 2020-03-05
Fix link errors with fmt for non-Debug build type

### LINKED ISSUE #264: Error while building non-Debug build type : Undefined symbols for architecture x86_64 int fmt::v5::internal::char_traits<char>::format_float<d
If we remove below from CMakeLists.txt

```
set(CMAKE_BUILD_TYPE Debug)
```

then we get:

```
[ 97%] Linking CXX executable ../../bin/nmodl_visitor
Undefined symbols for architecture x86_64:
  "int fmt::v5::internal::char_traits<char>::format_float<double>(char*, unsigned long, char const*, int, double)", referenced from:
      void fmt::v5::basic_writer<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >::write_double<double>(double, fmt::v5::basic_format_specs<char> const&) in libvisitor.a(kinetic_block_visitor.cpp.o)
      void fmt::v5::basic_writer<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >::write_double<double>(double, fmt::v5::basic_format_specs<char> const&) in libsymtab.a(symbol.cpp.o)
  "int fmt::v5::internal::char_traits<char>::format_float<long double>(char*, unsigned long, char const*, int, long double)", referenced from:
      void fmt::v5::basic_writer<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >::write_double<long double>(long double, fmt::v5::basic_format_specs<char> const&) in libvisitor.a(kinetic_block_visitor.cpp.o)
      void fmt::v5::basic_writer<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >::write_double<long double>(long double, fmt::v5::basic_format_specs<char> const&) in libsymtab.a(symbol.cpp.o)
  "fmt::v5::internal::error_handler::on_error(char const*)", referenced from:
      void fmt::v5::internal::parse_format_string<false, char, fmt::v5::format_handler<fmt::v5::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >, char, fmt::v5::basic_format_context<std::__1::back_insert_iterator<fmt::v5::internal::basic_buffer<char> >, char> >&>(fmt::v5::basic_string_view<char>, fmt::v5::format_handler<fmt::v5::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >, char, fmt::v5::basic_format_context<std::__1::back_insert_iterator<fmt::v5::internal::basic_buffer<char> >, char> >&&&) in libvisitor.a(kinetic_block_visitor.cpp.o)
      void fmt::v5::internal::parse_format_string<false, char, fmt::v5::format_handler<fmt::v5::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >, char, fmt::v5::basic_format_context<std::__1::back_insert_iterator<fmt::v5::internal::basic_buffer<char> >, char> >&>(fmt::v5::basic_string_view<char>, fmt::v5::format_handler<fmt::v5::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >, char, fmt::v5::basic_format_context<std::__1::back_insert_iterator<fmt::v5::internal::basic_buffer<char> >, char> >&&&)::writer::operator()(char const*, char const*) in libvisitor.a(kinetic_block_visitor.cpp.o)
      char const* fmt::v5::internal::parse_arg_id<char, fmt::v5::internal::id_adapter<fmt::v5::format_handler<fmt::v5::arg_formatter<fmt::v5::back_insert_range<fmt::v5::internal::basic_buffer<char> > >, char, fmt::v5::basic_format_context<std::__1::back_insert_iterator<fmt::v5::internal::basic_buffer<char> >, char> >&, char> >(char const*, char const*, ../../../s/c/codegen_c_visitor.hpp
```

pramodk:
- spdlog also ships fmt which is older version
  - use newer fmt consistently through utils/logger.hpp
  - no need to include fmt/format.h directly

fixes #264  #265 #266 #267 #268


mgeplf:
> no need to include fmt/format.h directly

Is this the best way to solve this?  It breaks the '[include what you use](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md)' rule of thumb.  To me it seems weird that I need to include `utils/logger.hpp` (say in `src/codegen/codegen_c_visitor.hpp`), when that file doesn't use any logging capabilities.

pramodk:
> Is this the best way to solve this? It breaks the 'include what you use' rule of thumb. 

@mgeplf : you are right!

* Currently we have incompatibility issue between use of fmt  and spdlog together.
* We need a newer version of fmt (and spdlog bring older version)
* If we include only fmt then include order matters (which is also bad!) and result into link errors.
* This issue we didn't see before because with `Debug` builds everything has worked until today.

I will check if we can resolve this by updating spdlog to newer version (where they have updated [fmt](https://github.com/gabime/spdlog/tree/v1.x/include/spdlog/fmt)) and if we can remove fmt altogether.

mgeplf:
I don't know anything about `spdlog`, but I note there is a cmake option called `SPDLOG_FMT_EXTERNAL`, perhaps that's the way to go?  Then you can choose your own `fmt` and not subject to upstream.

src/codegen/codegen_acc_visitor.cpp
```diff
@@ -5,8 +5,9 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
+#include "fmt/format.h"
```
ohm314: since these are external dependencies, can we not include via `<>` and "-System" to avoid the compiler stumbling over warnings that don't matter to us?

pramodk: yeah agree. I used the "existing" style -  which we should change that consistently.
As there are multiple external submodules included like this, I created separate issue #269 where we can change that everywhere.

pramodk:
> I don't know anything about spdlog, but I note there is a cmake option called SPDLOG_FMT_EXTERNAL, perhaps that's the way to go?

Currently spdlog repo "copies" fmt source instead of using submodule. So I was not sure about the compatibility aspects.

I switched to submodules approach for fmt as well as spdlog and they are working now correctly. Please take a look.

CMakeLists.txt
```diff
@@ -101,7 +101,9 @@ find_python_module(yaml 3.12 REQUIRED)
 include_directories(${NMODL_PROJECT_SOURCE_DIR}
                     ${NMODL_PROJECT_SOURCE_DIR}/src
                     ${PROJECT_BINARY_DIR}/src
-                    ${NMODL_PROJECT_SOURCE_DIR}/ext)
+                    ${NMODL_PROJECT_SOURCE_DIR}/ext
+                    ${NMODL_PROJECT_SOURCE_DIR}/ext/fmt/include
```
mgeplf: We use `add_subdirectory(fmt)` for our `fmt` submodule - I wonder if that's easier than the old `include_directories` way.

pramodk: but for include to work, that is needed. Isn't it?

.gitmodules
```diff
@@ -10,3 +10,9 @@
 [submodule "ext/eigen"]
 	path = ext/eigen
 	url = https://github.com/eigenteam/eigen-git-mirror.git
+[submodule "ext/spdlog"]
+	path = ext/spdlog
+	url = https://github.com/gabime/spdlog.git
```
ohm314: wait but now we're not pointing to the proper spdlog, are we?

pramodk: that's official repo if I am not mistaken

CMakeLists.txt
mgeplf: `fmt` has a bunch of stuff (https://github.com/fmtlib/fmt/blob/master/CMakeLists.txt#L10) that does the heavy lifting, so then you don't have to use the old cmake style include_directories (which is nice, because then you can control warnings and such.

pramodk: Hmm.. I have `add_subdirectory(ext/fmt EXCLUDE_FROM_ALL)` to exclude it from `make/make install` and then I am getting:

```
/gpfs/bbp.cscs.ch/project/proj16/kumbhar/pramod_scratch/tmp/nmodl/src/codegen/codegen_acc_visitor.cpp:8:10: fatal error: fmt/format.h: No such file or directory
 #include "fmt/format.h"
          ^~~~~~~~~~~~~~
```

So for now I will continue using include path.

pramodk: By the way, I will go ahead with the merge as we need this sooner. But if you have more suggestions, please post the review when time permits and I will accommodate that.

pramodk:
\> 5 hours for Travis CI build queue. I think it's time to make Travis CI optional and make Azure CI mandatory.

mgeplf:
>> I don't know anything about spdlog, but I note there is a cmake option called SPDLOG_FMT_EXTERNAL, perhaps that's the way to go? 

> Currently spdlog repo "copies" fmt source instead of using submodule. So I was not sure about the compatibility aspects.

Submodules and using the internal code are orthogonal.   It looks like `SPDLOG_FMT_EXTERNAL` allows you to control whether the internal one is used, which is probably what you want to do - disable the internal one, and only use a single instance of `fmt` in the project.

CMakeLists.txt
mgeplf: The include path is basically like a global variable, with all the problems that brings.  The problem you're having relates to not specifying the dependency for `codegen`.

I have an example PR on how to clean this up: https://github.com/BlueBrain/nmodl/pull/273

I don't know your policies well enough to shepherd it through, so I will leave it up to you on how to handle it - ie: you're welcome to take it over.

PR #270, opened 2020-03-07
Fix incompatibility with Bison >= 3.5

### LINKED ISSUE #251: nmodl build breaks with bison 3.5
See for example here:
https://dev.azure.com/pramodskumbhar/nmodl/_build/results?buildId=369&view=logs&j=2d7e015f-1feb-5038-1a42-96d7c1621f70&t=d9fddc9b-622a-595b-cdf0-89e9d11e6688

The latest azure pipeline runs pull on macos bison 3.5 through macbrew. I noticed that this causes the build to break for example at: `src/parser/diffeq_context.cpp:80` . We need to investigate the breaking changes in bison and update our code accordingly.

pramodk:
- bison 3.5 released symbol_type::token () removed
  - we use c++ parsers with variant semantic values and
    symbol_type::token was used for unit testing as well
    as to decide if next token is EoF.
  - now we have to use symbol_type.type_get() and by_type(yytokentype)
    to compare the tokens
  - all comparisons with switch cases need to be converted
    to if-else because of non-constexpr by_type() function
    call (bison without C++17)
  - see bison bug report for details :
    https://lists.gnu.org/archive/html/bug-bison/2020-01/msg00001.html
  - remove older bison installation for CI

fixes #251

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,9 +46,10 @@ addons:
   # for Mac builds, we use Homebrew
   homebrew:
     packages:
-      - flex
+      - bison
       - boost
       - cmake
+      - flex
       - python@3
     update: true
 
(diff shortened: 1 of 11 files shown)


pramodk:
* Yes, lets discuss this on Monday.
* I looked into bit details of this and here is summary:
  * if we look at "real" need for use of `by_type(token).type_get())`, we only use that for determining end of input (`by_type(token::END).type_get()`) while scanning some "custom" type token. For example, methods like `get_expr_for_nonlinear`, `scan_string`. This is max at 2-3 places so not big deal.
  * And rest of the 99% use of this is for:
      * testing : as Wolfgang Thaller mentioned that in bug report, the real use of that function was to check token types for unit testing (under test/*). If we are not writing detailed tests for lexer (to examine if every token returned is of correct type), we don't worry about the token type itself.
      * some utility : all other files with `main_*` are also utilities for testing/debugging. For example, when run `/binnmodl_lexer`, it prints every token, it's type and location.

In summary, given that the 99% use of this is for testing + debugging, I think it's not big deal. But if we find the nicer solution, let's do that.

PR #273, opened 2020-03-11
Update fmt library include

mgeplf:
* use <fmt/format.h>
* remove `ext/fmt/include` from global include list
* add it to the objects that are using it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,6 @@ include_directories(${NMODL_PROJECT_SOURCE_DIR}
                     ${NMODL_PROJECT_SOURCE_DIR}/src
                     ${PROJECT_BINARY_DIR}/src
                     ${NMODL_PROJECT_SOURCE_DIR}/ext
-                    ${NMODL_PROJECT_SOURCE_DIR}/ext/fmt/include
                     ${NMODL_PROJECT_SOURCE_DIR}/ext/spdlog/include)
 
 # =============================================================================
@@ -115,8 +114,7 @@ add_subdirectory(ext/pybind11)
 # Include fmt library
 # =============================================================================
 message(STATUS "INCLUDING FMT")
-add_subdirectory(ext/fmt EXCLUDE_FROM_ALL)
-set_property(TARGET fmt PROPERTY POSITION_INDEPENDENT_CODE ON)
+add_subdirectory(${NMODL_PROJECT_SOURCE_DIR}/ext/fmt)
 
 # =============================================================================
 # Include path from external libraries
--- a/src/codegen/CMakeLists.txt
+++ b/src/codegen/CMakeLists.txt
@@ -24,6 +24,7 @@ set(CODEGEN_SOURCE_FILES
 # Codegen library and executable
 # =============================================================================
 add_library(codegen STATIC ${CODEGEN_SOURCE_FILES})
+target_link_libraries(codegen fmt::fmt-header-only)
 
 add_dependencies(codegen lexer util visitor)
 
--- a/src/pybind/CMakeLists.txt
+++ b/src/pybind/CMakeLists.txt
@@ -38,7 +38,7 @@ add_dependencies(_nmodl pyastgen)
 add_dependencies(_nmodl lexer_obj)
 add_dependencies(_nmodl util_obj)
 
-target_link_libraries(_nmodl PRIVATE fmt::fmt)
+target_link_libraries(_nmodl PRIVATE fmt::fmt-header-only)
 
 add_custom_target(copy_python_files ALL DEPENDS ${NMODL_PYTHON_FILES_OUT})
 add_custom_command(OUTPUT ${NMODL_PYTHON_FILES_OUT}
--- a/src/symtab/CMakeLists.txt
+++ b/src/symtab/CMakeLists.txt
@@ -16,6 +16,7 @@ set(SYMTAB_HEADERS
 # =============================================================================
 add_library(symtab_obj OBJECT ${SYMTAB_SOURCES} ${SYMTAB_HEADERS})
 set_property(TARGET symtab_obj PROPERTY POSITION_INDEPENDENT_CODE ON)
+target_link_libraries(symtab_obj PRIVATE fmt::fmt-header-only)
 
 add_dependencies(symtab_obj lexer_obj)
 
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -5,7 +5,7 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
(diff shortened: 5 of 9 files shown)


src/codegen/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ set(CODEGEN_SOURCE_FILES
 # Codegen library and executable
 # =============================================================================
 add_library(codegen STATIC ${CODEGEN_SOURCE_FILES})
+target_link_libraries(codegen fmt::fmt-header-only)
```
mgeplf: Note: I can't make this `PRIVATE` since `nmodl` depends on this target.

CMakeLists.txt
```diff
@@ -115,8 +114,7 @@ add_subdirectory(ext/pybind11)
 # Include fmt library
 # =============================================================================
 message(STATUS "INCLUDING FMT")
-add_subdirectory(ext/fmt EXCLUDE_FROM_ALL)
-set_property(TARGET fmt PROPERTY POSITION_INDEPENDENT_CODE ON)
```
mgeplf: I am unsettled by the forcing of `POSITION_INDEPENDENT_CODE` in a lot of places, coupled by forcing some compilations to be `STATIC`.  What is the reasoning for the latter?

CMakeLists.txt
```diff
@@ -102,7 +102,6 @@ include_directories(${NMODL_PROJECT_SOURCE_DIR}
                     ${NMODL_PROJECT_SOURCE_DIR}/src
                     ${PROJECT_BINARY_DIR}/src
                     ${NMODL_PROJECT_SOURCE_DIR}/ext
-                    ${NMODL_PROJECT_SOURCE_DIR}/ext/fmt/include
```
mgeplf: It would probably be worthwhile to clean up the other includes on this list - like I said elsewhere, they are basically like a global variable (with all those problems).

src/symtab/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ set(SYMTAB_HEADERS
 # =============================================================================
 add_library(symtab_obj OBJECT ${SYMTAB_SOURCES} ${SYMTAB_HEADERS})
```
mgeplf: This confuses me too; why are the headers (including `.hpp`) being added through `add_library`?

CMakeLists.txt
ohm314: I sort of vaguely remember that this came up after including pybind. We had to force `-fPIC`. But I am not entirely sure.

src/pybind/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ add_dependencies(_nmodl pyastgen)
 add_dependencies(_nmodl lexer_obj)
 add_dependencies(_nmodl util_obj)
 
-target_link_libraries(_nmodl PRIVATE fmt::fmt)
+target_link_libraries(_nmodl PRIVATE fmt::fmt-header-only)
```
ohm314: I did not know about this header-only target!

src/symtab/CMakeLists.txt
ohm314: hmm.. good question.. I am currently not sure. These CMakeFiles need some cleanup...

src/symtab/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ set(SYMTAB_HEADERS
...
 set_property(TARGET symtab_obj PROPERTY POSITION_INDEPENDENT_CODE ON)
+target_link_libraries(symtab_obj PRIVATE fmt::fmt-header-only)
```
ohm314: So, here you are able to set it `PRIVATE` because elsewhere it isn't? Is there maybe a more intuitive way to do this?

mgeplf: No, I can set it `PRIVATE` because only `SYMTAB_SOURCES` use fmt, so the include paths and such only need to be set there.  This is the 'clean' way to do things, imo, because it makes the dependencies as localised as possible.  I didn't change the way `nmodl` was getting its `fmt` dependency through `codegen`, because I didn't want to overly re-org stuff.

CMakeLists.txt
mgeplf: Hmm, interesting.  If I were to continue on the cmake cleanup in nmodl, I'd be tempted to see if I can remove the explicit STATICs, and once they are gone, see if this can be changed.  However, I suspect you're right.

src/symtab/CMakeLists.txt
ohm314: Alright, I see. thanks for the explanation.

ohm314: Turns out this was a bad idea. This is only supported since CMake 3.12, but we support CMake < 3.3 for building nmodl. So that breaks the build for older CMakes...

mgeplf: Rough, which part, the `PRIVATE`? They can be dropped.

> but we support CMake < 3.3 for building nmodl

Didn't know that requirement was so low.  Might be worthwhile to set it in the CI, so that this would have failed.

ohm314: Let me correct `<=3.3` and we would need 3.12 for being able to do `target_link_libraries` on OBJECT libraries. We agreed that using by default version 3.10 is OK, but that wouldn't help us here either.

ohm314: and you are right, the CI should have the cmake version that we require to catch this type of issues in the future.

mgeplf: > 3.12 for being able to do target_link_libraries on OBJECT libraries

Bah, the uneven distribution of CMake things makes things hard.

It looks like one might be able to work around it w/:
https://cmake.org/cmake/help/v3.10/manual/cmake-buildsystem.7.html#object-libraries

```
Although object libraries may not be named directly in calls to the target_link_libraries() command, they can be “linked” indirectly by using an Interface Library whose INTERFACE_SOURCES target property is set to name $<TARGET_OBJECTS:objlib>.
```
However, I doubt that refactor is worth the time, since (hopefully?) CMake 3.12 will be adopted first?

ohm314: Unfortunately, we can't adopt 3.12, because nmodl is used in coreneuron and neuron and can be built along with them through submodules - and especially for neuron we can't simply bump the cmake version. Thanks for looking up the relevant part in the doc, I'll look into it and create a new PR when ready.

mgeplf: > and especially for neuron we can't simply bump the cmake version

Yeah, I figured.  I feel your pain.

> [...] create a new PR when ready.

I personally don't consider this a high priority - I think a general cleanup and possibly looking at the forced `STATIC` would help any CMake refactoring, but that's a tedious job too...so, really, up to you :)

PR #276, opened 2020-03-12
Removed the return-by-value getter of AST node children

ohm314:
Also got rid of the _cr name ending for those functions. Now children
are always returned as const refs. If we need a copy, we construct it
from the const ref.

### DIFF
--- a/src/language/templates/visitors/nmodl_visitor.cpp
+++ b/src/language/templates/visitors/nmodl_visitor.cpp
@@ -70,9 +70,9 @@ using namespace ast;
     {% else %}
         {% call guard(child.prefix, child.suffix) %}
     {% if child.is_pointer_node %}
-        node->get_{{ child.varname }}(){{ "->" if child.is_pointer_node else "." }}accept(*this);
+        node->get_{{ child.varname }}()->accept(*this);
     {% else %}
-        {{ child.class_name }}(node->get_{{ child.varname }}()){{ "->" if child.is_pointer_node else "." }}accept(*this);
+        {{ child.class_name }}(node->get_{{ child.varname }}()).accept(*this);
     {%- endif %}
         {% endcall %}
     {%- endif %}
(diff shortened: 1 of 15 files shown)


src/language/templates/visitors/nmodl_visitor.cpp
```diff
@@ -63,13 +63,17 @@ using namespace ast;
...
     {% elif node.class_name == node_info.BINARY_EXPRESSION_NODE and child.varname == node_info.BINARY_OPERATOR_NAME %}
-        std::string op = node->get_op_cr().eval();
+        std::string op = node->get_op().eval();
         if(op == "=" || op == "&&" || op == "||" || op == "==")
             op = " " + op + " ";
         printer->add_element(op);
     {% else %}
         {% call guard(child.prefix, child.suffix) %}
+    {% if child.is_pointer_node %}
         node->get_{{ child.varname }}(){{ "->" if child.is_pointer_node else "." }}accept(*this);
```
cattabiani: `if child.is_pointer_node` is superfluous.I suggest:
`node->get_{{ child.varname }}()->accept(*this);`

src/language/templates/visitors/nmodl_visitor.cpp
```diff
@@ -63,13 +63,17 @@ using namespace ast;
...
+    {% else %}
+        {{ child.class_name }}(node->get_{{ child.varname }}()){{ "->" if child.is_pointer_node else "." }}accept(*this);
```
cattabiani: Same comment as before:
`node->get_{{ child.varname }}().accept(*this);`

src/language/templates/visitors/nmodl_visitor.cpp
cattabiani: or we remove the line before:

`{% if child.is_pointer_node %}`

ohm314: yes, sorry, you're right

PR #279, opened 2020-03-26
Undid changes from PR #273

ohm314:
Unfortunately these changes did not turn out to be compatible with cmake
3.10 and all attempted workarounds failed. In the interest of time I am
mostly rolling back these changes for now.

Azure CI is updated to explictely build with cmake 3.10.2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,7 @@ include_directories(${NMODL_PROJECT_SOURCE_DIR}
                     ${NMODL_PROJECT_SOURCE_DIR}/src
                     ${PROJECT_BINARY_DIR}/src
                     ${NMODL_PROJECT_SOURCE_DIR}/ext
+                    ${NMODL_PROJECT_SOURCE_DIR}/ext/fmt/include
                     ${NMODL_PROJECT_SOURCE_DIR}/ext/spdlog/include)
 
 # =============================================================================
@@ -124,7 +125,8 @@ add_subdirectory(ext/pybind11)
 # Include fmt library
 # =============================================================================
 message(STATUS "INCLUDING FMT")
-add_subdirectory(${NMODL_PROJECT_SOURCE_DIR}/ext/fmt)
+add_subdirectory(${NMODL_PROJECT_SOURCE_DIR}/ext/fmt EXCLUDE_FROM_ALL)
+set_property(TARGET fmt PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 # =============================================================================
 # Include path from external libraries
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,33 +46,34 @@ find_package(BISON 3.0 REQUIRED)
 # =============================================================================
 # HPC Coding Conventions
 # =============================================================================
-set(NMODL_ClangFormat_EXCLUDES_RE ".*/ext/.*$$"
-    CACHE STRING "list of regular expressions to exclude C/C++ files from formatting"
-    FORCE)
-set(NMODL_CMakeFormat_EXCLUDES_RE ".*/ext/.*$$"
-    CACHE STRING "list of regular expressions to exclude CMake files from formatting"
-    FORCE)
-set(NMODL_ClangFormat_DEPENDENCIES pyastgen parser-gen
-    CACHE STRING "list of CMake targets to build before formatting C++ code"
-    FORCE)
+set(NMODL_ClangFormat_EXCLUDES_RE
+    ".*/ext/.*$$"
+    CACHE STRING "list of regular expressions to exclude C/C++ files from formatting" FORCE)
+set(NMODL_CMakeFormat_EXCLUDES_RE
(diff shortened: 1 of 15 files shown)


pramodk:
Please retest

CMakeLists.txt
```diff
@@ -124,7 +124,8 @@ add_subdirectory(ext/pybind11)
 # Include fmt library
 # =============================================================================
 message(STATUS "INCLUDING FMT")
-add_subdirectory(${NMODL_PROJECT_SOURCE_DIR}/ext/fmt)
+add_subdirectory(${NMODL_PROJECT_SOURCE_DIR}/ext/fmt EXCLUDE_FROM_ALL)
+set_property(TARGET fmt PROPERTY POSITION_INDEPENDENT_CODE ON)
```
pramodk: Just a note - I think there was a question why PIC objects were needed - IIRC, this was needed for building pybind based nmodl module.

cmake/Catch.cmake
```diff
@@ -109,43 +106,45 @@ function(catch_discover_tests TARGET)
...
+      "TEST_SPEC=${_TEST_SPEC}" -D "TEST_EXTRA_ARGS=${_EXTRA_ARGS}" -D
+      "TEST_PROPERTIES=${_PROPERTIES}" -D "TEST_PREFIX=${_TEST_PREFIX}" -D
+      "TEST_SUFFIX=${_TEST_SUFFIX}" -D "TEST_LIST=${_TEST_LIST}" -D "CTEST_FILE=${ctest_tests_file}"
+      -P "${_CATCH_DISCOVER_TESTS_SCRIPT}"
+    VERBATIM)
+
+  file(
+    WRITE "${ctest_include_file}"
+    "if(EXISTS \"${ctest_tests_file}\")\n" "  include(\"${ctest_tests_file}\")\n" "else()\n"
+    "  add_test(${TARGET}_NOT_BUILT-${args_hash} ${TARGET}_NOT_BUILT-${args_hash})\n" "endif()\n")
```
pramodk: If you compare left and right, you will agree that this `6.0` version is worse! But anyway!

src/symtab/CMakeLists.txt
```diff
@@ -2,24 +2,22 @@
...
 # =============================================================================
 # Symbol table library
 # =============================================================================
 add_library(symtab_obj OBJECT ${SYMTAB_SOURCES} ${SYMTAB_HEADERS})
 set_property(TARGET symtab_obj PROPERTY POSITION_INDEPENDENT_CODE ON)
-target_link_libraries(symtab_obj PRIVATE fmt::fmt-header-only)
-
 add_dependencies(symtab_obj lexer_obj)
 
+# target_link_libraries(symtab_obj PRIVATE fmt::fmt-header-only)
```
pramodk:
```suggestion
```

src/symtab/CMakeLists.txt
```diff
@@ -2,24 +2,22 @@
...
 add_library(symtab STATIC $<TARGET_OBJECTS:symtab_obj>)
+# target_link_libraries(symtab fmt::fmt-header-only)
```
pramodk:
```suggestion
```

azure-pipelines.yml
```diff
@@ -13,19 +13,27 @@ jobs:
...
+      export PATH=$(pwd)/$CMAKE_PKG/bin:/home/vsts/.local/bin:$PATH
       export CXX='g++-8'
       mkdir -p $(Build.Repository.LocalPath)/build
       cd $(Build.Repository.LocalPath)/build
+      cmake --version 
       cmake .. -DPYTHON_EXECUTABLE=`which python3.7` -DCMAKE_INSTALL_PREFIX=$HOME/nmodl -DCMAKE_BUILD_TYPE=Release
       make -j 2
       make test
+    env:
+      CMAKE_PKG: 'cmake-3.10.2-Linux-x86_64'
```
pramodk: curious - what is this used for?

CMakeLists.txt
ohm314: Exactly, in the meantime I saw this too and it is really needed for pybind. Maybe this could be reworked in the future, but not worth to touch right now.

cmake/Catch.cmake
ohm314: version `0.6` you meant. I agree.. but anyway...

ohm314: version `0.6` you meant... I agree, but anyway...

azure-pipelines.yml
ohm314: This is the azure mess that I had the pleasure to deal with this morning:
azure pipeline uses ubuntu 16.04, doing `apt-get install cmake` gets you cmake 3.5.3, but the pipeline image also has the latest stable cmake installed and its path prepended, so, no matter what you install you get cmake 3.16.x . The easiest way I found to force our preferred cmake version (3.10.2) is to download and unpack the binary release from github and prepend its bin path.

Issue #280, opened 2020-03-27
CONSERVE statement OUTSIDE of a KINETIC block -> seg. fault


cattabiani:
I noticed this problem working on CONSERVE statements. The kinetic block visitor transform them in CONSERVE statements in a DERIVATIVE block. If you print to nmodl this intermediate ast state and then re-run nmodl from there, you get the seg. fault. 

- **Solution 1**: add a flag that makes the code process stuff only inside a KINETIC block (and do nothing otherwise). 

- **Solution 2**: check the node parents. However, this [PR](https://github.com/BlueBrain/nmodl/pull/257) has not been merged yet.

cattabiani:
solved with branch #285

PR #281, opened 2020-03-27
Bugfix: CONSERVE statement OUTSIDE of a KINETIC block -> seg. fault

cattabiani:
Related to [issue 280](https://github.com/BlueBrain/nmodl/issues/280)

Solution: added a flag in the KineticBlockVisitor that inhibit
visit_* functions in case we are not visiting a kinetic block.

### DIFF
--- a/src/visitors/kinetic_block_visitor.cpp
+++ b/src/visitors/kinetic_block_visitor.cpp
@@ -101,251 +101,269 @@ std::shared_ptr<ast::Expression> create_expr(const std::string& str_expr) {
 }
 
 void KineticBlockVisitor::visit_conserve(ast::Conserve* node) {
-    // rewrite CONSERVE statement in form x = ...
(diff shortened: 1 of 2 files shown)


cattabiani:
Please retest

pramodk:
@cattabiani : I merge lastest master as fmt link related issue was fixed in Omar's PR.

cattabiani:
yes, the plan is to remove the conserve from the derivative block. However, if there is, a conserve statement outside the kinetic this generates a seg. fault. The issue is somewhat orthogonal to what you are saying

nrnhines:
In NEURON the CONSERVE keyword is only allowed in a KINETIC block. Same with COMPARTMENT and LONGITUDINAL_DIFFUSION.

cattabiani:
A different [branch](https://github.com/BlueBrain/nmodl/pull/285) takes care of this bug in a more organic way. Thus, this branch was dropped.

PR #282, opened 2020-04-01
sanity: remove backticks, default to $(…)

matz-e:
This is POSIX compatible, should work everywhere and is cleaner to
read/modify since it is a proper delimiter and can be nested.


cattabiani:
I upvote this change

Issue #284, opened 2020-04-02
KineticBlockVisitor corrupts CONSERVE statements


cattabiani:
While KineticBlockVisitor rearranges CONSERVE satements to a form that can be passed to the SympySolverVisitor it malforms the statement. In particular, the react child changes derived class:

reactVarName -> varName. 

This malformed conserve statement is not understood by the SympySolverVisitor, thus it is discarded.

PR #285, opened 2020-04-02
CONSERVE statements processing (issues #280, #284)

cattabiani:
- Bugfix (issue #284): conserve statements in derivative blocks are not corrupt anymore
- Bugfix (issue #280): KineticBlockVisitor visits only kinetic blocks (no more seg. fault)
- Added a warning to enlight the user that the nmodl printed in between the KineticBlockVisitor and the SympyBlockVisitor, in presence of CONSERVE statements, is not standard (for example it does not work in mod2c). It should be used only for debugging/developing

### DIFF
--- a/src/nmodl/main.cpp
+++ b/src/nmodl/main.cpp
@@ -356,7 +356,13 @@ int main(int argc, const char* argv[]) {
             logger->info("Running KINETIC block visitor");
             KineticBlockVisitor().visit_program(ast.get());
             SymtabVisitor(update_symtab).visit_program(ast.get());
-            ast_to_nmodl(ast.get(), filepath("kinetic"));
+            const auto filename = filepath("kinetic");
+            ast_to_nmodl(ast.get(), filename);
+            if (nmodl_ast) {
+                logger->warn(
+                    "{} may present non-standard CONSERVE statements in DERIVATIVE blocks. In this case, use it only for debugging/developing"_format(
+                        filename));
+            }
         }
 
         {
--- a/src/nmodl/main.cpp
+++ b/src/nmodl/main.cpp
@@ -354,13 +354,14 @@ int main(int argc, const char* argv[]) {
         /// in global scope
         {
             logger->info("Running KINETIC block visitor");
-            KineticBlockVisitor().visit_program(ast.get());
+            auto kineticBlockVisitor = KineticBlockVisitor();
+            kineticBlockVisitor.visit_program(ast.get());
             SymtabVisitor(update_symtab).visit_program(ast.get());
             const auto filename = filepath("kinetic");
             ast_to_nmodl(ast.get(), filename);
-            if (nmodl_ast) {
+            if (nmodl_ast && kineticBlockVisitor.get_conserve_statement_count()) {
                 logger->warn(
-                    "{} may present non-standard CONSERVE statements in DERIVATIVE blocks. In this case, use it only for debugging/developing"_format(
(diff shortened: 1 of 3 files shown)


src/nmodl/main.cpp
```diff
@@ -356,7 +356,13 @@ int main(int argc, const char* argv[]) {
...
             KineticBlockVisitor().visit_program(ast.get());
             SymtabVisitor(update_symtab).visit_program(ast.get());
-            ast_to_nmodl(ast.get(), filepath("kinetic"));
+            const auto filename = filepath("kinetic");
+            ast_to_nmodl(ast.get(), filename);
+            if (nmodl_ast) {
+                logger->warn(
+                    "{} may present non-standard CONSERVE statements in DERIVATIVE blocks. In this case, use it only for debugging/developing"_format(
+                        filename));
+            }
```
pramodk, line 365: * Here we will always generate this warning even if CONSERVE statement doesn't exist.
* I would prefer to do lookup in the AST and see if there is any CONSERVE statement inside DERIVATIVE block and then only issue warning. Its extra lookup cost but would be cleaner.

src/visitors/kinetic_block_visitor.cpp
```diff
@@ -459,8 +464,13 @@ void KineticBlockVisitor::visit_program(ast::Program* node) {
...
     }
 
+    auto kineticBlockNodes = AstLookupVisitor().lookup(node, ast::AstNodeType::KINETIC_BLOCK);
     // replace reaction statements within each kinetic block with equivalent ODEs
-    node->visit_children(*this);
+    for (const auto& ii: kineticBlockNodes) {
+        ii->accept(*this);
+    }
+
+    //    node->visit_children(*this);
```
pramodk, line 473:
```suggestion
```

src/nmodl/main.cpp
cattabiani: I think I have an idea to not do another lookup...

cattabiani: done!

src/visitors/kinetic_block_visitor.hpp
```diff
@@ -132,6 +135,9 @@ class KineticBlockVisitor: public AstVisitor {
 
   public:
     KineticBlockVisitor() = default;
```
pramodk, line 137:
```suggestion
    KineticBlockVisitor() = default;
    
```

src/visitors/kinetic_block_visitor.hpp
```diff
@@ -132,6 +135,9 @@ class KineticBlockVisitor: public AstVisitor {
...
+    inline int get_conserve_statement_count() const {
```
pramodk, line 138: nitpick: 
```suggestion
    int get_conserve_statement_count() const {
```

src/visitors/kinetic_block_visitor.cpp
```diff
@@ -459,8 +466,11 @@ void KineticBlockVisitor::visit_program(ast::Program* node) {
         }
     }
 
+    auto kineticBlockNodes = AstLookupVisitor().lookup(node, ast::AstNodeType::KINETIC_BLOCK);
     // replace reaction statements within each kinetic block with equivalent ODEs
-    node->visit_children(*this);
+    for (const auto& ii: kineticBlockNodes) {
```
pramodk, line 471:
```suggestion
    for (const auto& ii: kinetic_block_nodes) {
```

PR #291, opened 2020-04-09
First attempt to develop an override checker in jinja2 templates

cattabiani:
some compilers (i.e. PGI) complain if there is no override keyword for overriding member functions. This addition to the code adds the "override" keyword in case the signature is already present in the base classes. For now it is added only for setters functions (that do not present overrides) so I could not test it. I need to find a suitable function that is not present in ast_common, is declared in a virtual class, and overridden in a derived one. Suggestions are welcome!

### DIFF
--- a/src/language/code_generator.py
+++ b/src/language/code_generator.py
@@ -58,8 +58,11 @@ if clang_format_file.exists():
 
 env = jinja2.Environment(loader=jinja2.FileSystemLoader(str(templates_dir)),
                          trim_blocks=True,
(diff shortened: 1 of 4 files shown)


pramodk:
I will wait for fixes before reviewing this as CI is failing.

cattabiani:
please retest

cattabiani:
master progressed too much and rebasing is harder than just retrying this from zero. Beside, I am not sure it is needed anymore

Issue #295, opened 2020-04-14
Ast::set_parent_in_children should not be virtual


tristan0x:
This member function is called in Ast classes constructors exclusively AFAIK. Virtual methods should not be called in constructors, more details in [isocpp.org wiki](https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctors)
I suggest making it a simple private member function in every sub-classes of ast having children.

cattabiani:
I agree. Copy-paste error

PR #296, opened 2020-04-14
Remove virtual from set_parent_for_children (issue #295)

cattabiani:
set_parent_for_children is called only in the constructors. The
virtual keyword unnecessary and generates only confusion. 
Check issue #295 for more info


pramodk:
@cattabiani : can you rename on master and resolve conflict?

pramodk:
I was bit concerned by "non-private non-virtual function being called through base class pointer pointing to derived class". But @tristan0x made that private and this is good.

PR #310, opened 2020-04-24
Avoid including ast/ast.hpp in visitor class declarations.

tristan0x:
Changes in visitor classes:
* declarations only use Ast node forward declarations
* do not inline virtual member functions and move the definitions in .cpp

### DIFF
--- a/src/language/templates/visitors/checkparent_visitor.cpp
+++ b/src/language/templates/visitors/checkparent_visitor.cpp
@@ -9,10 +9,12 @@
 /// THIS FILE IS GENERATED AT BUILD TIME AND SHALL NOT BE EDITED.
 ///
 
+#include "visitors/checkparent_visitor.hpp"
+
 #include <string>
 #include <fmt/format.h>
 
-#include "visitors/checkparent_visitor.hpp"
+#include "ast/ast.hpp"
 
 namespace nmodl {
 namespace visitor {
--- a/src/language/templates/visitors/lookup_visitor.cpp
+++ b/src/language/templates/visitors/lookup_visitor.cpp
@@ -9,8 +9,11 @@
 /// THIS FILE IS GENERATED AT BUILD TIME AND SHALL NOT BE EDITED.
 ///
 
-#include <algorithm>
 #include "visitors/lookup_visitor.hpp"
+
+#include <algorithm>
+
+#include "ast/ast.hpp"
 
 
 namespace nmodl {
--- a/src/printer/decl.hpp
+++ b/src/printer/decl.hpp
@@ -1,0 +1,21 @@
+/*************************************************************************
+ * Copyright (C) 2018-2020 Blue Brain Project
+ *
+ * This file is part of NMODL distributed under the terms of the GNU
+ * Lesser General Public License. See top-level LICENSE file for details.
+ *************************************************************************/
+
+#pragma once
+
+/**
+ * \file
+ * \brief Forward references of symbols defined in namespace nmodl::printer
+ */
+
+namespace nmodl {
+namespace printer {
+
+class JSONPrinter;
+
+}
+}  // namespace nmodl
--- a/src/symtab/decl.hpp
+++ b/src/symtab/decl.hpp
@@ -1,0 +1,20 @@
+/*************************************************************************
+ * Copyright (C) 2018-2020 Blue Brain Project
+ *
+ * This file is part of NMODL distributed under the terms of the GNU
+ * Lesser General Public License. See top-level LICENSE file for details.
+ *************************************************************************/
+
+#pragma once
+
+/// \file
+/// \brief Forward declarations of symbols in namespace nmodl::symtab
+
+namespace nmodl {
+namespace symtab {
+
+class Symbol;
+class SymbolTable;
+
+}  // namespace symtab
+}  // namespace nmodl
--- a/src/visitors/defuse_analyze_visitor.cpp
+++ b/src/visitors/defuse_analyze_visitor.cpp
@@ -55,7 +55,7 @@ std::ostream& operator<<(std::ostream& os, DUState state) {
 }
 
 /// DUInstance to JSON string
-void DUInstance::print(JSONPrinter& printer) {
+void DUInstance::print(JSONPrinter& printer) const {
     if (children.empty()) {
         printer.add_node(to_string(state));
     } else {
@@ -68,7 +68,7 @@ void DUInstance::print(JSONPrinter& printer) {
 }
 
 /// DUChain to JSON string
-std::string DUChain::to_string(bool compact) {
+std::string DUChain::to_string(bool compact) const {
     std::stringstream stream;
     JSONPrinter printer(stream);
     printer.compact_json(compact);
--- a/src/visitors/defuse_analyze_visitor.cpp
+++ b/src/visitors/defuse_analyze_visitor.cpp
@@ -5,10 +5,13 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
+#include "visitors/defuse_analyze_visitor.hpp"
+
 #include <algorithm>
 #include <utility>
 
-#include "visitors/defuse_analyze_visitor.hpp"
+#include "ast/ast.hpp"
+#include "utils/logger.hpp"
 
 namespace nmodl {
 namespace visitor {
@@ -128,7 +131,6 @@ DUState DUInstance::sub_block_eval() {
 DUState DUInstance::conditional_block_eval() {
     DUState result = DUState::NONE;
     bool block_with_none = false;
-    bool block_non_def_cdef = false;
 
     for (auto& chain: children) {
         auto child_state = chain.eval();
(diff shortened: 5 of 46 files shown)


tristan0x:
tests pass on my mac with appleclang but I can reproduce the issue on BB5. I will investigate next week

src/visitors/sympy_conductance_visitor.cpp
```diff
@@ -5,13 +5,18 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
+#include "visitors/sympy_conductance_visitor.hpp"
```
pramodk, line 8: could you clarify why this is needed at the top? if some type info needed, could before `#include "ast/ast.hpp"` with extra spacing to avoid sorting?

src/visitors/perf_visitor.hpp
```diff
@@ -199,94 +198,48 @@ class PerfVisitor: public AstVisitor {
 
     void visit_program(ast::Program& node) override;
 
-    void visit_plot_block(ast::PlotBlock& node) override {
-        measure_performance(&node);
-    }
+    void visit_plot_block(ast::PlotBlock& node) override;
 
     /// skip initial block under net_receive block
```
pramodk, line 203: general question : 
* I see that moving definitions to `.cpp` avoids include into headers.
* It perfectly make sense for common headers like `ast` or `ast_visitor`.
* But, if those headers are not used in other code but in their own compilation units, does it make sense to move single line functions to .cpp?
* I don't see any advantage in cases like `perf_visitor`. Separating declaration and definition for single liner function seems just code bloat.

src/visitors/perf_visitor.cpp
```diff
@@ -5,9 +5,12 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
+#include "visitors/perf_visitor.hpp"
```
pramodk, line 8: why this needs to before `include <utility>` ?

src/language/templates/visitors/checkparent_visitor.cpp
```diff
@@ -9,10 +9,12 @@
 /// THIS FILE IS GENERATED AT BUILD TIME AND SHALL NOT BE EDITED.
 ///
 
+#include "visitors/checkparent_visitor.hpp"
```
pramodk, line 12: could this be after `<include <fmt/format.h>` ?

src/language/templates/visitors/lookup_visitor.cpp
```diff
@@ -9,9 +9,12 @@
 /// THIS FILE IS GENERATED AT BUILD TIME AND SHALL NOT BE EDITED.
 ///
 
-#include <algorithm>
 #include "visitors/lookup_visitor.hpp"
```
pramodk, line 12: same

src/printer/decl.hpp
```diff
@@ -0,0 +1,21 @@
...
+ * \brief Forward references of symbols defined in namespace nmodl::printer
+ */
+
+namespace nmodl {
+namespace printer {
+
+class JSONPrinter;
+
+}
+}  // namespace nmodl
```
pramodk, line 21: 👍

src/symtab/decl.hpp
```diff
@@ -0,0 +1,20 @@
...
+/// \brief Forward declarations of symbols in namespace nmodl::symtab
+
+namespace nmodl {
+namespace symtab {
+
+class Symbol;
+class SymbolTable;
+
+}  // namespace symtab
+}  // namespace nmodl
```
pramodk, line 20: 👍

src/visitors/defuse_analyze_visitor.cpp
```diff
@@ -5,10 +5,13 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
+#include "visitors/defuse_analyze_visitor.hpp"
```
pramodk, line 8: same question : I might be missing something obvious!

src/visitors/defuse_analyze_visitor.cpp
```diff
@@ -270,6 +272,73 @@ void DefUseAnalyzeVisitor::visit_verbatim(ast::Verbatim& node) {
...
+}
+
+void DefUseAnalyzeVisitor::visit_lin_equation(ast::LinEquation& node) {
+    visit_unsupported_node(node);
+}
+
+void DefUseAnalyzeVisitor::visit_partial_boundary(ast::PartialBoundary& node) {
+    visit_unsupported_node(node);
+}
+
```
pramodk, line 293: See my comments for `perf_visitor.hpp` which are applicable here. Some files I reviewed from the bottom :)

src/visitors/defuse_analyze_visitor.hpp
```diff
@@ -237,71 +234,34 @@ class DefUseAnalyzeVisitor: public AstVisitor {
...
-    }
-
-    void visit_non_lin_equation(ast::NonLinEquation& node) override {
-        visit_unsupported_node(&node);
-    }
-
-    void visit_lin_equation(ast::LinEquation& node) override {
-        visit_unsupported_node(&node);
-    }
-
```
pramodk, line 251: Same as perf_visitor.hpp - does it make sense to move out one liner function if those header are not used out-of-their-own-compilation-units?

src/visitors/sympy_conductance_visitor.cpp
tristan0x: I start thinking that every .cpp should begin with the inclusion of its corresponding header. This way, you can ensure that the .hpp is not missing a required header.
If it does, better have a compilation issue in the corresponding .cpp than in the rest of the project (or in another project if the .hpp is public) that includes it in order to use it.

If you agree, we can generalize that to the rest of the project in a future PR. Otherwise, I can move the `#include` to the proper location.

src/visitors/perf_visitor.hpp
tristan0x: > * But, if those headers are not used in other code but in their own compilation units, 

I am not sure to follow. Isn't a header useless if only useful to its compilation unit? Isn't its purpose to expose symbols used by other compilation units?

Most visitors expose an API that deal with AST top nodes like Ast, Program, ...
From what I saw in this project, most code that uses visitors only manipulate these nodes, so there is not point overwhelm them with `ast.hpp`

> does it make sense to move single line functions to .cpp?

From a performance point of view, inline virtual member methods are not inlined by the compiler. And this is the case in most of the time here.

> I don't see any advantage in cases like perf_visitor. Separating declaration and definition for single liner function seems just code bloat.

Code bloating is subjective, and I tend to not agree with you. IMO having `inline` method definitions inside a class definition is a necessary evil to have more efficient code. I wish headers were clean, and user oriented. For instance, I would even mark the `visit_` method `private` because completely out of context for the final class user.

But I respect the beliefs of the code maintainers :) Let me know if you prefer me to rollback the changes in `perf_visitor`

src/visitors/sympy_conductance_visitor.cpp
pramodk: > I start thinking that every .cpp should begin with the inclusion of its corresponding header.

I also didn't realise before why this is necessary. It perfectly make sense now (to make sure the header is not dependent on other includes). I see this also documented [here]( https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

src/visitors/perf_visitor.hpp
pramodk: > I am not sure to follow. Isn't a header useless if only useful to its compilation unit? Isn't its purpose to expose symbols used by other compilation units?
> Code bloating is subjective, and I tend to not agree with you. I

Ah, I meant in the following context:

```
# this is foo.hpp

#include "bar.hpp"

class Foo {
    void some_simple_one_liner_method() {
           bar_obj->some_method();
    }

   // other complex methods that will go into foo.cpp
   void some_complex_method();
}
```
In this example, `bar` forward declaration is not sufficient. So to avoid include `bar.hpp`, we move `some_simple_one_liner_method` to `foo.cpp`. Given that:
* foo.hpp is not included by other files 
* do we need to move `one liner` methods to `.cpp` ?

 thought that is duplication. While writing this, I see an example where `foo.hpp` will be included in `main.cpp` where `Foo` will be instantiated and it will pull in `bar.hpp`! I see the point.

> For instance, I would even mark the visit_ method private because completely out of context for the final class user.

yes, for these classes it make sense.

src/visitors/sympy_conductance_visitor.cpp
tristan0x: Perfect. Thank you for the link, I was considering adding this to our coding conventions :)

src/visitors/perf_visitor.hpp
tristan0x: I pushed 2 last changes. One about const-ness, the other to hide some visit_ methods when they are not meant to be called as we discussed in this thread.

Issue #298, opened 2020-04-16
Reduce compilation time of visitors


tristan0x:
## Target

We can speed up the compilation time by:
1. avoid including `ast/ast.hpp` when not necessary. For instance visitor class declaration should only use forward declarations
1. most visitors only visit a sub-set of the AST nodes. Therefore, visitor class definitions in .cpp should only include the AST nodes required.

## Proposal

* First point can be easily changed, actually I have a branch where it is already done.
* Second point requires more changes:
  * generate one C++ header per AST class. nmodl.yaml should already contain the necessary information to build the AST nodes inter-dependencies to create the proper `#include <ast/ast_XXX.hpp>` directives on-top of each header file.
   * There is no need to split ast.cpp.
   * Update ast.hpp to either:
     1.include all ast_XXX.hpp headers but it will increase the amount of I/O 
     2. duplicate the content in `ast.hpp` but add preprocessor fences to prevent a class to be declared twice if a C++ file includes both ast.hpp and ast_xxx.hpp. _This is my favorite option_

cattabiani:
First point: no problemo, +1
Second point: it could be done. However, auto-generate the includes could lead to compilation errors. Easy to debug but potentially high in number. I am not sure how much faster the compilation would be. I should look more closely into this topic

ohm314:
I like your proposal @tristan0x ! Regarding the last point, I think 2. will be the better solution even though I dislike preproc fences.

ohm314:
@cattabiani I think compilation will be substantially faster, also the IDE does not choke anymore on the big headers being included. The auto-generation of the `ast_XXX.hpp` files should be not a problem at all, the code is already generated correctly, now it should just go into a new file for each class. Most dependencies between ast class declarations can be done with fwd declarations too, no?

pramodk:
I will post some details in the evening but it would be worth looking at pyast and see if it’s only header size or something else (e.g. pybind11).

I am mentioning this because I never noticed the issue with cpp compilation but as soon as pybind stuff comes in, it causes the problem.

(we should definitely reduce the header size for sure but just pointing out the issue that has caused this discussion ).

pramodk:
Spent some time to look into root cause, see #299 and summary/conclusion.

> First point can be easily changed, actually I have a branch where it is already done.

👍 straightforward.

[...]

PR #312, opened 2020-05-01
Refactor code generator / Improve usage of AST classes

tristan0x:
Compilation units now only include the AST classes they use.
It essentially required the code generator to create one C++ header file per AST node.

It led to a refactoring of `src/language/code_generator.py`:
* do not use global variables
* split the procedural code into functions
* use object oriented paradigms
* do no change working directory during execution to ease file management
* Improve dependency managements of Jinja tasks.
  A task is now considered out of date is the output file is missing or
  if one of its dependencies has changed:
  - the Jinja template(s)
  - the python files used by this program
  - the YAML files describing the NMODL language
* Code genearator now create C++ headers in BUILD_DIR/ast/, one file per AST node
* Code generator creates a new file: `src/language/code_generator.cmake`
  It defines a set CMake variables to be used elsewhere in the CMake
  project to manipulate inputs and outputs of the code generator.
  This file is updated by the code generator if a Jinja template is added
  or removed, or is a NMODL AST class is added for instance.
  The file is added to revision control on purpose to break the cyclic dependency
  but is not meant to be edited manually.
* Some existing output files have been changed:
  - rename build_dir/src/ast/ast.hpp to build_dir/src/ast/all.hpp
  - build_dir/src/ast/ast.hpp now only provide nmodl::Ast class declaration
* Ast nodes: move the definition `get_node_name` and `set_name methods` in
  compilation units to avoid providing child class definitions in header file.
  These methods were not inlined anyway because `virtual`.
* Fix typos in nmodl.yaml

Along the way:
* More const& for parameters and return types
* More `std::move` essentially in constructors
* More `const noexcept` for member methods
* Prefer references for Ast node parameters instead of pointers when possible
* Hide some visitors `visit_*` methods that should not be called directly.
* Use brackets to include catch2 header
* Use `std::vector::reserve` when possible
* Remove useless header inclusions
* `CodegenCompatibilityVisitor`:
  - `unhandled_ast_types_func` member variable is now a *const static* std::map.
    Move definition in .cpp to avoid including headers in .hpp that are
    specific to the implementation
  - update visibility of some member methods to `private`


src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -5,12 +5,25 @@
...
+#include "ast/global.hpp"
+#include "ast/independent_block.hpp"
+#include "ast/match_block.hpp"
+#include "ast/name.hpp"
+#include "ast/partial_block.hpp"
+#include "ast/pointer_var.hpp"
+#include "ast/solve_block.hpp"
+#include "ast/string.hpp"
+#include "ast/terminal_block.hpp"
+#include "ast/verbatim.hpp"
```
pramodk, line 26: Thanks @tristan0x ! I will take a detailed look but did you see my comment in https://github.com/BlueBrain/nmodl/issues/298#issuecomment-614939876?

tristan0x: I half-read it apparently. I didn't see your proposal...
Indeed this patch doesn't address the issues with pyast.cpp and pyvisitor.cpp but I experience a 10-15% build time improvements on my imac.

src/language/code_generator.cmake
```diff
@@ -0,0 +1,222 @@
+#
+# THIS FILE IS GENERATED AND SHALL NOT BE EDITED.
```
ohm314, line 2: probably a stupid question - but if this is generated, does it need to be committed to the repo?

src/language/templates/code_generator.cmake
```diff
@@ -0,0 +1,37 @@
+#
+# THIS FILE IS GENERATED AND SHALL NOT BE EDITED.
+#
+
+# cmake-format: off
+set(CODE_GENERATOR_JINJA_FILES
```
ohm314, line 6: I'm just curious, so it looks like you generate this cmake file too. How does that go into the build process. Don't you need all cmake files to be there already before you can run `cmake ..`? It's really cool that this works, but I don't quite understand how

src/language/templates/pybind/pyast.cpp
```diff
@@ -9,10 +9,11 @@
 /// THIS FILE IS GENERATED AT BUILD TIME AND SHALL NOT BE EDITED.
 ///
 
+#include "pybind/pyast.hpp"
```
ohm314, line 12: Don't we put project-internal includes _after_ external includes?

src/language/templates/code_generator.cmake
iomaganaris: I have the same question. As far as I understood this file is added to the repo so that the `cmake` can run for the first time. It's not very clear to me how when we add an `ast` class then `cmake` is going to use the updated `code_generator.cmake` to set all the variables though

src/language/code_generator.cmake
tristan0x: Allow me copy/paste my PR description:

> Code generator creates a new file: src/language/code_generator.cmake
It defines a set CMake variables to be used elsewhere in the CMake
project to manipulate inputs and outputs of the code generator.
This file is updated by the code generator if a Jinja template is added
or removed, or is a NMODL AST class is added for instance.
The file is added to revision control on purpose to break the cyclic dependency
but is not meant to be edited manually.

Description of the cyclic dependency.
1. This file is required by CMake
1. This file is generated by code_generator.py
1. code_generator.py is driven by CMake

Add the file to the repo is not ideal but that's fine IMO

src/language/templates/pybind/pyast.cpp
tristan0x: TLDR: it has changed last week :)
We should now use [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) conventions about header inclusion.

This is something @pramodk and I discussed [here](https://github.com/BlueBrain/nmodl/pull/310#discussion_r417095638)

* there will be a PR in nmodl shortly to update all compilation units accordingly
* we should communicate about in BBP, at least within HPC

src/codegen/codegen_compatibility_visitor.hpp
```diff
@@ -47,35 +47,7 @@ class CodegenCompatibilityVisitor: public visitor::AstVisitor {
 
     /// Unordered_map to find the function needed to be called in
```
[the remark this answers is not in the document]
tristan0x: Done in 5cf2484
Do not hesitate to provide [GitHub suggestions](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request) for such changes.

src/codegen/codegen_compatibility_visitor.cpp
pramodk: As this is a significant change, I would mark this PR as `changes requested` until I test this thoroughly . My first though is that we have around >120 AST classes in the NMODL. Having those as a separate header and including individually is bit inconvenient (I will elaborate this later).

src/codegen/codegen_compatibility_visitor.hpp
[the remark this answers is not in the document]
iomaganaris: You're right, I am sorry and thanks for the fix!

src/codegen/codegen_compatibility_visitor.cpp
tristan0x: One improvement I can think of is to generate one additional header for every group of classes under Ast with a maximum depth of 1 or 2 i.e:

* `ast/expressions.hpp`
  * `ast/numbers.hpp`
  * `ast/identifiers.hpp`
  * `ast/blocks.hpp`
* `ast/statements.hpp`

pramodk:
As briefly discussed last week, I build this PR on MacBook pro laptop:

```
# Master Branch
[100%] Built target nmodl
real	3m1.734s

# This PR
[100%] Built target nmodl
real	2m49.733s
```

So about ~7% improvement with the current changes. Obviously any improvement for the build time is good but this still doesn't address yet the main problem (coming from pybin11d module compilation?).

> One improvement I can think of is to generate one additional header for every group of classes > under Ast with a maximum depth of 1 or 2 i.e:
> ast/expressions.hpp
> ast/numbers.hpp
> ast/identifiers.hpp
> ast/blocks.hpp
> ast/statements.hpp

This could be one possibility. Another way I was thinking is more semantic structuring : e.g. only `NMODL` construct derived from `Block` can have their separate header. These are main, top level building blocks in NMODL.

**But** before making another design change and testing improvement at the end, what if we do following:
* with this PR we have now finer granularity (i.e. each node with it's header)
* we try to solve the main issue i.e. see how pybind and related classes could be split as well and build those faster
* once we get a good idea of what needs to be done to speedup `hotspot` for the build, we can decide what is best approach.
* I hope testing main hotspots would be faster. If it's helpful, we can disable build of many non-essential targets/objects and just focus on the largest time taking files.

src/language/templates/code_generator.cmake
ohm314: OK, so this was explained elsewhere. The file is assumed to exist and is updated whenever a new template file is added. I'm not too happy that this file is generated, I would rather this step is done manually and then this cyclic dependency and possible confusion isn't there in the first place. But I also don't feel strongly about it either way.

ohm314:
I think this PR is in good shape, the additional improvements to the way pybind headers and pybind jinja tempalates are handled could go into a separate PR. In principle I don't see why this contribution cannot be merged soon.

pramodk:
>  In principle I don't see why this contribution cannot be merged soon.

@ohm314 : See my [earlier comments](https://github.com/BlueBrain/nmodl/pull/312#issuecomment-627016766) comments. Certainly it improves the build time and overall good approach, we must address the main issues before merging this.

tristan0x:
> I think this PR is in good shape, the additional improvements to the way pybind headers and pybind jinja tempalates are handled could go into a separate PR. In principle I don't see why this contribution cannot be merged soon.

I agree with @ohm314 : The initial purpose of this PR was not about solving the pybind visitor build issue. It provides more flexibility that may help its resolution in a future PR though.

pramodk:
> I agree with @ohm314 : The initial purpose of this PR was not about solving the pybind visitor build issue.

Ok, let me re-review then considering this aspect.

pramodk:
Ok, I went through entire PR now. For sure, **this-is-a-significant-addition**!!

Here is feedback:

* The code generator logic in `src/language/code_generator.py` is getting bit advanced that the previous implementation but most likely this will remain static anyway. So not an issue.

* while building code we see hundred of lines showing which file is being updated. I suggest to remove this and just print single status line, something like `Updating ast header files X / Y` (X out of total Y or something similar.

```
[ 12%] Building CXX object test/CMakeFiles/testprinter.dir/printer/printer.cpp.o
Updated out of date template files:
  templates/visitors/visitor.hpp             -> /Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/build/src/visitors/visitor.hpp
  templates/visitors/nmodl_visitor.cpp       -> /Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/build/src/visitors/nmodl_visitor.cpp
  templates/visitors/json_visitor.hpp        -> /Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/build/src/visitors/json_visitor.hpp
  templates/visitors/lookup_visitor.hpp      -> /Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/build/src/visitors/lookup_visitor.hpp
  templates/visitors/ast_visitor.cpp         -> /Users/kumbhar/workarena/repos/bbp/incubator/nocmodl/build/src/visitors/ast_visitor.cpp
...
```

* `src/language/templates/code_generator.cmake` also caught my attention like Omar and Ioannis. Bit concerned by auto-generation in Cmake as well (**_more moving parts_**) but hopefully this will be robust and won't change much. No strong opinion but we can raise concerns if we see issues while developing further.

* No doubt that this PR can be used as a base for future PR to split pybind related headers and reduce build times. But, granularity of AST splitting (per node type or certain depth or some logical splitting based on node types) is something that we haven't defined yet. Including large number of headers in some visitors is not convincing to me because of 1) to keep simplicity 2) some of the node types doesn't have semantic meaning in NMODL language (they were created to  complete grammar rules). So I suggest following:
* there are many files where we include 2 or max 3 node headers. we can use the new, small headers as done in this PR.
* there are following headers where we include more number of headers, I would suggest to include there `ast/all.hpp` for now. Once we decide on the granularity of nodes in subsequent PRs  , we can introduce those granular includes everywhere. (note : I am not saying we shouldn't split AST but lets not introduce those individual headers right away everywhere). Those files are:
  * src/codegen/codegen_compatibility_visitor.cpp
  * src/codegen/codegen_ispc_visitor.cpp
  * src/visitors/constant_folder_visitor.cpp
  * src/visitors/rename_visitor.cpp
  * src/visitors/solve_block_visitor.cpp
  * src/visitors/steadystate_visitor.cpp
  * src/visitors/sympy_conductance_visitor.cpp
  * src/visitors/sympy_solver_visitor.cpp
  * src/visitors/units_visitor.cpp
  * src/visitors/visitor_utils.cpp
  * test/visitor/units.cpp

tristan0x:
Thank you for the review @pramodk 
Your remarks have been taken into account in the 5 last commits. CI is green again!

pramodk:
Thanks @tristan0x! I will review this today!

src/codegen/codegen_compatibility_visitor.cpp
pramodk: this suggested we will incorporate into separate PR.

PR #318, opened 2020-05-03
Rename context to eq_context for compatibility with Bison v3.6.0

pramodk:
- upcoming bison release 3.6.0 will have new type called `class context`
 - this will conflict with the variable name used in differential
   equation parser and hence we rename it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,14 @@ matrix:
         - MATRIX_EVAL="CXX=c++"
 
 #=============================================================================
+# Cache directories
+#=============================================================================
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -26,6 +26,14 @@ matrix:
       env:
         - MATRIX_EVAL="CXX=c++"
 
+#=============================================================================
+# Cache directories
```
ohm314, line 30: nice! I saw this in my periphery, but didn't get a chance to look into it. Very useful.

PR #320, opened 2020-05-05
Fix incompatibility issues with manylinux1 platform

### LINKED ISSUE #319: Fix portability issues with wheel on manylinux1 platform
As we are targeting NMODL release with NEURON wheel, build the NMODL on manylinux1 platform and address any incompatibilities if any.

pramodk:
#### Motivation
We would like to ship nmodl with NEURON manylinux1 wheel during upcoming release 8.0 

#### Changes
- manylinux1 platform provides gcc 4.8.2 for wider compatibility
- gcc 4.8.2 is first gnu compiler supporting `almost all` C++11
  except that it lacks `const_iterator` overload for std::vector::erase
  and std::vector::insert
- in this PR we introduce const_iter_cast thar turns const iterator
  into non-const-iterator to have minimal changes to the code base
- change the compatibility to gcc 4.8.2 for cmake as well as json library
- change `pybind11::literals::operator""_a` to `namespace pybind11::literals`
  as documented in https://pybind11.readthedocs.io/en/stable/basics.html#keyword-arguments

fixes #319

#### Considerations
* This PR is **not an attempt** to bind to the oldest compiler
* We have [started investigation to upgrade to new gcc](https://github.com/neuronsimulator/nrn/issues/511) (e.g. v7 or v8) under manylinux platfor. This should happen soon after 8.0 release.
* For this 8.0 relate, NEURON wheel infrastructure is setup with manylinux1 and would be good to have current nmodl ship with NEURON
* As the changes are minimal here, it won't be a big deal to get this in.

### DIFF
--- a/src/utils/common_utils.hpp
+++ b/src/utils/common_utils.hpp
@@ -48,6 +48,32 @@ T remove_extension(T const& filename) {
     return p > 0 && p != T::npos ? filename.substr(0, p) : filename;
 }
 
+/**
+ * Retrun non-const iterator corresponding to the const_iterator in a vector
+ *
+ * Some old compilers like GCC v4.8.2  has C++11 support but missing erase and insert
+ * with const_iterator implementation. This is a workaround to handle build issues with
+ * such compilers especially on manylnux1 platform.
+ *
+ * See bug report : https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57158
+ *
+ * \todo Remove this after move to manylinux2010 platform.
+ */
+#if (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) < 40900
+template<typename T>
+typename std::vector<T>::iterator
+const_iter_cast(std::vector<T>& v, typename std::vector<T>::const_iterator iter) {
+    return v.begin() + (iter - v.cbegin());
+}
+#else
+template<typename T>
+typename std::vector<T>::const_iterator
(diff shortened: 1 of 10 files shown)


src/utils/common_utils.hpp
```diff
@@ -48,6 +48,32 @@ T remove_extension(T const& filename) {
...
+ *
+ * \todo Remove this after move to manylinux2010 platform.
+ */
+#if (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) < 40900
+template<typename T>
+typename std::vector<T>::iterator
+const_iter_cast(std::vector<T>& v, typename std::vector<T>::const_iterator iter) {
+    return v.begin() + (iter - v.cbegin());
+}
+#else
```
pramodk, line 68: I did this if-else logic so that newer compilers don’t need to use workaround solution.

src/utils/common_utils.hpp
```diff
@@ -48,6 +48,32 @@ T remove_extension(T const& filename) {
     return p > 0 && p != T::npos ? filename.substr(0, p) : filename;
 }
 
+/**
+ * Retrun non-const iterator corresponding to the const_iterator in a vector
```
ohm314, line 52:
```suggestion
 * Return non-const iterator corresponding to the const_iterator in a vector
```

cattabiani:
checking jenkins looks like there is some code left for debugging that prints the pythonpath for every test...

pramodk:
> checking jenkins looks like there is some code left for debugging that prints the pythonpath for every test...

that might be just some old builds. I dont see any debug code into latest commits.

PR #321, opened 2020-05-05
Support for building with Python 3.5

pramodk:
- NMODL can be installed with any Python 3 version except for the
    dependency on python interpreter v3.6 for AST classes generation.
  - In this PR, we find libraries and dependency modules for any
    python version but try to find python v3.6 for AST class generation.
  - This allows to build the wheel for Python 3.5 because newer
    python interpreter is also available on the system.


CMakeLists.txt
```diff
@@ -145,6 +145,34 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/config/config.cpp.in
...
+set(MEMORYCHECK_COMMAND_OPTIONS
+    "--trace-children=yes \
+                                 --leak-check=full \
+                                 --track-origins=yes \
+                                 --show-possibly-lost=no")
+# do not enable tests if nmodl is used as submodule
+if(NOT NMODL_AS_SUBPROJECT)
+  include(CTest)
+  add_subdirectory(test)
+endif()
```
pramodk, line 161: this part is just got moved at the top

CMakeLists.txt
```diff
@@ -145,6 +145,34 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/config/config.cpp.in
...
+# =============================================================================
+# ~~~
+# AST classes are generated by python scripts which requires Python >= 3.6
+# In case of older python, try to find newer python interpreter
+# ~~~
+if(PYTHON_VERSION_MINOR VERSION_LESS 6)
+  unset(PYTHONINTERP_FOUND CACHE)
+  unset(PYTHON_EXECUTABLE CACHE)
+  find_package(PythonInterp 3.6 REQUIRED)
+endif()
```
pramodk, line 174: this is only new part here

CMakeLists.txt
```diff
@@ -145,6 +145,34 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/config/config.cpp.in
...
+if(NOT NMODL_AS_SUBPROJECT)
+  include(CTest)
+  add_subdirectory(test)
+endif()
+
+# =============================================================================
+# Check newer python for generating AST classes
+# =============================================================================
+# ~~~
+# AST classes are generated by python scripts which requires Python >= 3.6
```
cattabiani, line 167: require

pramodk: thanks! fixed.

PR #325, opened 2020-05-07
Move the pybind11 embedded python into a dynamically loadable plugin

ohm314:
- Create a wrapper library that handles all the access to libpython*.so
- Add code to dlopen/dlsym the wrapper library after loading libpython
- Redirect access to pybind11/libpython through the wrapper specifically in the
  sympy visitor.
- Update cmake files accordingly

also:
- Update pybind11 to 2.5.0 release, which fixes a number of issues and improves
  support for newer python versions

### To be completed
before this becomes a grown-up PR:
- [x] make `nmodl::pybind_wrappers::load_libraries` platform independent and improve error reporting
- [x] investigate how to fit both the case of `LINK_AGAINST_PYTHON' `True` and `False` properly
- [x] fix `testvisitor`

### For later

- avoid copies when creating the `locals` dictionary in the `SympyVisitor`. 
  since this dictionnary is really quite small with strings (just the input/output equations, I think in most cases this really wouldn't be that slow. Having this said, the testvisitor is running quite slow when it encounters the sympy tests, I wonder if we need to look into this already now or if this could be deferred.
- see if the factory and cleanup functions can be done in a more elegant way (currently there is for each wrapper struct one factory and cleanup function because of the limitation that only one `extern "C"` function can have the same name (no overloading allowed).  
  I have an idea on how this could be done, but the current implementation is good enough, this can be improved later.


ohm314:
please retest

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -424,47 +398,25 @@ void SympySolverVisitor::visit_diff_eq_expression(ast::DiffEqExpression& node) {
...
+    diffeq_solver->vars = vars;
+    diffeq_solver->use_pade_approx = use_pade_approx;
+    diffeq_solver->function_calls = function_calls;
+    diffeq_solver->method = solve_method;
+    (*diffeq_solver)();
     if (solve_method == codegen::naming::EULER_METHOD) {
-        logger->debug("SympySolverVisitor :: EULER - solving: {}", node_as_nmodl);
         // replace x' = f(x) differential equation
         // with forwards Euler timestep:
         // x = x + f(x) * dt
```
pramodk, line 413: nitpick : as these are now in `diffeq_solver`, they can babe removed?

src/visitors/sympy_conductance_visitor.cpp
```diff
@@ -79,26 +80,17 @@ std::vector<std::string> SympyConductanceVisitor::generate_statement_strings(
...
-                            except Exception as e:
-                                # if we fail, fail silently and return empty string
-                                solution = ""
-                                exception_message = str(e)
-                        )",
-                     py::globals(),
-                     locals);
-            auto dIdV = locals["solution"].cast<std::string>();
-            auto exception_message = locals["exception_message"].cast<std::string>();
+            auto analytic_diff = nmodl::pybind_wrappers::EmbeddedPythonLoader::get_instance()
```
pramodk, line 83: nitpick : `nmodl::` could be avoided as we are already in nmodl namesake

ohm314:
> Had a quick look and overall looks neat + nice!!
> 
> Most my comments are minor. One thing that I noticed is that when I run tests with `-DLINK_AGAINST_PYTHON=OFF`, fewer tests are run. I assume that's not expected?
> 
> I will do additional testing tomorrow.

When you build with `-DLINK_AGAINST_PYTHON=OFF` you need to prepare your environment 
* `NMODL_PYLIB` is set to the `libpython.so` path
* `NMODL_PYWRAPPER` is set to the libpywrapp.so` path
* optionally `PYTHONPATH` points to where sympy and other libs are, if they are not in the standard system path.
this needs to be done before `cmake` or prepended to the `cmake` command because ctest needs to pick up these environment variable such that at test discovery the environment is correctly set. This doesn't matter for the user, but is only for ctest to work correctly. Ping me on slack if you have questions about this.

src/nmodl/CMakeLists.txt
```diff
@@ -8,19 +8,35 @@ set(NMODL_SOURCE_FILES ${CMAKE_CURRENT_SOURCE_DIR}/main.cpp)
...
+  target_link_libraries(
+    nmodl
+    printer
+    codegen
+    visitor
+    symtab
+    util
+    lexer
+    pywrapper
+    pyembed
```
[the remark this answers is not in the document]
ohm314: You are right! I had the same thought, but didn't do it yet. Will fix

src/nmodl/main.cpp
```diff
@@ -256,7 +256,9 @@ int main(int argc, const char* argv[]) {
     utils::make_path(scratch_dir);
 
     if (sympy_opt) {
-        pybind11::initialize_interpreter();
+        nmodl::pybind_wrappers::EmbeddedPythonLoader::get_instance()
+            .api()
+            ->initialize_interpreter();
```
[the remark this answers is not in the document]
ohm314: You mean, to shorten this command and hide `api()` ? I can do that. I wanted to keep the `api()` struct explicit and have this as the only Point of reference for the developer. I thought this makes the API easier to use for someone later. They basically only need to know what the `py_wrapper_api` struct contains. But for `init` and `fin` I can do these shorthands too.

src/pybind/CMakeLists.txt
```diff
@@ -18,6 +18,37 @@ foreach(
...
+    ${PROJECT_BINARY_DIR}/src/pybind/pysymtab.cpp
+    ${NMODL_PROJECT_SOURCE_DIR}/src/pybind/pynmodl.cpp)
+
+set(PYNMODL_HEADERS
+    ${PROJECT_BINARY_DIR}/src/ast/ast_decl.hpp
+    ${NMODL_PROJECT_SOURCE_DIR}/src/ast/ast_common.hpp
+    ${PROJECT_BINARY_DIR}/src/visitors/visitor.hpp
+    ${NMODL_PROJECT_SOURCE_DIR}/src/pybind/pybind_utils.hpp
+    ${PROJECT_BINARY_DIR}/src/pybind/pyast.hpp
+    ${PROJECT_BINARY_DIR}/src/pybind/pyvisitor.hpp)
```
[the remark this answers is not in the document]
ohm314: I don't understand? you mean, i did a bad rebase?

src/pybind/pyembed.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+
+struct py_executor {
+    virtual ~py_executor() {}
+
+    virtual void operator()() = 0;
+};
+
+
+struct solve_linear_system_executor: public py_executor {
```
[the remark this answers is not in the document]
ohm314: humm.. good point. There is no good reason. I should fix.

src/pybind/pyembed.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "pybind/pyembed.hpp"
+#include "utils/logger.hpp"
+
+namespace nmodl {
+
+namespace pybind_wrappers {
+
+bool EmbeddedPythonLoader::have_wrappers() {
+    wrappers = static_cast<pybind_wrap_api*>(dlsym(RTLD_NEXT, "wrapper_api"));
```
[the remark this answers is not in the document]
ohm314: Ah yes, will fix.

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -424,47 +398,25 @@ void SympySolverVisitor::visit_diff_eq_expression(ast::DiffEqExpression& node) {
...
-                                           use_pade_approx)
-                except Exception as e:
-                    # if we fail, fail silently and return empty string
-                    solution = ""
-                    exception_message = str(e)
-            )",
-                 py::globals(),
-                 locals);
     } else {
         // for other solver methods: just collect the ODEs & return
```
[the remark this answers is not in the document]
ohm314: I think not? at least I did not modify this branch... can you elaborate?

src/visitors/sympy_solver_visitor.cpp
ohm314: I'm keeping these branches for logging - to not lose this behavior. the python wrappers I want to keep free of other code, it should be really the minimum glue for running the python code. But I agree it's not very satisfactory...

src/nmodl/main.cpp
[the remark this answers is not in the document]
pramodk: anything is fine - I just noticed that clang-format split that on 3 lines and hence thought.

src/pybind/CMakeLists.txt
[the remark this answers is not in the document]
pramodk: Are these `PYNMODL_HEADERS` and `PYNMODL_SOURCES` variables? I didn't see them but I might have missed.

src/visitors/sympy_solver_visitor.cpp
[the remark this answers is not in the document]
pramodk: If I am reading correctly - initially `py::exec(...)` was only called for two `if` statements (EULER and CNEXP). In the last `else` block we just collect the ODEs. Hence thought  `(*diffeq_solver)();` shouldn't be called outside (i.e. in case of `else` statement).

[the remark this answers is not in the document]
ohm314: Ah, but then you see that inside `diffeq_solver` we have the same branches and there the `else` is also left empty.

src/pybind/CMakeLists.txt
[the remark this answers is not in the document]
ohm314: No.. I think i screwed up a rebase, removing those right now!

src/pybind/CMakeLists.txt
```diff
@@ -18,6 +18,30 @@ foreach(
...
 
+add_library(pyembed ${CMAKE_CURRENT_SOURCE_DIR}/pyembed.cpp)
+set_property(TARGET pyembed PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+if(NOT LINK_AGAINST_PYTHON)
+  add_library(pywrapper SHARED ${CMAKE_CURRENT_SOURCE_DIR}/wrapper.cpp)
+else()
+  add_library(pywrapper ${CMAKE_CURRENT_SOURCE_DIR}/wrapper.cpp)
+  set_property(TARGET pywrapper PROPERTY POSITION_INDEPENDENT_CODE ON)
+  target_compile_definitions(pyembed PRIVATE STATIC_PYWRAPPER=1)
```
pramodk, line 29: I am not sure if Tristan document it in hpc coding convention but general suggestion was to name macros with project name i.e. `NMODL_STATIC_PYWRAPPER`. This way, they won't interfere if parent project accidentally introduce the same macros (e.g. in case of submodule).

src/pybind/CMakeLists.txt
```diff
@@ -18,6 +18,30 @@ foreach(
...
+endif()
+
+target_include_directories(pyembed PRIVATE ${PYBIND11_INCLUDE_DIR} ${PYTHON_INCLUDE_DIRS})
+target_include_directories(pywrapper PRIVATE ${pybind11_INCLUDE_DIRS} ${PYTHON_INCLUDE_DIRS})
```
pramodk, line 33: Ah, I got confused here but  see now this `PYBIND11_INCLUDE_DIR` and `pybind11_INCLUDE_DIRS` is very subtle!
Just FYI : when we build the wheel, there are no python lib/include available and hence PYBIND11_INCLUDE_DIR== pybind11_INCLUDE_DIRS.

src/pybind/pyembed.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include "pybind/pyembed.hpp"
+#include "utils/logger.hpp"
+
+namespace nmodl {
+
+namespace pybind_wrappers {
+
+bool EmbeddedPythonLoader::have_wrappers() {
+#if defined(STATIC_PYWRAPPER)
```
pramodk, line 19: as commented earlier, would be nice to have `NMODL_` prefix to project macros.

src/pybind/pyembed.cpp
```diff
@@ -0,0 +1,85 @@
...
+    return wrappers != nullptr;
+#endif
+}
+
+void EmbeddedPythonLoader::load_libraries() {
+    const auto pylib_env = std::getenv("NMODL_PYLIB");
+    if (!pylib_env) {
+        logger->critical("NMODL_PYLIB environment variable must be set to load embedded python");
+        throw std::runtime_error("NMODL_PYLIB not set");
+    }
```
pramodk, line 34: yesterday when I was building this, I was seeing exception at the time of `make` and build was finishing without error status and then there were 26 tests instead of 27.
Is it the same behaviour? If so, do you think something easily can be done to propagate the error?

src/pybind/pyembed.cpp
```diff
@@ -0,0 +1,85 @@
...
+        logger->critical("Tried but failed to load {}", pylib_env);
+        logger->critical(errstr);
+        throw std::runtime_error("Failed to dlopen");
+    }
+    const auto pybind_wraplib_env = std::getenv("NMODL_WRAPLIB");
+    if (!pybind_wraplib_env) {
+        logger->critical(
+            "NMODL_WRAPLIB environment variable must be set to load the pybind wrapper library");
+        throw std::runtime_error("NMODL_WRAPLIB not set");
+    }
```
pramodk, line 49: For developers, may be we should document these `NMODL_PYLIB ` and `NMODL_WRAPLIB` variables in some relevant section of documentation? e.g. somewhere in INSTALL.md?

src/pybind/pyembed.hpp
```diff
@@ -0,0 +1,181 @@
...
+    // may also return a python exception message:
+    std::string exception_message;
+
+    // executor function
+    virtual void operator()() override;
+};
+
+struct AnalyticDiffExecutor: public PythonExecutor {
+    // input
+
```
pramodk, line 84: `nitpick` : extra line?

src/pybind/wrapper.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+void destroy_ads_executor_func(AnalyticDiffExecutor* exec) {
+    delete exec;
+}
+
+void initialize_interpreter_func() {
+    pybind11::initialize_interpreter(true);
+    const auto python_path_cstr = std::getenv("PYTHONPATH");
+    if (python_path_cstr) {
+        pybind11::module::import("sys").attr("path").cast<pybind11::list>().insert(
```
pramodk, line 192: just curious ! if the `PYTHONPATH` is set before launching the nmodl, it doesn't get used? i.e. it needs to be inserted like this?

src/pybind/wrapper.cpp
```diff
@@ -0,0 +1,221 @@
...
+    return new SolveLinearSystemExecutor();
+}
+
+SolveNonLinearSystemExecutor* create_nsls_executor_func() {
+    return new SolveNonLinearSystemExecutor();
+}
+
+DiffeqSolverExecutor* create_des_executor_func() {
+    return new DiffeqSolverExecutor();
+}
```
pramodk, line 166: most likely I don't have global workflow of the code but curious why it was not possible return smart ptr to avoid explicit call to `destroy*`.

src/visitors/sympy_solver_visitor.cpp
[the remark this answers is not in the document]
pramodk: Ah ok..it's bit confusing because rest of code is inside and last block outside. But its ok.

test/CMakeLists.txt
```diff
@@ -78,6 +80,18 @@ target_link_libraries(testunitparser lexer test_util config)
...
 # =============================================================================
+set(testvisitor_env "PYTHONPATH=${PROJECT_BINARY_DIR}/lib/python:$ENV{PYTHONPATH}")
+if(NOT LINK_AGAINST_PYTHON)
+  list(APPEND testvisitor_env "NMODL_PYLIB=$ENV{NMODL_PYLIB}")
+  if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+    list(APPEND testvisitor_env
+                "NMODL_WRAPLIB=${PROJECT_BINARY_DIR}/lib/python/nmodl/libpywrapper.so")
+  elseif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    list(APPEND testvisitor_env
+                "NMODL_WRAPLIB=${PROJECT_BINARY_DIR}/lib/python/nmodl/libpywrapper.dylib")
```
pramodk, line 91: Ahh..I don't need to set these manually?

src/pybind/CMakeLists.txt
ohm314: hum.. wait.. the `Python.h` won't be available when _building_ the wheel? that might become a problem. I believe pybind11 will need it to build embedded python related code!

src/pybind/CMakeLists.txt
ohm314: I forgot! fixed.

src/pybind/pyembed.cpp
ohm314: That's exactly the issue that I had explained earlier in our slack conversation. What happens is that during build, when a test executable is built, catch2 will then run it with `--list-tests`, if the execution fails for some reason that error is not propagated from the CatchAddTests cmake_command further up to `make` and the build does not fail. I believe this is a bug in `CatchAddTests.cmake`. Shall I try and fix it locally?
What is happening in your case is that the python wrapper is not found, so we proceed with trying to load lipython. If at that point we don't find the environment variable or libpython, then we throw an exception and the process terminates. So that behavior is correct, I think.

src/pybind/pyembed.cpp
ohm314: Yes, that is true. Wanted to do that but forgot.

src/pybind/wrapper.cpp
ohm314: That is correct. This because we are running an embedded python. It is up to the programmer to provide the environment to libpython. If you are running the python executable, it will do this for you.
This is a very interesting and relevant discussion on this topic: https://bugs.python.org/issue22213

pramodk: ok! thanks for the reference!

src/pybind/CMakeLists.txt
pramodk: I have built pybind11 (and part of nmodl) before on manylinux and it's not a problem. (basically, with those linker flags, we ignore undefined symbols from libpython)

src/pybind/pyembed.cpp
pramodk: >  I believe this is a bug in CatchAddTests.cmake. Shall I try and fix it locally?

I assume we will use this LINK_AGAINST_PYTHON=OFF mode while only building wheels, so I think this is not on critical path and blocker for this PR. (this improvement we can do later, we can just create an issue)

src/pybind/wrapper.cpp
ohm314: passing smart pointers through dynamically loaded libraries.. hmm.. not sure what would happen! The reason why I insist on the destructor being called from within the `.so` is because I want to make sure destructor code is definitely reachable. I don't claim I fully understand the details. But here is some relevant documentation:
http://www.faqs.org/docs/Linux-mini/C++-dlopen.html#loadingclasses
(not end of sec 3.3)

I would have liked to have all implementations of `PythonExecutor` to have the same interface rather than these structs where one must set the fields from outside, but what I cam up with was ugly.. so I deferred this for later in the interest of time.

test/CMakeLists.txt
ohm314: not for testing anymore 😄

src/pybind/CMakeLists.txt
ohm314: Just to conclude on this. We discussed offline and we agree that the python headers are indeed available, so all good there - no change required.

src/pybind/wrapper.cpp
pramodk: yup, that's fine!

Issue #340, opened 2020-05-28
Erase collection of non-consecutive nodes from vector


cattabiani:
Since this function is requested often: 

it would be cool to have a function (for every node vector) that erases an arbitrary collection of non-consecutive nodes from a vector. 

Interesting note:
since our vectors are typically vectors of pointers there is no risk of having duplicate entries since every pointer goes to a different location in memory (if we compare pointers)

cattabiani:
It is useful to simplify https://github.com/BlueBrain/nmodl/pull/230

PR #341, opened 2020-05-28
Erase collection of non-consecutive nodes from vectors (issue #340)

cattabiani:
useful for simplifying https://github.com/BlueBrain/nmodl/pull/230

### DIFF
--- a/src/language/nodes.py
+++ b/src/language/nodes.py
@@ -272,6 +272,32 @@ class ChildNode(BaseNode):
                             auto last_it = const_iter_cast({self.varname}, last);
                             return {self.varname}.erase(first_it, last_it);
                          }}
+                         /**
+                          * \\brief Erase non-consecutive members to {self.varname}
+                          * 
+                          * loosely following the cpp reference of remove_if
+                          */
+                         size_t erase_{to_snake_case(self.class_name)}(std::unordered_set<{self.class_name}*>& to_be_erased) {{
+                            auto first = {self.varname}.begin();
+                            auto last = {self.varname}.end();
+                            auto result = first;
+                        
+                            while (first != last) {{
+                                    // automatically erase dangling pointers from the uset while
+                                    // looking for them to erase them in the vector
+                                    if (to_be_erased.erase(first->get()) == 0) {{
+                                        reset_{to_snake_case(self.class_name)}(result, *first);
+                                        ++result;
+                                    }}
+                                    ++first;
+                                }}
+                            
+                            size_t out = last - result;
+                            erase_{to_snake_case(self.class_name)}(result, last);
+                            
+                            return out;
+                         }}
+                        
 
                          /**
                           * \\brief Insert member to {self.varname}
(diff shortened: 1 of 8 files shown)


src/language/nodes.py
```diff
@@ -272,6 +272,27 @@ def get_add_methods(self):
...
+                          */
+                         void erase_{to_snake_case(self.class_name)}(const std::unordered_set<{self.class_name}*>& to_be_erased) {{
+                            // loosely following the cpp reference of remove_if
+
+                            auto first = {self.varname}.begin();
+                            auto last = {self.varname}.end();
+                            auto result = first;
+                        
+                            while (first != last) {{
+                                if (to_be_erased.find(first->get()) == to_be_erased.end()) {{
```
ohm314: `<nitpick>`Unless I'm mistaken what you do here is move all the elements out of the way that should be kept, and then you delete the rest. Wouldn't be more efficient to do the opposite? Only move elements that should be deleted?

src/visitors/kinetic_block_visitor.hpp
```diff
@@ -14,8 +14,8 @@
 
 #include <map>
 #include <memory>
-#include <set>
 #include <string>
+#include <unordered_set>
```
ohm314, line 18: why the switch?

cattabiani: There was no reason to have a set. Unordered sets perform better for finding nodes (required in the routines even before my addition)

src/language/nodes.py
```diff
@@ -272,6 +272,27 @@ def get_add_methods(self):
                             auto last_it = const_iter_cast({self.varname}, last);
                             return {self.varname}.erase(first_it, last_it);
                          }}
+                         /**
+                          * \\brief Erase non-consecutive members to {self.varname}
+                          */
+                         void erase_{to_snake_case(self.class_name)}(const std::unordered_set<{self.class_name}*>& to_be_erased) {{
```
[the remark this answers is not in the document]
cattabiani: it is because the ordering is not required anywhere and set.find performs better

src/language/nodes.py
cattabiani: Example: I need to remove an element at the beginning of the vector. If I call std::erase(begin()), the code moves all the other elements at the beginning and then erases the last one. In fact [here](http://www.cplusplus.com/reference/vector/vector/erase/) is stated that the complexity is:

> Linear on the number of elements erased (destructions) plus the number of elements after the last element deleted (moving).

So, in any case I need to move the remaining elements at the beginning and then erase the back part. The algorithm that I use is very similar to the one stated for [remove_if](http://www.cplusplus.com/reference/algorithm/remove_if/) as explained in the comment

iomaganaris:
I would feel a bit safer if we had a test specifically for this function and using as an example a `StatementBlock` but apart from this looks good to me

cattabiani:
> I would feel a bit safer if we had a test specifically for this function and using as an example a StatementBlock but apart from this looks good to me

I can try

cattabiani:
I checked and the routine is already tested plenty of times in the sympy visitor tests. In the various tests it erases one, all and some of the elements of the `statements` vector with the sympy visitor. Thus, I do not think a specific test is required.

PR #343, opened 2020-05-29
Change map -> umap for improved performance

cattabiani:
There is no need to keep the items in order. Umaps perform
better.


ohm314:
I was thinking the same thing as @pramodk at the other PR, but since I didn't have time to look at it in detail, I thought that anyway at these very small number of elements, unordered or ordered map/set are the same. In general I agree with @pramodk that ordered maps (and I believe sets) are to be preferred, because they are more predictable and the constant that goes into your big-Oh is smaller.

ohm314:
Having this said, I think we can still merge this PR 😄

PR #347, opened 2020-06-08
Transform top level LOCAL variables to ASSIGNED

### LINKED ISSUE #301: LOCAL variables in global scope are not handled correctly in code generation
Consider an example in https://github.com/Neurosim-lab/netpyne/blob/coreneuron/examples/coreneuron/m1/mod/kapcb.mod

```
TITLE K-A channel from Klee Ficker and Heinemann
: modified by Brannon and Yiota Poirazi (poirazi@LNC.usc.edu)
: to account for Hoffman et al 1997 proximal region kinetics
: used only in soma and sections located < 100 microns from the soma

NEURON {
	SUFFIX kapcb
	USEION k READ ko, ki WRITE ik		:Changed from READ ek, 23/04/2010,Nassi
        RANGE gkabar, ik
        :GLOBAL ninf,linf,taul,taun,lmin
}

UNITS {
	(mA) = (milliamp)
	(mV) = (millivolt)
}


PARAMETER {                       :parameters that can be entered when function is called in cell-setup
       	gkabar = 0      (mho/cm2) :initialized conductance
        vhalfn = 11     (mV)      :activation half-potential
        vhalfl = -56    (mV) 	  :inactivation half-potential
	:vhalfl = -56    (mV) 	  :inactivation half-potential
        a0n = 0.05      (/ms)     :parameters used
        zetan = -1.5    (1)       :in calculation of (-1.5)
        zetal = 3       (1)       :steady state values(3)
        gmn = 0.55      (1)       :and time constants(0.55) change to get an effect on spike repolarization
        gml = 1         (1)
	:gml = 1         (1)
	lmin = 2        (ms)
	nmin = 0.1      (ms)
	pw = -1         (1)
	tq = -40	(mV)
	qq = 5		(mV)
	q10 = 5                   :temperature sensitivity

}

ASSIGNED {       :parameters needed to solve DE
	v               (mV)
        ek              (mV)      :K reversal potential  (reset in cell-setup.hoc)
	celsius         (degC)
	ik              (mA/cm2)
        ninf
        linf      
        taul            (ms)
        taun            (ms)
	ko		(mM)
	ki		(mM)
}

STATE {          :the unknown parameters to be solved in the DEs 
	n l
}

LOCAL qt

INITIAL {		:initialize the following parameter using rates()
        qt = q10^((celsius-24)/10(degC))         : temprature adjustment factor
	rates(v)
	n = ninf
	l = linf
}

BREAKPOINT {
	SOLVE states METHOD cnexp
	ek=25*log(ko/ki)			:Changed, added, 23/04/2010, Nassi
	ik = gkabar*n*l*(v-ek)
}

DERIVATIVE states {
	rates(v)
        n' = (ninf - n)/taun
        l' = (linf - l)/taul
}



PROCEDURE rates(v (mV)) {                  :callable from hoc
        LOCAL a
	
        a = alpn(v)
        ninf = 1/(1 + a)                   : activation variable steady state value
        taun = betn(v)/(qt*a0n*(1+a))      : activation variable time constant
	if (taun<nmin) {taun=nmin}         : time constant not allowed to be less than nmin
        
	a = alpl(v)
        linf = 1/(1+ a)                    : inactivation variable steady state value
	taul = 12 (ms)
	:taul = 0.26(ms/mV)*(v+50)               : inactivation variable time constant
	:if (taul<lmin) {taul=lmin}         : time constant not allowed to be less than lmin

}

FUNCTION alpn(v(mV)) { LOCAL zeta 
  zeta = zetan+pw/(1+exp((v-tq)/qq))
UNITSOFF
  alpn = exp(1.e-3*zeta*(v-vhalfn)*9.648e4/(8.315*(273.16+celsius))) 
UNITSON
}

FUNCTION betn(v(mV)) { LOCAL zeta
  zeta = zetan+pw/(1+exp((v-tq)/qq))
UNITSOFF
  betn = exp(1.e-3*zeta*gmn*(v-vhalfn)*9.648e4/(8.315*(273.16+celsius))) 
UNITSON
}

FUNCTION alpl(v(mV)) {
UNITSOFF
  alpl = exp(1.e-3*zetal*(v-vhalfl)*9.648e4/(8.315*(273.16+celsius))) 
UNITSON
}

FUNCTION betl(v(mV)) {
UNITSOFF
  betl = exp(1.e-3*zetal*gml*(v-vhalfl)*9.648e4/(8.315*(273.16+celsius))) 
UNITSON
}
```

And CoreNEURON compilation produces:

```
[INFO] Running make with params: ROOT=/Users/kumbhar/workarena/repos/bbp/coreneuron/build/install MODS_PATH=./examples/coreneuron/m1/mod/
 -> Generating x86_64/core/mod2c/_mod_func.cpp
 -> Generating x86_64/core/mod2c/_kinderiv.h
 -> Compiling x86_64/core/mod2c/kapcb.cpp
x86_64/core/mod2c/kapcb.cpp:304:39: error: use of undeclared identifier 'qt'
        inst->taun[id] = betn_in_0 / (qt * kapcb_global.a0n * (1.0 + a));
                                      ^
x86_64/core/mod2c/kapcb.cpp:375:17: error: use of undeclared identifier 'qt'
                qt = pow(kapcb_global.q10, ((celsius - 24.0) / 10.0));
                ^
x86_64/core/mod2c/kapcb.cpp:393:51: error: use of undeclared identifier 'qt'
                    inst->taun[id] = betn_in_0 / (qt * kapcb_global.a0n * (1.0 + a));
                                                  ^
x86_64/core/mod2c/kapcb.cpp:489:47: error: use of undeclared identifier 'qt'
                inst->taun[id] = betn_in_0 / (qt * kapcb_global.a0n * (1.0 + a));
                                              ^
4 errors generated.
make: *** [x86_64/core/build/kapcb.o] Error 1
```

This is because qt is not printed in the code.

iomaganaris:
- Top level LOCAL variables were not handled until now in NMODL
- Added visitor which checks if any of them are written and then adds them to the ASSIGNED block
- ASSIGNED variables are by default RANGE variables
- Those variables need to be RANGE since they might be written by multiple threads in INITIAL or BREAKPOINT blocks
- Fixes #301


src/visitors/local_var_visitor.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    const auto& top_level_nodes = node.get_blocks();
+    const auto& symbol_table = node.get_symbol_table();
+
+    for (auto& top_level_node: top_level_nodes) {
+        /// only process local_list statements
+        if (top_level_node->is_local_list_statement()) {
+            const auto& local_variables =
+                std::static_pointer_cast<ast::LocalListStatement>(top_level_node)->get_variables();
+            for (auto& local_variable: local_variables) {
```
cattabiani, line 38: const auto& could be clearer

src/language/templates/ast/all.hpp
```diff
@@ -20,6 +20,9 @@
  */
 
 #include "ast/ast.hpp"
+{% for header in node_info.CIRCULAR_DEPENDECIES %}
```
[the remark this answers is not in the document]
iomaganaris: Without this this is the issue I get:
```
In file included from /Users/magkanar/nmodl_local/src/parser/nmodl_driver.cpp:11:
In file included from /Users/magkanar/nmodl_local/src/lexer/nmodl_lexer.hpp:16:
In file included from /Users/magkanar/nmodl_local/src/parser/nmodl.yy:31:
/Users/magkanar/nmodl_local/build/src/ast/all.hpp:7233:10: error: member access into incomplete type 'nmodl::ast::AssignedDefinition'
        n->set_parent(this);
         ^
```
I believe that the issue is not with forward declaring the class, since this is done in `ast_decl.hpp` but the method `AssignedDefinition::set_parent()`. To solve this issue we should forward declare all the methods then, which I am not convinced that is a better option

src/visitors/visitor_utils.cpp
```diff
@@ -35,6 +35,11 @@ std::string get_new_name(const std::string& name,
     return (name + "_" + suffix + "_" + std::to_string(counter));
 }
 
+std::shared_ptr<ast::Name> get_name_from_string(const std::string& name) {
```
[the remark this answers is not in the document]
iomaganaris: True. It's not necessary. I had added it because this was needed in a commit but then removed the related code. We can remove it to make it clearer but maybe it could be used at some point in the future. As you wish

src/visitors/local_var_visitor.cpp
```diff
@@ -0,0 +1,84 @@
...
+                auto variable_name = local_variable->get_node_name();
+                /// check if local variable is being updated in the mod file
+                if (symbol_table->lookup(variable_name)->get_write_count() > 0) {
+                    assigned_variables.emplace_back(
+                        new ast::AssignedDefinition(local_variable->get_name(),
```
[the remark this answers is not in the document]
iomaganaris: You're right. Good catch 👍

src/visitors/local_var_visitor.cpp
iomaganaris: I think that we had a similar conversation in the `GlobalToRange` PR. @pramodk asked to remove the unnecessary `const` declaration and since some functions return `const ref`s I should maybe remove a couple of `const` from the code

src/language/templates/ast/all.hpp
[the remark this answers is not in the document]
ohm314: Since @tristan0x was working on these things, I really would like to hear his opinion too?

src/visitors/local_var_visitor.cpp
ohm314: You're right.. one just has to remember that these functions actually return const refs, that's a bit the annoying part. I have no clear preference but wonder what people typically do...

src/visitors/local_var_visitor.cpp
```diff
@@ -0,0 +1,84 @@
...
+                /// check if local variable is being updated in the mod file
+                if (symbol_table->lookup(variable_name)->get_write_count() > 0) {
+                    assigned_variables.emplace_back(
+                        std::make_shared<ast::AssignedDefinition>(local_variable->get_name(),
+                                                                  nullptr,
+                                                                  nullptr,
+                                                                  nullptr,
+                                                                  nullptr,
+                                                                  nullptr,
+                                                                  nullptr));
```
ohm314, line 49: I think we are missing some default initialization in this constructor...

src/visitors/local_var_visitor.cpp
```diff
@@ -0,0 +1,84 @@
...
+                    local_variables_to_remove.emplace(local_variable.get());
```
ohm314, line 50:
```suggestion
                    local_variables_to_remove.insert(local_variable.get());
```
we are not constructing this, right?

src/visitors/local_var_visitor.cpp
```diff
@@ -0,0 +1,84 @@
...
+                }
+            }
+
+            /// remove offending local variables
+            std::static_pointer_cast<ast::LocalListStatement>(top_level_node)
+                ->erase_local_var(local_variables_to_remove);
+
+            /// add empty local statements to local_statements_to_remove
+            if (local_variables.empty()) {
```
ohm314, line 59: I'm curious.. once you have processed all the variables in the top-level `LOCAL` statement, dont' you expect `local_variables` to be always empty? the whole aim is to move these variables out of top level. So my question is, what if it's not empty, in which case does this happen and what do we do then?

test/visitor/local_to_assigned.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return ast;
+}
+
+SCENARIO("LOCAL to ASSIGNED variable transformer", "[visitor][localtoassigned]") {
+    GIVEN("mod file with LOCAL variables that are written") {
+        std::string input_nmodl = R"(
+            NEURON {
+                SUFFIX test
+            }
+            LOCAL x, y, z
```
ohm314, line 49: question: is it legal in NMODL to have a variable with same name in LOCAL and in the NEURON block?

src/visitors/visitor_utils.cpp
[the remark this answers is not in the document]
tristan0x: If it is not used, then it is dead code. At least, it should be called/validated in the tests.

src/language/templates/ast/all.hpp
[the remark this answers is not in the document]
tristan0x: Moving the definitions of these member methods to `ast.cpp` fixes most of the compilation issues but for the template member function 

```c++
/**
 * \brief Insert members to statements
 */
template <class NodeType, class InputIterator>
void StatementBlock::insert_statement(StatementVector::const_iterator position,
                                      NodeType& to,
                                      InputIterator first,
                                      InputIterator last) {
    for (auto it = first; it != last; ++it) {
        auto& n = *it;
        // set parents
        n->set_parent(this);
    }
    auto pos_it = const_iter_cast(statements, position);
    auto first_it = const_iter_cast(to, first);
    auto last_it = const_iter_cast(to, last);
    statements.insert(pos_it, first_it, last_it);
}
```

Possible solutions:
1. put the definition of this member method after the definitions of all the Ast classes in `ast/all.hpp`.
1. Put its definition in `ast/ast.cpp` and use explicit template instantiations for all possible cases
1. Question the usefulness of this member function:
    1. `const_iter_cast` is a workaround for an issue that we might not have anymore (see definition of this helper function)
    1. `first` and `last` are always iterators over a subclass of statements. There should be an easy way to express the same thing without using a template.

src/visitors/visitor_utils.cpp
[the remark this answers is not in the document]
iomaganaris: I didn't think of it this way Tristan. You're right. This should go away 👍

src/visitors/local_var_visitor.cpp
iomaganaris: >I have no clear preference but wonder what people typically do...

I had exactly the same question but I couldn't find any clear answer on the webz. I will keep looking though

src/language/code_generator.cmake
```diff
@@ -191,6 +191,14 @@ set(AST_GENERATED_SOURCES
     ${PROJECT_BINARY_DIR}/src/ast/write_ion_var.hpp
 )
 
+set(PYBIND_GENERATED_SOURCES
```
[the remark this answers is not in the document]
iomaganaris: I created an issue about that ( #350 ) and I think that we can tackle this in another PR, if it's okay

src/language/templates/ast/all.hpp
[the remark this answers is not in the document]
iomaganaris: Thanks a lot for your input @tristan0x !
1. Unfortunately this is not possible as fas as I know because this would mean moving all the `Statement`s on top of `Expression`s, which would fail again due to other circular dependencies
2. The definition of this for every case in `ast/ast.cpp` is something I would like to avoid, since I don't believe it's very nice
3. This might be a proper solution and I need to investigate it.
Thanks a lot again!

src/visitors/local_var_visitor.cpp
iomaganaris: Normally all the top level `LOCAL` variables should be written, so we remove them from the `LocalListStatement` which should be empty and removed from the `node`.
The only case that `local_variables` might not be empty is if one of the defined `LOCAL` variables is not written, which is something that probably won't exist but in case a variable is defined but not used, I still leave the `LOCAL` declaration there instead of deleting it.
Do you agree with this policy or do you think that it would be better to remove the `LOCAL` statement completely in this case as well?

test/visitor/local_to_assigned.cpp
iomaganaris: It's possible yes, however the variable with this visitor should be defined as `RANGE` otherwise the `CodegenCompatibilityVisitor` will complain with:
```
[NMODL] [error] :: Code Incompatibility :: "x" variable found at [3.8] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations
```

src/visitors/local_var_visitor.cpp
iomaganaris: I believe that for this case it will be the same but I can use the `insert` for more expressiveness 👍

src/visitors/local_var_visitor.cpp
iomaganaris: True. I am not sure though for which classes we should add a default constructor. Should it be all of them? We can discuss this maybe tomorrow

src/visitors/local_var_visitor.cpp
iomaganaris: I still didn't find any solid answer on the internet. It seems that there is absolutely no functional difference between the two. It's just a matter of preference.
To me it looks more readable with the `const` it front as well. Maybe this is a policy we should or shouldn't follow? 
I would also like to hear @tristan0x and @pramodk opinion on this

tristan0x: Given a C++ class exposing a non-modifiable member variable through a member function, if its API evolves and now needs to allow the member variable to be modified, then **it should provide an additional member function** rather than removing the `const` specifier in the first one.

For instance:

```c++
class LocalListStatement {
  const LocalVarVector& get_variables() const;
  LocalVarVector& get_variables();
};
```

tristan0x: In this case, I am much more bothered by the instruction below than the `const-ness` of the loop.

```
const auto& local_variables =
                std::static_pointer_cast<ast::LocalListStatement>(top_level_node)->get_variables();
```

1. The very same call to `std::static_pointer_case` is made 10 lines below, so the result shared_ptr should be stored in a local variable.
1. It lacks clarity

tristan0x: If the code was manipulating a  `std::static_pointer_cast<const ast::LocalListStatement>`, then IMO I would not be so strict at repeating `const` everywhere afterward.

Buts since the result of `std::static_pointer_cast` returns a shared_ptr of a modifiable `ast::LocalListStatement`, IMO code manipulating it should indicate whether it is modifying it or not.

So I agree with @cattabiani that the range-based loop should indicate it is not modifying `local_variable`.

tristan0x: According to my comments, here is what I would change:

```diff
diff --git a/src/visitors/local_var_visitor.cpp b/src/visitors/local_var_visitor.cpp
index ef4dd74..f2b2d42 100644
--- a/src/visitors/local_var_visitor.cpp
+++ b/src/visitors/local_var_visitor.cpp
@@ -30,12 +30,11 @@ void LocalToAssignedVisitor::visit_program(ast::Program& node) {
     auto& top_level_nodes = node.get_blocks();
     const auto& symbol_table = node.get_symbol_table();

-    for (auto& top_level_node: top_level_nodes) {
+    for (const auto& top_level_node: top_level_nodes) {
         /// only process local_list statements
         if (top_level_node->is_local_list_statement()) {
-            auto& local_variables =
-                std::static_pointer_cast<ast::LocalListStatement>(top_level_node)->get_variables();
-            for (auto& local_variable: local_variables) {
+            const auto& node = std::static_pointer_cast<ast::LocalListStatement>(top_level_node);
+            for (const auto& local_variable: node->get_variables()) {
                 auto variable_name = local_variable->get_node_name();
                 /// check if local variable is being updated in the mod file
                 if (symbol_table->lookup(variable_name)->get_write_count() > 0) {
@@ -52,11 +51,10 @@ void LocalToAssignedVisitor::visit_program(ast::Program& node) {
             }

             /// remove offending local variables
-            std::static_pointer_cast<ast::LocalListStatement>(top_level_node)
-                ->erase_local_var(local_variables_to_remove);
+            node->erase_local_var(local_variables_to_remove);

             /// add empty local statements to local_statements_to_remove
-            if (local_variables.empty()) {
+            if (node->get_variables().empty()) {
                 local_nodes_to_remove.emplace(top_level_node.get());
             }
         }
```

iomaganaris: I see.. I agree with @tristan0x points and with that `const` should be added in the `for-loop` iterators. Tristan's suggestion looks nice and clean to me, so I am going to adopt it. Thank you Tristan! 🙂

src/visitors/local_var_visitor.cpp
ohm314: Maybe one can get rid of it completely? @pramodk what do you think?

src/visitors/local_var_visitor.cpp
pramodk: Just out of curiosity, what is about using forms like:

```cpp
     const auto& top_level_nodes = node.get_blocks();
     for (auto& top_level_node: top_level_nodes) {
```
i.e. not using `const` in for loop iterator as your var is already const.

src/visitors/local_var_visitor.cpp
pramodk: yup!  if it's empty then it's being removed now. So all good here.

src/visitors/local_var_visitor.cpp
pramodk: The default arguments to constructors of AST classes is quite involved/convoluted because - the order of argument is determined by the grammar definition in bison specification. So in this PR I wouldn't change those arguments.

src/language/templates/ast/all.hpp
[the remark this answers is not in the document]
pramodk: fixed in #351

PR #348, opened 2020-06-11
Rename variables that match with ISPC double constants namings

### LINKED ISSUE #302: ISPC doesn't support d[0-9]+ as global variable
Consider example https://github.com/Neurosim-lab/netpyne/blob/coreneuron/examples/coreneuron/m1/mod/cagk.mod

```
TITLE CaGk
: Calcium activated K channel.
: Modified from Moczydlowski and Latorre (1983) J. Gen. Physiol. 82

UNITS {
	(molar) = (1/liter)
}

UNITS {
	(mV) =	(millivolt)
	(mA) =	(milliamp)
	(mM) =	(millimolar)
}


NEURON {
	SUFFIX cagk
	USEION ca READ cai
	USEION k READ ek WRITE ik
	RANGE gbar,gkca,ik
        RANGE oinf, tau
}

UNITS {
	FARADAY = (faraday)  (kilocoulombs)
	R = 8.313424 (joule/degC)
}

PARAMETER {
	celsius		(degC)
	v		(mV)
	gbar=.01	(mho/cm2)	: Maximum Permeability
	cai 		(mM)
	ek		(mV)

	d1 = .84
	d2 = 1.
	k1 = .48e-3	(mM)
	k2 = .13e-6	(mM)
	abar = .28	(/ms)
	bbar = .48	(/ms)
        st=1            (1)
}

ASSIGNED {
	ik		(mA/cm2)
	oinf
	tau		(ms)
        gkca          (mho/cm2)
}

INITIAL {
        rate(v,cai)
        o=oinf
}

STATE {	o }		: fraction of open channels

BREAKPOINT {
	SOLVE state METHOD cnexp
	gkca = gbar*o^st
	ik = gkca*(v - ek)
}

DERIVATIVE state {	: exact when v held constant; integrates over dt step
	rate(v, cai)
	o' = (oinf - o)/tau
}
FUNCTION alp(v (mV), c (mM)) (1/ms) { :callable from hoc
	alp = c*abar/(c + exp1(k1,d1,v))
}
FUNCTION bet(v (mV), c (mM)) (1/ms) { :callable from hoc
	bet = bbar/(1 + c/exp1(k2,d2,v))
}
FUNCTION exp1(k (mM), d, v (mV)) (mM) { :callable from hoc
	exp1 = k*exp(-2*d*FARADAY*v/R/(273.15 + celsius))
}
PROCEDURE rate(v (mV), c (mM)) { :callable from hoc
	LOCAL a
	a = alp(v,c)
	tau = 1/(a + bet(v, c))
	oinf = a*tau
}
```

and compiling with ISPC gives:

```
Warning: No --target specified on command-line. Using default system target "avx2-i32x8".
[NMODL] [info] :: Processing examples/coreneuron/m1/mod//vecstim.mod
[NMODL] [info] :: Running symtab visitor
[NMODL] [info] :: Running code compatibility checker
[NMODL] [error] :: Code incompatibility detected
[NMODL] [error] :: Cannot translate mod file to .cpp file
[NMODL] [error] :: Fix the following errors and try again
[NMODL] [error] :: Code Incompatibility :: DESTRUCTOR construct found at [67.1-74.1] is not handled
make: *** [x86_64/core/mod2c/vecstim.cpp] Error 1
make: *** Waiting for unfinished jobs....
x86_64/core/mod2c/cagk.ispc:100:41: Error: syntax error, unexpected double constant, expecting ',' or ';'.
                    d_in_0 = cagk_global.d1_;
                                        ^^^

x86_64/core/mod2c/cagk.ispc:114:41: Error: syntax error, unexpected double constant, expecting ',' or ';'.
                    d_in_1 = cagk_global.d2_;
                                        ^^^

x86_64/core/mod2c/cagk.ispc:194:41: Error: syntax error, unexpected double constant, expecting ',' or ';'.
                    d_in_0 = cagk_global.d1_;
                                        ^^^

x86_64/core/mod2c/cagk.ispc:208:41: Error: syntax error, unexpected double constant, expecting ',' or ';'.
                    d_in_1 = cagk_global.d2_;
                                        ^^^
```

### LINKED ISSUE #329: Issue with variable definition for ISPC backend
During compilation of https://github.com/Neurosim-lab/netpyne/blob/coreneuron/examples/coreneuron/m1/mod/kctin.mod there is a wrong definition of `dt_saved_value` from the generated code as `double` instead of `double uniform`.

```
x86_64/core/mod2c/kctin.ispc:142:19: Error: Can't convert from type "varying double" to type "uniform double" for
        assignment operator.
        nt->_dt = dt_saved_value;
                  ^^^^^^^^^^^^^^
```

iomaganaris:
- Fixes #302


src/language/code_generator.cmake
```diff
@@ -207,14 +215,6 @@ set(VISITORS_GENERATED_SOURCES
     ${PROJECT_BINARY_DIR}/src/visitors/visitor.hpp
 )
 
-set(PYBIND_GENERATED_SOURCES
-    ${PROJECT_BINARY_DIR}/src/pybind/pyast.cpp
-    ${PROJECT_BINARY_DIR}/src/pybind/pyast.hpp
-    ${PROJECT_BINARY_DIR}/src/pybind/pysymtab.cpp
-    ${PROJECT_BINARY_DIR}/src/pybind/pyvisitor.cpp
-    ${PROJECT_BINARY_DIR}/src/pybind/pyvisitor.hpp
-)
```
pramodk, line 216: by the way, there is something wrong with cmake-format that constantly reorder this, have seen something with @cattabiani  as well. Something with the cmake-format version?

src/nmodl/main.cpp
```diff
@@ -310,6 +311,13 @@ int main(int argc, const char* argv[]) {
             SymtabVisitor(update_symtab).visit_program(*ast);
         }
 
+        /// Rename variables that match ISPC compiler double constants
+        if (ispc_backend) {
+            IspcRenameVisitor(ast).visit_program(*ast);
+            SymtabVisitor(update_symtab).visit_program(*ast);
+            ast_to_nmodl(*ast, filepath("ispcdoublerename"));
```
pramodk, line 318:
```suggestion
            ast_to_nmodl(*ast, filepath("ispc_double_rename"));
```

src/visitors/ispc_rename_visitor.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace visitor {
+
+/**
+ * @addtogroup visitor_classes
+ * @{
+ */
+
+/**
+ * \class IspcRenameVisitor
+ * \brief Rename variable names to match ISPC standards
```
pramodk, line 31:
```suggestion
 * \brief Rename variable names to match with ISPC backend requirement
```

src/visitors/ispc_rename_visitor.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (std::regex_match(name, double_regex)) {
+        auto& value = node.get_value();
+        const auto& vars = get_global_vars(*ast);
+        value->set(suffix_random_string(vars, new_var_name_prefix + name));
+    }
+}
+
+/** Prime name has member order which is an integer. In theory
+ * integer could be "macro name" and hence could end-up renaming
+ * macro. In practice this won't be an issue as we order is set
```
pramodk, line 30:
```suggestion
 * macro. In practice this won't be an issue as the order is set
```
we -> the?

src/visitors/ispc_rename_visitor.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace nmodl {
+namespace visitor {
+
+/// rename matching variable
+void IspcRenameVisitor::visit_name(ast::Name& node) {
+    const auto& name = node.get_node_name();
+    if (std::regex_match(name, double_regex)) {
+        auto& value = node.get_value();
+        const auto& vars = get_global_vars(*ast);
+        value->set(suffix_random_string(vars, new_var_name_prefix + name));
```
pramodk, line 24: We should show warning from logger here because user might be accessing these variables from neuron and renaming those would be a problem. With warning, at least they will be aware (if they read!) .

src/visitors/ispc_rename_visitor.cpp
```diff
@@ -0,0 +1,64 @@
...
+ * Lesser General Public License. See top-level LICENSE file for details.
+ *************************************************************************/
+
+#include "visitors/ispc_rename_visitor.hpp"
+
+#include "ast/all.hpp"
+#include "parser/c11_driver.hpp"
+#include "visitors/visitor_utils.hpp"
+
+
```
pramodk, line 14: one general question here: why we re-use RenameVisitor here? i.e. I can following possibilities:
* In this visitor only find problematic variable and then invoke RenameVisitor
* Look into symbol table for all variable names, find out problematic ones and then invoke RenameVisitor

src/language/code_generator.cmake
ohm314: and same happened in my PR.
Also "there is something wrong with cmake-format" is a tautology...

src/visitors/ispc_rename_visitor.cpp
ohm314: I had exactly the same question...

src/visitors/ispc_rename_visitor.cpp
ohm314: you know.. when someone does some `VERBATIM` ... 😄

ohm314: also do we need the random suffix here? the names are already unique, just not ispc compatible, the prefix should already take care of that.

src/visitors/ispc_rename_visitor.cpp
iomaganaris: The reason why I didn't reuse the `RenameVisitor` is that this would mean two passes on the same variables. One to find which need changing and one to change them with the `RenameVisitor`. What do you think about that?

src/visitors/ispc_rename_visitor.cpp
iomaganaris: @pramodk you're absolutely right about the log for the renaming
@ohm314 I added the random suffix just in case there is already a `var_d1` variable already defined. There is a bug the way I do this now though I think which I need to fix

ohm314: Yes of course! actually now I remember when I did the renaming stuff I did the same, sorry!

src/language/code_generator.cmake
iomaganaris: I have noted this issue as well before. I create an issue for this https://github.com/BlueBrain/nmodl/issues/350

src/visitors/ispc_rename_visitor.cpp
```diff
@@ -0,0 +1,91 @@
...
+    driver.scan_string(text);
+    auto tokens = driver.all_tokens();
+
+    std::string result;
+    for (auto& token: tokens) {
+        if (std::regex_match(token, double_regex)) {
+            /// Check if variable is already renamed and use the same naming otherwise add the
+            /// new_name to the renamed_variables map
+            std::string new_name;
+            if (renamed_variables.find(token) != renamed_variables.end()) {
```
cattabiani, line 72: Here you do the search twice. I suggest something like this (feel free to change variable names):
const auto rename_variables_it = renamed_variables.find(token);
if (rename_variables_it != renamed_variables.end()) {
new_name = renamed_variables_it->second;
}

src/visitors/ispc_rename_visitor.cpp
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    const auto& statement = node.get_statement();
+    auto text = statement->eval();
+    parser::CDriver driver;
+
+    driver.scan_string(text);
+    auto tokens = driver.all_tokens();
+
+    std::string result;
```
cattabiani, line 66: maybe a std::stringstream is better for concatenating

src/visitors/ispc_rename_visitor.hpp
```diff
@@ -0,0 +1,79 @@
...
+    /// regex that matches double constant expressions
+    const std::regex double_regex = std::move(
+        std::regex("([0-9\\.]*d[\\-0-9]+)|([0-9\\.]+d[\\-0-9]*)"));
+
+    /// new name
+    const std::string new_var_name_prefix = "var_";
+
+    /// Map that keeps the renamed variables to keep the same random suffix when a variable is
+    /// renamed accross the whole file
+    std::unordered_map<std::string, std::string> renamed_variables;
```
cattabiani, line 54: Do you need it ordered? What about an unordered_map?

src/visitors/sympy_solver_visitor.cpp
```diff
@@ -155,31 +154,16 @@ std::vector<std::string> SympySolverVisitor::filter_string_vector(
     return filtered_vector;
 }
 
-std::string SympySolverVisitor::suffix_random_string(const std::string& original_string) const {
```
cattabiani, line 158: wow, I did not notice this was here. GJ moving it away!

src/visitors/ispc_rename_visitor.hpp
iomaganaris: It's actually an unordered map but maybe I should also state it in the comment 👍

src/visitors/ispc_rename_visitor.cpp
iomaganaris: I don't think there is any gain using a `std::stringstream` and since we would have to return it as a `string` in the end I would prefer to keep this simple

src/visitors/ispc_rename_visitor.hpp
cattabiani: wow, I am blind. Sorry

src/visitors/ispc_rename_visitor.cpp
cattabiani: this is a search:

renamed_variables.find(token)

and also this:

new_name = renamed_variables[token];

src/visitors/ispc_rename_visitor.cpp
cattabiani: because every time you add stuff to a string you need to allocate space and copy the stuff in the new location. For stringstream it just allocates a new chunk and creates a link between the old string and the new one (I think). do you want to disproof me on godbolt?

src/visitors/ispc_rename_visitor.cpp
iomaganaris: `renamed_variables` is an `unordered_map`, so average search (first case) and access (second case) complexity is `O(1)`. For the sake of code simplicity I would keep it as it is, since then the `rename_variables_it` wouldn't be used in the other branch of the `if-else`

src/visitors/ispc_rename_visitor.cpp
iomaganaris: >For stringstream it just allocates a new chunk and creates a link between the old string and the new one 

Tbh I didn't now how `stringstream` worked underneath and I based my previous answer to some `stack overflow` answers, since this is not a very important part of the code. However since you mention this I wanted to learn more about it.
First I looked how `std::stringstream` and `std::string` insert and append strings. The difference is an append (with an allocation) for the `std::string` (https://en.cppreference.com/w/cpp/string/basic_string/operator%2B%3D) vs an `insert` for `std::stringstream` (https://en.cppreference.com/w/cpp/io/basic_ostream/operator_ltlt) .
Then I created a simple benchmark in (quick-bench.com) and checked the produced assembly:
```
static void StringAppendingStr(benchmark::State& state) {
  // Code inside this loop is measured repeatedly
  std::string str;
  std::string new_str = "stringstring";
  for (auto _ : state) {
    for (int i = 0; i < 5; ++i) {
        str += new_str;
    }
    //const auto& ret_str = str;
    // Make sure the variable is not optimized away by compiler
    //benchmark::DoNotOptimize(created_string);
  }
}
// Register the function as a benchmark
BENCHMARK(StringAppendingStr);
#include <sstream>
static void StringAppendingStrStrm(benchmark::State& state) {
  std::stringstream str_strm;
  std::string new_str = "stringstring";
  //std::string ret_str;
  for (auto _ : state) {
    for (int i = 0; i < 5; ++i) {
        str_strm << new_str;
    }
    //const auto& ret_str = str_strm.str();
    //benchmark::DoNotOptimize(ret_str);
  }
}
BENCHMARK(StringAppendingStrStrm);
```
By only appending to the string the `std::string` is only x1.2 times faster than the `std::stringstream`, however when you call the `str()` function to return a string (until c++20) it does a copy of the internal string and it makes the call very costly (https://en.cppreference.com/w/cpp/io/basic_stringstream/str).
I believe we can leave the `std::string` as it is then.

src/visitors/ispc_rename_visitor.cpp
iomaganaris: @cattabiani There is already a conversation here regarding editing the `RenameVisitor` and the separate `IspcRenameVisitor`. I think that creating a new visitor is okay to avoid having making the `RenameVisitor` more complex for something that is related only for the `ispc` backend. If you guys think otherwise I can change it

cattabiani: ok!

src/visitors/ispc_rename_visitor.cpp
cattabiani: interesting but this example does not really test. I agree that append to a string is faster for short/not so many appends. I think stringstream is faster for longer strings/more frequent appends. I did not know the threshold where stringstream takes over. 

Everything said, since this is your usual test case, I guess strings are better.

iomaganaris: You're right, it depends on the number of appends and the sizes of the strings👍  Due to the call the `str()` in the end I believe we can avoid the `stringstream` for this case. Otherwise it's pretty cool 👍

src/visitors/ispc_rename_visitor.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * \class IspcRenameVisitor
+ * \brief Rename variable names to match with ISPC backend requirement
+ *
+ * Here are some examples of double constants that
+ * ISPC compiler parses: 3.14d, 31.4d-1, 1.d, 1.0d,
+ * 1d-2. Rename variables based on regex that matches
+ * those presentations.
+ *
```
pramodk, line 38: Similar to other visitors, we should expand the documentation bit here.

src/visitors/ispc_rename_visitor.hpp
```diff
@@ -0,0 +1,80 @@
...
+ *              SUFFIX test
+ *              RANGE d1, d2, var_d3
+ *          }
+ *          INITIAL {
+ *               d1 = 1
+ *               d2 = 2
+ *               var_d3 = 3
+ *          }
+ *          PROCEDURE func () {
+ *          VERBATIM d4 = 4 ENDVERBATIM
```
pramodk, line 55:
```suggestion
 *          VERBATIM d4 = 4; ENDVERBATIM
```

src/visitors/rename_visitor.cpp
```diff
@@ -49,8 +77,33 @@ void RenameVisitor::visit_verbatim(ast::Verbatim& node) {
...
+                if (add_prefix) {
+                    new_name = new_var_name_prefix + token;
+                } else {
+                    new_name = new_var_name;
+                }
+            }
+            result += new_name;
+            logger->warn("RenameVisitor :: Renaming variable {} in VERBATIM block to {}",
+                         token,
+                         new_name);
```
pramodk, line 106: this is quite duplicated above at line 27 ? can be common function?

azure-pipelines.yml
```diff
@@ -57,6 +64,23 @@ jobs:
...
+      cmake .. -DCORENRN_ENABLE_MPI=OFF -DCORENRN_ENABLE_NMODL=ON -DCORENRN_NMODL_DIR=$HOME/nmodl -DPYTHON_EXECUTABLE=$(which python3.7) -DCORENRN_NMODL_FLAGS="sympy --analytic" -DCORENRN_ENABLE_ISPC=ON -DCMAKE_ISPC_COMPILER=$(Build.Repository.LocalPath)/$CMAKE_PKG/ispc/bin/ispc
+      make -j 2
+      if [ $? -ne 0 ]
+      then
+        make VERBOSE=1
+      fi
+      ./bin/nrnivmodl-core $(Build.Repository.LocalPath)/test/integration/mod
+    env:
+      CMAKE_PKG: 'cmake-3.10.2-Linux-x86_64'
+    displayName: 'Build and Run Integration Tests with ISPC compiler'
```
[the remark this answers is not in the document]
iomaganaris: You mean throw away installing `CoreNEURON` with `NMODL` in the travis CI?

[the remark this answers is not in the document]
pramodk: up to you : if it's completely duplicate , can be removed?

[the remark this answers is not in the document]
iomaganaris: Yes, I think you're right

PR #351, opened 2020-06-18
Ast generated code: fix template members / move definitions in compilation unit

tristan0x:
Ast generated code: fix template members / move definitions in compilation unit

* In `ast/all.hpp` the definition of template member functions are
now after the definition of all Ast classes.
* The definition of some member methods has been moved in `ast/ast.cpp`
compilation units instead of inlining them.

### DIFF
--- a/src/language/templates/ast/all.hpp
+++ b/src/language/templates/ast/all.hpp
@@ -24,7 +24,9 @@
 {% for node in nodes %}
 #ifndef {{ node.cpp_fence }}
 #define {{ node.cpp_fence }}
+{% if node.has_template_methods %}
 #define {{ node.cpp_fence }}_INLINE_DEFINITION_REQUIRED
+{% endif %}
 {% include "ast/node_class.template" %}
 #endif // !{{ node.cpp_fence }}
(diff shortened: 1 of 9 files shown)


src/language/templates/ast/all.hpp
```diff
@@ -20,13 +20,25 @@
...
 
 #include "ast/ast.hpp"
-{% for header in node_info.CIRCULAR_DEPENDECIES %}
-#include "ast/{{header}}.hpp"
-{% endfor %}
 
 {% for node in nodes %}
 #ifndef {{ node.cpp_fence }}
 #define {{ node.cpp_fence }}
+#define {{ node.cpp_fence }}_INLINE_DEFINITION_REQUIRED
```
iomaganaris, line 27: This could be added only `if node.has_template_methods`?

tristan0x: Absolutely, there is no need to pollute with this macro if there is not inline definitions below.

pramodk:
Doing `--force-with-lease` to make it up to date. Hope I don't break anything!

PR #346, opened 2020-06-05
Allow the same variable in read and write in USEION

cattabiani:
Fix issue #215

PR #353, opened 2020-06-23
Fix declaration of dt_saved_value in ISPC backend

### LINKED ISSUE #329: Issue with variable definition for ISPC backend

iomaganaris:
- `dt_saved_value` should be defined as uniform in the ISPC backend while other variables should be varying (simply double)
- Fixes #329

### DIFF
--- a/src/codegen/codegen_ispc_visitor.cpp
+++ b/src/codegen/codegen_ispc_visitor.cpp
@@ -67,22 +67,27 @@ void CodegenIspcVisitor::visit_local_list_statement(ast::LocalListStatement& nod
     if (!codegen) {
         return;
     }
+    /// Name of the variable _dt given by
+    /// nmodl::visitor::SteadystateVisitor::create_steadystate_block()
+    const std::string steadystate_dt_variable_name = "dt_saved_value";
     auto type = CodegenCVisitor::local_var_type() + " ";
     printer->add_text(type);
     auto local_variables = node.get_variables();
-    std::set<std::string> uniform_variables;
-    auto dt_saved_value_exists = false;
+    bool dt_saved_value_exists = false;
 
     /// Remove dt_saved_value from local_variables if it exists
-    for (auto it_local_variables = local_variables.begin();
-         it_local_variables < local_variables.end();
-         ++it_local_variables) {
-        if ((*it_local_variables)->get_node_name() == "dt_saved_value") {
-            local_variables.erase(it_local_variables);
-            dt_saved_value_exists = true;
-            break;
-        }
-    }
+    local_variables.erase(std::remove_if(local_variables.begin(),
+                                         local_variables.end(),
+                                         [&](const std::shared_ptr<ast::LocalVar>& local_variable) {
+                                             if (local_variable->get_node_name() ==
+                                                 steadystate_dt_variable_name) {
+                                                 dt_saved_value_exists = true;
+                                                 return true;
+                                             } else {
+                                                 return false;
+                                             }
+                                         }),
+                          local_variables.end());
 
     /// Print the local_variables like normally
     CodegenCVisitor::print_vector_elements(local_variables, ", ");
@@ -93,7 +98,7 @@ void CodegenIspcVisitor::visit_local_list_statement(ast::LocalListStatement& nod
         printer->add_newline();
 
         type = CodegenCVisitor::local_var_type() + " uniform ";
-        printer->add_text(type + "dt_saved_value");
+        printer->add_text(type + steadystate_dt_variable_name);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -63,6 +63,39 @@ void CodegenIspcVisitor::visit_var_name(ast::VarName& node) {
...
 
+void CodegenIspcVisitor::visit_local_list_statement(ast::LocalListStatement& node) {
+    if (!codegen) {
+        return;
+    }
+    auto type = CodegenCVisitor::local_var_type() + " ";
+    printer->add_text(type);
+    auto local_variables = node.get_variables();
+    std::set<std::string> uniform_variables;
+    auto dt_saved_value_exists = false;
```
tristan0x, line 74:
```suggestion
    bool dt_saved_value_exists = false;
```

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -63,6 +63,39 @@ void CodegenIspcVisitor::visit_var_name(ast::VarName& node) {
...
+
+    /// Remove dt_saved_value from local_variables if it exists
+    for (auto it_local_variables = local_variables.begin();
+         it_local_variables < local_variables.end();
+         ++it_local_variables) {
+        if ((*it_local_variables)->get_node_name() == "dt_saved_value") {
+            local_variables.erase(it_local_variables);
```
tristan0x, line 81: `std::vector<T>::erase` invalidates iterators.

```suggestion
            it_local_variables = local_variables.erase(it_local_variables);
```

I would replace the for loop with a `std::remove_if` though.

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -63,6 +63,39 @@ void CodegenIspcVisitor::visit_var_name(ast::VarName& node) {
...
+    printer->add_text(type);
+    auto local_variables = node.get_variables();
+    std::set<std::string> uniform_variables;
+    auto dt_saved_value_exists = false;
+
+    /// Remove dt_saved_value from local_variables if it exists
+    for (auto it_local_variables = local_variables.begin();
+         it_local_variables < local_variables.end();
+         ++it_local_variables) {
+        if ((*it_local_variables)->get_node_name() == "dt_saved_value") {
```
tristan0x, line 80: I would use a constant instead of the hardcoded string `dt_saved_value`.

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -63,6 +63,39 @@ void CodegenIspcVisitor::visit_var_name(ast::VarName& node) {
...
 }
 
+void CodegenIspcVisitor::visit_local_list_statement(ast::LocalListStatement& node) {
+    if (!codegen) {
+        return;
+    }
+    auto type = CodegenCVisitor::local_var_type() + " ";
+    printer->add_text(type);
+    auto local_variables = node.get_variables();
+    std::set<std::string> uniform_variables;
```
tristan0x, line 73: `uniform_variables` variable seems unused

src/codegen/codegen_ispc_visitor.cpp
iomaganaris: Good point. I added it as a constant and some documentation on how this name is generated 👍

src/codegen/codegen_ispc_visitor.cpp
iomaganaris: True. I missed it 😞

src/codegen/codegen_ispc_visitor.cpp
iomaganaris: You're right. Fixed that. 👍

iomaganaris:
> Is it possible to easily add a UT for this?

I am in favour of adding a test for this however it's not very easy to do so, since this is used in the code generation of the `.ispc` file and we don't have set up a system for testing the code generation. 
It makes a lot of sense to add an integration test which tests this but it's not merged yet (PR #346 ).  I can keep a note in the related issue and add a test when it is merged.
For now I can just provide the different output:
Before:
```
        inst->ist[id] = kctin_global.ist0;
                double dt_saved_value, old_cst, old_ost;
        dt_saved_value = nt->_dt;
```
After:
```
        inst->ist[id] = kctin_global.ist0;

        double old_cst, old_ost;
        double uniform dt_saved_value;
        dt_saved_value = nt->_dt;
```

PR #361, opened 2020-07-02
Fix sorting of dict to have consistent order in Jinja rendering

### LINKED ISSUE #350: Cmake-format changes order of code on every PR
In a lot of PR we have figured out that the following code changes the order it is printed, going up or down the code.
```
set(PYBIND_GENERATED_SOURCES	
    ${PROJECT_BINARY_DIR}/src/pybind/pyast.cpp	
    ${PROJECT_BINARY_DIR}/src/pybind/pyast.hpp	
    ${PROJECT_BINARY_DIR}/src/pybind/pysymtab.cpp	
    ${PROJECT_BINARY_DIR}/src/pybind/pyvisitor.cpp	
    ${PROJECT_BINARY_DIR}/src/pybind/pyvisitor.hpp	
)
```
There is probably some issue with `cmake-format` and some specific which we need to figure out

tristan0x:
Fixes #350

### DIFF
--- a/src/language/templates/code_generator.cmake
+++ b/src/language/templates/code_generator.cmake
@@ -21,7 +21,7 @@ set(CODE_GENERATOR_YAML_FILES
 {% endfor %}
 )
 
-{% for dir, files in outputs.items() | sort(attribute="1") %}
+{% for dir, files in outputs | dictsort %}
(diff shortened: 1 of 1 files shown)


pramodk:
@cattabiani : CVF CI to ignore?

PR #364, opened 2020-07-14
Fix printing token in ISPC rename when there is none

iomaganaris:
- If there is no token in variable which is being renamed then there is a segmentation fault
- For example when running localize pass with ISPC rename visitor

### DIFF
--- a/src/utils/common_utils.hpp
+++ b/src/utils/common_utils.hpp
@@ -81,9 +81,17 @@ bool make_path(const std::string& path);
 /// Check if directory with given path exist
 bool is_dir_exist(const std::string& path);
 
+/// Enum to wrap bool variable to select if random string
+/// should have numbers or not
+enum UseRandomNumbers : bool
+{
+    WithNumbers = true,
+    WithoutNumbers = false
+};
+
 /// Generate random std::string of length len based on a
 /// uniform distribution
-std::string generate_random_string(int len);
+std::string generate_random_string(int len, UseRandomNumbers use_numbers);
 
 /**
  * \class SingletonRandomString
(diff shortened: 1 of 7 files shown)


src/visitors/rename_visitor.cpp
```diff
@@ -46,9 +46,9 @@ void RenameVisitor::visit_name(ast::Name& node) {
     if (std::regex_match(name, var_name_regex)) {
         std::string new_name = new_name_generator(name);
         node.get_value()->set(new_name);
-        logger->warn("RenameVisitor :: Renaming variable {} in {} to {}",
+        logger->warn("RenameVisitor :: Renaming variable {}{} to {}",
                      name,
-                     node.get_token()->position(),
+                     node.get_token() != nullptr ? " at " + node.get_token()->position() : "",
```
pramodk, line 51: pure nitpick : can you take this out of the `warn` call and use variable instead?

src/utils/common_utils.hpp
```diff
@@ -81,9 +81,13 @@ bool make_path(const std::string& path);
...
 bool is_dir_exist(const std::string& path);
 
+/// Enum to wrap bool variable to select if random string
+/// should have numbers or not
+enum UseNumbersInString : bool { WithNumbers = true, WithoutNumbers = false };
+
 /// Generate random std::string of length len based on a
 /// uniform distribution
-std::string generate_random_string(int len);
+std::string generate_random_string(int len, UseNumbersInString use_numbers);
```
pramodk, line 90: @iomaganaris : we were discussing to keep here default argument `WithoutNumbers` so that all other API usage remains same i.e. no modifications?

iomaganaris: `generate_random_string` is only used in one place, inside the `SingletonRandomString`, and needs to get this variable as a parameter to select if it will add numbers or not, so this is why I kept my change in this signature

PR #374, opened 2020-07-28
squash! more log

cattabiani:


pramodk:
@cattabiani : 

> libc++abi.dylib: terminating with uncaught exception of type pybind11::error_already_set: ModuleNotFoundError: No module named 'nmodl._nmodl'

I have typically seen this when you are in the directory that contains the `nmodl` folder which is treated as module and hence `_nmodl` was not found. Worth verifying your PWD when things get executed.

PR #385, opened 2020-08-04
All LALR1 parsers share the same location class

### LINKED ISSUE #380: nmodl::parser::location class is defined more than once
I wanted to extend the _format_ library to deal with `nmodl::parser::location` as instructed [here](https://fmt.dev/latest/api.html#formatting-user-defined-types) but realized that every bison parser in NMODL has its owned `location.hh` file:
* `${CMAKE_BINARY_DIR}/src/parser/diffeq/location.hh`
* `${CMAKE_BINARY_DIR}/src/parser/nmodl/location.hh`
* `${CMAKE_BINARY_DIR}/src/parser/unit/location.hh`

Each of them defines a `location` class, all in the same `nmodl::parser` namespace. There should be either one single `nmodl::parser::location` class used by all parsers or they should all be declared in a dedicated namespace. This is an issue that should be addressed before considering extending the _format_ to deal with `nmodl::parser::location`.

_Originally posted by @tristan0x in https://github.com/BlueBrain/nmodl/pull/372#discussion_r463422929_

tristan0x:
`%define api.location.type {}` bison directive allows one to provide a custom location class.
See https://www.gnu.org/software/bison/manual/html_node/User-Defined-Location-Type.html

In this PR, all LALR1 parsers share the same `nmodl::parser::location` class, the one generated by `src/parser/nmodl.yy`.

`src/parser/verbatim.yy` remains unchanged because it is a reentrant parser whose _location_  is hidden in the compilation unit:
```c++
struct YYLTYPE
{
  int first_line;
  int first_column;
  int last_line;
  int last_column;
};
```

Fixes #380

### DIFF
--- a/src/lexer/CMakeLists.txt
+++ b/src/lexer/CMakeLists.txt
@@ -2,8 +2,10 @@
 # Various project components and their source files
 # =============================================================================
 set(BISON_GENERATED_SOURCE_FILES
+    ${PROJECT_BINARY_DIR}/src/parser/nmodl/location.hh
     ${PROJECT_BINARY_DIR}/src/parser/nmodl/nmodl_parser.cpp
     ${PROJECT_BINARY_DIR}/src/parser/nmodl/nmodl_parser.hpp
+    ${PROJECT_BINARY_DIR}/src/parser/nmodl/position.hh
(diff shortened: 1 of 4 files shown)


tristan0x:
The CI issue will be fixed by #387 
I will rebased once it is merged

PR #387, opened 2020-08-05
Fix nmodl::utils::generate_random_string helper function / Improve SingleRandomString

tristan0x:


src/utils/common_utils.hpp
```diff
@@ -107,10 +107,9 @@ class SingletonRandomString {
...
 
     /// Function to instantiate the SingletonRandomString class
-    static std::shared_ptr<SingletonRandomString> instance(const UseNumbersInString use_numbers_) {
-        static std::shared_ptr<SingletonRandomString> s{new SingletonRandomString};
-        s->use_numbers = use_numbers_;
-        return s;
+    static SingletonRandomString& instance() {
+        static SingletonRandomString srs;
+        return srs;
     }
```
pramodk, line 113: clean!! didn't need shared_ptr.

Skimming through this change - what was causing garbage name? uninitialised static variable?

tristan0x: You can take a look at the comment of the first commit 3c0459f30f105c720d317f834e2d08067ae6517d for an explanation of the bug.

pramodk: Oops...ok!

Issue #388, opened 2020-08-12
ion variable copy optimization


cattabiani:
Currently, it is hard-coded that ion variables are optimized:

```
bool CodegenCVisitor::optimize_ion_variable_copies() const {
    return true;
}
```

In this way the variables exposed for recording are not updated. In other words the simulation is correct but there is no way to record the ion variables. In addition, the "recordable" variables are still present but not updated, misleading the user to think that there is a bug. As @pramodk suggested, we could add a cli flag that disables this optimization so that the internal variables are the same as the ones recordable by coreneuron

PR #389, opened 2020-08-12
Set CodegenCVisitor::optimize_ion_variable_copies by a CLI flag

cattabiani:
`CodegenCVisitor::optimize_ion_variable_copies` was hard-coded as true. Now it can be set by a CLI flag

- Fix issue #388

### DIFF
--- a/src/nmodl/main.cpp
+++ b/src/nmodl/main.cpp
@@ -254,7 +254,7 @@ int main(int argc, const char* argv[]) {
     codegen_opt->add_flag("--force",
         force_codegen,
         "Force code generation even if there is any incompatibility");
-    codegen_opt->add_flag("--opt-ionvar-cp",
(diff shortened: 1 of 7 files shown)


src/nmodl/main.cpp
```diff
@@ -251,6 +254,9 @@ int main(int argc, const char* argv[]) {
     codegen_opt->add_flag("--force",
         force_codegen,
         "Force code generation even if there is any incompatibility");
+    codegen_opt->add_flag("--opt-ionvar-cp",
```
pramodk, line 257:
```suggestion
    codegen_opt->add_flag("--opt-ionvar-copy",
```

`nitpick` : copy might be better?

pramodk:
Oh, forgot to mention - CVF CI should be fixed. It's bit confusing (or annoying) when things are not passing.

Issue #399, opened 2020-09-17
symtab get_variables and get_variables_with_properties default inputs


cattabiani:
Currently, there are only workarounds to get all the symbols (without any matching property) from symtab in python since these 2 functions require matching properties as inputs

PR #400, opened 2020-09-17
Add symtab get_variables and get_variables_with_properties defaults. Fix #403

### LINKED ISSUE #403: CI travis mac failing because of bison
so, nmodl CI is failing on travis mac because of these lines:

```
        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/187c0d5/Formula/bison.rb;
        export PATH=/usr/local/opt/flex/bin:/usr/local/opt/bison/bin:$PATH;
        brew unlink python@2;
        brew link --overwrite python@3.8;
        export SDKROOT="$(xcrun --show-sdk-path)";
```

the error is:

`Error: Calling Non-checksummed download of bison formula file from an arbitrary URL is disabled! Use 'brew extract' or 'brew create' and 'brew tap-new' to create a formula file in a tap on GitHub instead.`

cattabiani:
Fix issue #399. Now these 2 functions can be called without
properties as inputs to collect all the symbols

Fix issue #403

Edit: due to the misleading -1 value and the fact that get_variables_with_properties does not make much sense with a default value, the default of this function has been dropped. Get_variables still returns all the variables in case no input is present

### DIFF
--- a/src/symtab/symbol_table.hpp
+++ b/src/symtab/symbol_table.hpp
@@ -152,7 +152,7 @@ class SymbolTable {
      * \param all all/any
      */
     std::vector<std::shared_ptr<Symbol>> get_variables_with_properties(
-        syminfo::NmodlType properties = syminfo::NmodlType(-1),
+        syminfo::NmodlType properties,
         bool all = false) const;
 
     std::vector<std::shared_ptr<Symbol>> get_variables_with_status(syminfo::Status status,
--- a/src/symtab/symbol_table.hpp
+++ b/src/symtab/symbol_table.hpp
@@ -131,11 +131,28 @@ class SymbolTable {
         return parent ? parent->name() : "None";
     }
 
-    std::vector<std::shared_ptr<Symbol>> get_variables(syminfo::NmodlType with,
-                                                       syminfo::NmodlType without);
+    /**
+     * get variables
+     *
+     * \param with variables with properties. 0 matches everything
+     * \param without variables without properties. 0 matches nothing
+     *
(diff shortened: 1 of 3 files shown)


src/symtab/symbol_table.hpp
```diff
@@ -131,11 +131,28 @@ class SymbolTable {
...
+     * \param with variables with properties. 0 matches everything
+     * \param without variables without properties. 0 matches nothing
+     *
+     * The two different behaviors for 0 depend on the fact that we get
+     * get variables with ALL the with properties and without ANY of the
+     * without properties
+     */
+    std::vector<std::shared_ptr<Symbol>> get_variables(
+        syminfo::NmodlType with = syminfo::NmodlType::empty,
+        syminfo::NmodlType without = syminfo::NmodlType::empty);
```
pramodk, line 146: So with this, call to `get_variables()` returns now empty vector? (or, it is supposed to return all variables?)

cattabiani: it returns all the variables because in get_variables there is a call to get_variables_with_properties(..., all = true). Already tested with the python interface

cattabiani: it is a little confusing but for get_variables_with_properties(..., true) you get all the variables with NmodlType = 0 and nothing with NmodlType = -1. If you use get_variables_with_properties(..., false) exactly the opposite.

cattabiani: since returning no variable is trivial I switched the behavior to the most useful one

pramodk (approved):
LGTM

PR #405, opened 2020-09-23
Fixes issue with net_send_buffering declared and defined after net_init

iomaganaris:
With the recent changes in the mod files for the `delayed connection` there are issues in the source files generated by `NMODL` because `net_send_buffering()` is called by `net_init()` but defined after it.
This PR prints the `net_send_buffering()` function before `net_init()` to solve this issue.
I have an additional question, whether `net_receive_buffering()` should also be printed before `net_init()` to avoid issues in the future.
cc: @pramodk @ohm314 @cattabiani


pramodk (approved):

cattabiani (approved):

ohm314 (approved):
lgtm

Issue #367, opened 2020-07-19
Remove superfluous warning `v [Argument] in rates shadows`


pramodk:
In most of the methods v is an argument to function. Even if `v` is external variable, we should suppress this warning:

```
[NMODL] [warning] :: SYMTAB :: v [Argument] in rates shadows <assigned_definition extern_neuron_var> definition in NMODL_GLOBAL
[NMODL] [info] :: Running LOCAL to ASSIGNED visitor
[NMODL] [warning] :: SYMTAB :: v [Argument] in rates shadows <assigned_definition extern_neuron_var> definition in NMODL_GLOBAL
```

cattabiani:
may you point me to a file where this happens? I have a patch but I do not find any file that presents this to test it

pramodk:
> may you point me to a file where this happens?

Sure. For example:

```
$ ./bin/nmodl /gpfs/bbp.cscs.ch/ssd/apps/hpc/jenkins/deploy/applications/2020-02-01/linux-rhel7-x86_64/intel-19.0.4/neurodamus-neocortex-1.0-44v36k2yn3/lib/mod/StochKv3.mod
[NMODL] [info] :: Processing /gpfs/bbp.cscs.ch/ssd/apps/hpc/jenkins/deploy/applications/2020-02-01/linux-rhel7-x86_64/intel-19.0.4/neurodamus-neocortex-1.0-44v36k2yn3/lib/mod/StochKv3.mod
[NMODL] [info] :: Running symtab visitor
[NMODL] [warning] :: SYMTAB :: v [Argument] in trates shadows <parameter extern_neuron_var> definition in NMODL_GLOBAL
[NMODL] [warning] :: SYMTAB :: N [Argument] in brand shadows <range assigned_definition> definition in NMODL_GLOBAL
[NMODL] [info] :: Running LOCAL to ASSIGNED visitor
[NMODL] [warning] :: SYMTAB :: v [Argument] in trates shadows <parameter extern_neuron_var> definition in NMODL_GLOBAL
[NMODL] [warning] :: SYMTAB :: N [Argument] in brand shadows <range assigned_definition> definition in NMODL_GLOBAL
[NMODL] [info] :: Running code compatibility checker
[NMODL] [info] :: Running verbatim rename visitor
[NMODL] [info] :: Running KINETIC block visitor
[...]

PR #409, opened 2020-10-21
Suppress warning for v argument shadowing (#367)

cattabiani:


cattabiani:
if we forecast more suppressions a more structured solution can be put in place. @pramodk what do you think?

ohm314 (approved):
I think this makes sense

ohm314:
> if we forecast more suppressions a more structured solution can be put in place. @pramodk what do you think?

Even though I'm not @pramodk , I'd say maybe for now it's not a big deal to keep this treated as a special case :)

Issue #390, opened 2020-08-12
Implement support for FOR_NETCON using CoreNEURON data structures


pramodk:
See corresponding change in MOD2C here : https://github.com/BlueBrain/mod2c/pull/43

cattabiani:
right now, for_netcons is not even producing working code from nmodl. It compiles to a cpp file but the syntax is not correct. I get this error:

```
error: expected ‘;’ before ‘fnpreftprefnpostftpost’
  245 |                 (*w)fnpreftprefnpostftpost{
```

PR #410, opened 2020-10-26
Add support for for_netcon

cattabiani:
Tackles issue #390

For_netcon applies the operations of the block to the weights of
all the netcons. Since all the weights are on the same vector,
weights, this means appling the same operations to a mask of values
with the appropriate offset. The mask is the arguments of the
for_netcon block.

Note: the local variables are renamed to the correct position
in the corenrn vectors (no new variables added). Some of these
local variables may have been already substituted with special
characters that break rename_visitor (i.e. (*w) ). Thus, we need
to sanitize first the regex names that we are going to substitute.

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -2713,6 +2713,7 @@ void CodegenCVisitor::print_mechanism_register() {
         printer->add_line(pnt_recline);
     }
     if (info.for_netcon_used) {
+        // index where information about FOR_NETCON is stored in the integer array
         const auto index =
             std::find_if(info.semantics.begin(), info.semantics.end(), [](const IndexSemantics& a) {
                 return a.name == naming::FOR_NETCON_SEMANTIC;
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -3777,15 +3777,17 @@ void CodegenCVisitor::visit_for_netcon(ast::ForNetcon& node) {
                                                      : "{} + id*{}"_format(index, num_int);
     printer->add_line("const size_t offset = {};"_format(offset));
     printer->add_line(
-        "for (auto i = nt->_fornetcon_perm_indices[indexes[offset]]; "
-        "i < nt->_fornetcon_perm_indices[indexes[offset]+1]; ++i) {");
+        "const size_t for_netcon_start = nt->_fornetcon_perm_indices[indexes[offset]]; ");
+    printer->add_line(
+        "const size_t for_netcon_end = nt->_fornetcon_perm_indices[indexes[offset] + 1]; ");
+
+    printer->add_line("for (auto i = for_netcon_start; i < for_netcon_end; ++i) {");
     printer->increase_indent();
     print_statement_block(*statement_block, false, false);
     printer->decrease_indent();
 
     printer->add_line("}");
 }
-
 
 void CodegenCVisitor::print_net_receive_kernel() {
     if (!net_receive_required()) {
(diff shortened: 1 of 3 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -2710,6 +2712,14 @@ void CodegenCVisitor::print_mechanism_register() {
             method_name("net_receive"), net_recv_init_arg);
         printer->add_line(pnt_recline);
     }
+    if (info.for_netcon_used) {
+        const auto index =
```
pramodk, line 2716:
```suggestion
        // index where information about FOR_NETCON is stored in the integer array
        const auto index =
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -3736,6 +3746,49 @@ void CodegenCVisitor::print_net_send_buffering() {
...
+        std::find_if(info.semantics.begin(), info.semantics.end(), [](const IndexSemantics& a) {
+            return a.name == naming::FOR_NETCON_SEMANTIC;
+        })->index;
+    const auto num_int = int_variables_size();
+
+
+    printer->add_line("");  // the first line gets indented in a strange way.
+    if (layout == LayoutType::soa)
+        printer->add_line("//I think it is soa");
+
```
pramodk, line 3776:
```suggestion

File: src/codegen/codegen_c_visitor.cpp [hunk 3]
```diff
@@ -3736,6 +3746,49 @@ void CodegenCVisitor::print_net_send_buffering() {
...
+
+    std::string offset = (layout == LayoutType::soa) ? "{}*pnodecount + id"_format(index)
+                                                     : "{} + id*{}"_format(index, num_int);
+    printer->add_line("const size_t offset = {};"_format(offset));
```
pramodk, line 3780: Just a `nitpick` : if you write : 

```
int start = nt->_fornetcon_perm_indices[indexes[offset]];
int end = nt->_fornetcon_perm_indices[indexes[offset]+1];

for (auto i = start; i < end....)
```
it would be bit better for readability?

src/codegen/codegen_c_visitor.cpp
```diff
@@ -3736,6 +3746,49 @@ void CodegenCVisitor::print_net_send_buffering() {
...
+    RenameVisitor v;
+    auto& statement_block = node.get_statement_block();
+    for (size_t i_arg = 0; i_arg < args.size(); ++i_arg) {
+        // sanitize node_name since we want to substitute names like (*w) as they are
+        auto old_name =
+            std::regex_replace(args[i_arg]->get_node_name(), regex_special_chars, R"(\$&)");
+        auto new_name = "weights[{} + nt->_fornetcon_weight_perm[i]]"_format(i_arg);
+        v.set(old_name, new_name);
+        statement_block->accept(v);
+    }
```
pramodk, line 3764: Just curious -> if we transform this FOR_NETCON block before visiting NET_RECEIVE, would it avoid this regex search/replace stuff?

what I meant is : in the visit_net_receieve, before going to children's, fix FOR_NETCON first if exist.

pramodk (commented):
overall looks good, few minor comments.

src/codegen/codegen_c_visitor.cpp
cattabiani: I was worried that "start" and "end" could be already used before in the code...

src/codegen/codegen_c_visitor.cpp
cattabiani: mmm I am not sure how would you do this. In print_net_init we rename the variables. Then there are other functions that print out changed variables like print_watch_activate(). Then we visit for_netcon and print it. If we print for_netcon first we did not call print_watch_activate or we did print with the wrong names (if we replace after fornetcon)

src/codegen/codegen_c_visitor.cpp
cattabiani: done

cattabiani:
please retest

pramodk (approved):
lgtm

Issue #52, opened 2019-03-11
parser error on numbers with leading "+"


lkeegan:
`./bin/nmodl_parser <(echo "PARAMETER { x = +5 }")`
gives the error
```
[NMODL] [info] :: Processing file : /proc/self/fd/12
terminate called after throwing an instance of 'std::runtime_error'
  what():  NMODL Parser Error : syntax error, unexpected +, expecting REAL or INTEGER or DEFINEDVAR or - [Location : 1.17]
[1]    22042 abort      ./bin/nmodl_parser <(echo "PARAMETER { x = +5 }")
```
Is this the expected behaviour?

A minus sign, or no sign both work fine:
`./bin/nmodl_parser <(echo "PARAMETER { x = -5 }")`
`./bin/nmodl_parser <(echo "PARAMETER { x = 5 }")`

PR #419, opened 2020-11-06
fix for #52 : solve parser error on numbers with leading +

WeinaJi:


WeinaJi:
please retest

ohm314:
please retest

WeinaJi:
please retest

ohm314 (approved):
lgtm

src/parser/nmodl.yy
```diff
@@ -707,6 +707,10 @@ number          :   NUMBER
                         $2->negate();
                         $$ = $2;
                     }
+                |   "+" NUMBER
```
pramodk, line 710: @nrnhines : out of curiosity : for any reason this rule wasn't there in original implementation?

pramodk (changes requested):
LGTM.
@WeinaJi : we should one test that checks this rule. See below file where we add nmodl constructs for testing : https://github.com/BlueBrain/nmodl/blob/6d39a657b8908933875f2f09b7dede3a9748757b/test/unit/utils/nmodl_constructs.cpp#L189

WeinaJi:
> LGTM.
> @WeinaJi : we should one test that checks this rule. See below file where we add nmodl constructs for testing :
> 
> https://github.com/BlueBrain/nmodl/blob/6d39a657b8908933875f2f09b7dede3a9748757b/test/unit/utils/nmodl_constructs.cpp#L189

@pramodk hmm... there is one issue with test case ```testvisitor/Scenario: Convert AST back to NMODL form```
The "+" sign is removed after passing so when converting back to NMODL, the generated text doesn't match the origin format. I think that's expected, but how can I solve the matching error ?

pramodk:
Sorry, missed your comment.

In that case you can pass an additional argument with the expected nmodl text. See e.g. https://github.com/BlueBrain/nmodl/blob/6d39a657b8908933875f2f09b7dede3a9748757b/test/unit/utils/nmodl_constructs.cpp#L396

pramodk (approved):

pramodk:
thanks @WeinaJi !

Issue #417, opened 2020-11-06
Automated wheel building


pramodk:
Similar to NEURON, we should build the wheel for NMODL in automated way. Here [are instructions from NEURON](https://github.com/neuronsimulator/nrn/tree/master/packaging/python).

Here are steps:

* Adapt Dockerfile for NMODL (if we are using C++14 and if GCC 4.8.5 is not sufficient, we might need to use Manylinux2010)
* This is how docker image is launched to build for linux : [Pipeline](https://github.com/neuronsimulator/nrn/blob/cccff1461ebf15dd6686ff966ae9dd9f7cc0775a/azure-pipelines.yml#L39)
* Update wheel [building script](https://github.com/neuronsimulator/nrn/blob/master/packaging/python/build_wheels.bash)
* Create [test script](https://github.com/neuronsimulator/nrn/blob/master/packaging/python/test_wheels.sh) for NMODL

cc: @ohm314

ohm314:
We use C++14 for now, so on that front we're safe

PR #420, opened 2020-11-06
Automated python wheel building

ohm314:
Following the work done in NEURON and as described in #417, this adds the necessary scripts for building linux and osx python wheels for nmodl.


cattabiani:
can I try to build a wheel in docker using the provided readme?

packaging/python/README.md
```diff
@@ -0,0 +1,64 @@
+
+### Building Python Wheels
+
+Note: This is only slightly adapted from NEURONs [scripts](https://github.com/neuronsimulator/nrn/tree/master/packaging/python).
+
+NMODL wheels are built in a manylinux2010 image. Since the generic docker image is very basic (CentOS 6) a new image, which brings updated cmake3 (3.12), flex and bison was prepared and made available at https://hub.docker.com/r/bluebrain/nmodl (tag: wheel).
+
+#### Setting up Docker
+
+[Docker](https://en.wikipedia.org/wiki/Docker_(software)) is required for building Linux wheels. You can instructions to
```
cattabiani, line 10:
```suggestion
[Docker](https://en.wikipedia.org/wiki/Docker_(software)) is required for building Linux wheels. You can find instructions to
```

cattabiani:
following the readme, I reach this error when I launch `build_wheels.bash`

```
CMake Error at /nmodlwheel/cmake/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find PythonInterp: Found unsuitable version "3.5.10", but
  required is at least "3.6" (found
```

azure-pipelines.yml
```diff
@@ -1,6 +1,24 @@
...
 #=============================================================================
+variables:
+- name: ReleaseWheelBuild
+  value: false
+- name: NightlyWheelBuild
+  value: false
+- name: buildWheel
+  value: $[or(in(variables['Build.Reason'], 'Schedule'),
+              and(in(variables['Build.Reason'], 'Manual'), eq(variables.NightlyWheelBuild, true)))]
+
```
pramodk, line 12: I am wondering when/how `NightlyWheelBuild` variable becomes True.

azure-pipelines.yml
```diff
@@ -82,23 +100,23 @@ jobs:
...
+#      sudo apt-get install -y python3.7-venv
+#      python3.7 -m pip install -U scikit-build
+#      python3.7 setup.py sdist bdist_wheel -j 6
+#      rm -rf venv _skbuild
+#      python3.7 -m venv venv
+#      source ./venv/bin/activate
+#      pip install --upgrade pip
+#      pip install dist/*.whl
+#      nmodl $(Build.Repository.LocalPath)/test/integration/mod/cabpump.mod sympy --analytic
+#      rm cabpump.cpp
```
pramodk, line 116: Just a note : when we will test wheel, we should make sure to run `cabpump.mod` and other mod files. These are added to see if generated binary/wheel has everything needed including sympy.

azure-pipelines.yml
```diff
@@ -123,26 +141,108 @@ jobs:
...
+        python.version: '3.9'
+  steps:
+  - checkout: self
+    submodules: true
+    condition: and(succeeded(), or(eq(variables.buildWheel, true), eq(variables['python.version'], '3.9')))
+  - script: |
+      docker run --rm \
+        -w /root/nmodl \
+        -v $PWD:/root/nmodl \
+        -e NMODL_NIGHTLY_TAG="-nightly" \
```
pramodk, line 189: this shouldn't be hard-coded here.

packaging/test_wheel.bash
```diff
@@ -0,0 +1,72 @@
...
+  venv_name="nmodl_test_venv_${python_ver}"
+  $python_exe -m venv $venv_name
+  . $venv_name/bin/activate
+  python_exe=`which python`
+else
+  echo " == Using global install == "
+fi
+
+# on osx we need to install pip from source
+if [[ "$OSTYPE" == "darwin"* ]] && [[ "$python_ver" == "35" ]]; then
```
pramodk, line 54: if you are not enabling python 35 then this can go away.

src/config/config.h
```diff
@@ -64,7 +65,14 @@ struct NrnUnitsLib {
...
             }
         }
-        throw std::runtime_error("Could not found nrnunits.lib");
+        std::ostringstream err_msg;
+        err_msg << "Could not find nrnunits.lib in any of:\n";
+        for (const auto& path: NRNUNITSLIB_PATH) {
+            err_msg << path << "\n";
+        }
+
+
```
pramodk, line 74: nitpick : extra whitespace

pramodk (commented):
Here are comments from quick look at this PR.

azure-pipelines.yml
ohm314: This is a user variable that you set whenever you do a manual build that should produce a wheel. The aim is that not every manual build shoudl produce. Only when you set the variable to true when you start the build you get a wheel

azure-pipelines.yml
ohm314: Good point, i missed to add that into the `test_wheel.bash` script

pywheel/shim/_binwrapper.py
```diff
@@ -37,6 +41,9 @@ def _config_exe(exe_name):
     # add nmodl home to environment (i.e. necessary for nrnunits.lib)
     os.environ["NMODLHOME"] = NMODL_HOME
 
+    # set PYTHONPATH for embedded python to properly find the nmodl module
+    os.environ["PYTHONPATH"] = working_set.by_key[package_name].location + ':' + os.environ.get("PYTHONPATH", "")
```
cattabiani, line 45: yes but... why do we need this now?

packaging/build_requirements.txt
```diff
@@ -0,0 +1,4 @@
+Jinja2>=2.9.3
+PyYAML>=3.13
+pytest
+sympy>=1.3,<1.6
```
cattabiani, line 4: this is going to collide with my pr

packaging/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+For OSX on a system with the all dependencies you have to clone the NMODL repository and do:
+
+```
+cd nrn
+bash packaging/python/build_wheels.bash osx
+```
+
+
```
cattabiani, line 47:
```suggestion
```

cattabiani: knitpick: too many returns?

cattabiani (commented):
ok, just a few questions and nitpicks. 

I checked a little the azure-pipeline + wheels.yml but I am no expert

pywheel/shim/_binwrapper.py
ohm314: I think we may have just not tested this properly in the usecase of wheel+python venv. Why it works on linux and not on mac without the PYTHONPATH set is unclear to me too. But I believe the patch does fix the issue and it makes sense given that embedded python instances are known to have trouble finding site-packages modules (https://bugs.python.org/issue22213)

packaging/build_requirements.txt
ohm314: can you share your versions? then depending on who merges first will have to fix the file.

azure-pipelines.yml
pramodk: IIUC, wheel won't be uploaded here unless NightlyWheelBuild is set explicitly
 In case of nightly wheel we want to upload it for every scheduled build. So we can do opposite condition here i.e. don't upload if user asked not to do. (see neuron)

pywheel/shim/_binwrapper.py
pramodk: As discussed on slack, embedded python has some weird (platform specific) behaviour.

packaging/build_wheels.bash
```diff
@@ -0,0 +1,129 @@
...
+
+    deactivate
+}
+
+
+build_wheel_osx() {
+    echo "[BUILD WHEEL] Building with interpreter $1"
+    local skip=
+    setup_venv "$1"
+    (( $skip )) && return 0
```
pramodk, line 74: for clarity : how/when is this skipped?

pramodk (commented):
One minor thing otherwise we are good to merge this!

azure-pipelines.yml
pramodk: I made a mistake here ! (I didn't read `$[or` at the beginning). Makes perfect sense!

packaging/build_requirements.txt
pramodk: I think Alessandro can rebase, shouldn't be big conflict

pramodk (approved):
LGTM!

(Katta mentioned some nitpicks/extra lines etc)

packaging/build_wheels.bash
pramodk: For bash nube like me:
```
local foo="" and local foo= are exactly equivalent. In both cases,
the right-hand side of the equal sign is an empty string.

local foo and local foo= are different: local foo leaves foo unset
while local foo= sets foo to an empty string. 
```

I see that `setup_venv` is doing check and setting the value.

ohm314: if version is less than 36, that comes out of `setup_venv`

pramodk:
Just a note here : build times on azure are bit high now : from ~14 mins to ~30 mins

image

PR #421, opened 2020-11-09
Print variable with different precision depending of LEGACY_UNITS

alkino:
Use more format instead of stringstream in codegen.

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -986,17 +986,17 @@ std::vector<SymbolType> CodegenCVisitor::get_shadow_variables() {
 /****************************************************************************************/
 
 std::string CodegenCVisitor::get_parameter_str(const ParamVector& params) {
-    std::stringstream param_ss;
+    std::string param{};
     for (auto iter = params.begin(); iter != params.end(); iter++) {
-        param_ss << "{}{} {}{}"_format(std::get<0>(*iter),
-                                       std::get<1>(*iter),
-                                       std::get<2>(*iter),
-                                       std::get<3>(*iter));
+        param += "{}{} {}{}"_format(std::get<0>(*iter),
+                                    std::get<1>(*iter),
+                                    std::get<2>(*iter),
+                                    std::get<3>(*iter));
         if (!utils::is_last(iter, params)) {
-            param_ss << ", ";
+            param += ", ";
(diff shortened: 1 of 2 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -986,17 +986,17 @@ std::vector<SymbolType> CodegenCVisitor::get_shadow_variables() {
 /****************************************************************************************/
 
 std::string CodegenCVisitor::get_parameter_str(const ParamVector& params) {
-    std::stringstream param_ss;
+    std::string param{};
```
iomaganaris, line 989: I am curious. Why does this function needs to change to using string as well?

alkino: it does not have to. But this is the last one using stringstream in this file, all other methods use fmt.

iomaganaris: I see. Thanks for the answer!

iomaganaris (approved):
LGTM

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1994,10 +1994,17 @@ void CodegenCVisitor::print_nmodl_constants() {
...
-            ss << it->get_value()->get_value() << ";";
-            printer->add_line(ss.str());
+#ifndef USE_LEGACY_UNITS
+#define USE_LEGACY_UNITS 0
+#endif
+#if USE_LEGACY_UNITS
+            std::string format_string = "static const double {} = {:g};";
+#else
+            std::string format_string = "static const double {} = {:.18g};";
+#endif
```
pramodk, line 2004: Just a `nitpick` : you can directly use `#if defined(USE_LEGACY_UNITS)`  here?

alkino: No because it can be defined to `0`

pramodk: Ok. I see now in CMakeLists.txt it's either 0 to 1 i.e. always defined. Why we need:
```
#ifndef USE_LEGACY_UNITS
 #define USE_LEGACY_UNITS 0
 #endif
```
?

pramodk (approved):

pramodk:
I also make travis optional because its taking too long!

PR #423, opened 2020-11-09
fixup! ispc visitor rounding small doubles

### LINKED ISSUE #422: ispc visitor rounding doubles
Currently the ispc visitor rounds doubles with too little precision (fmt default). We need more precision because, in some mod files (i.e. TaDa_t or TXXDynamicSwitch) there are thresholds at 1e-12

cattabiani:
fix #422 

We use 16 decimal points (as it is enough for converting doubles)

### DIFF
--- a/src/codegen/codegen_ispc_visitor.cpp
+++ b/src/codegen/codegen_ispc_visitor.cpp
@@ -136,7 +136,7 @@ std::string CodegenIspcVisitor::float_to_string(float value) {
         return "{:.1f}"_format(value);
     }
     if ((value <= 1.0f) && (value >= -1.0f)) {
-        return "{:f}"_format(value);
+        return "{:.6f}"_format(value);
     } else {
         auto e = std::log10(std::abs(value));
         if (e < 0.0f) {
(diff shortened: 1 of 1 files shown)


src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -115,7 +115,7 @@ std::string CodegenIspcVisitor::double_to_string(double value) {
         return "{:.1f}d"_format(value);
     }
     if ((value <= 1.0) && (value >= -1.0)) {
-        return "{:f}d"_format(value);
+        return "{:.16f}d"_format(value);
```
iomaganaris, line 118: I am thinking whether we should use the same precision for the rest of the float/double strings where `:f` is used to make sure we have the best precision and avoid any other possible issues. 
Especially in `float_to_string` in the same case. We should just use `:.6f` there.

pramodk: I don't remember/know all implications. I will let @ohm314 to provide suggestion.
(the goal would be to retain representation close to what is written in MOD file)

pramodk: Ok, for now lets change as Ioannis suggested and then we can merge this.

pramodk (approved):

cattabiani:
please retest

src/codegen/codegen_ispc_visitor.cpp
ohm314: So, it should be really only here and `float_to_string` where this formatting is needed.. that's good. The problem I see with always printing all digits is that sometimes you will then get at the the last digits slightly wrong representation.. but it's true that overall this approach should lead to less surprises. So I agree with @iomaganaris

PR #425, opened 2020-11-11
In CMAKE use VERSION of project command

alkino:


ohm314 (approved):
lgtm - why is this better? just shorter?

alkino:
It's shorter and more straightforward. Variables are declared automatically.

alkino:
Please retest

pramodk:
@alkino 's `make` should be aliased to `make && make test && make cmake-format  && make clane-format` :smiley:

alkino:
> make && make test && make cmake-format && make clang-format

`make -j8 all test cmake-format clang-format`

pramodk:
even better! thanks!

CMakeLists.txt
```diff
@@ -6,13 +6,14 @@
 # =============================================================================
 
 cmake_minimum_required(VERSION 3.3.0 FATAL_ERROR)
-project(NMODL CXX)
+project(
+  NMODL
+  VERSION 0.2
+  LANGUAGES CXX)
```
pramodk, line 12: just a nitpick : does cmake-format force splitting this on multiple line? quite superfluous !

alkino: Yes, I agree ;)

pramodk: @alkino : note there is `# ~~~` marker in cmake-format to disable it (and avoid absurd formatting)

alkino: This is the default rule of no more than 3 words in a command on the same line. Maybe we can change it to 10?

pramodk: do you mean number of args settings? I don't know if there is an option but I would say why not!

PR #426, opened 2020-11-12
fallback to c generator when ispc keyword in mod file vars

### LINKED ISSUE #424: "in" variable in mod files break ispc compilation (reserved keyword)
In ican.mod (thalamus) there is this:

> USEION n READ en WRITE **in** VALENCE 1

Unfortunately, **in** is a reserved keyword for ispc:

> The following identifiers are reserved as language keywords: bool, break, case, cdo, cfor, char, cif, cwhile, const, continue, default, do, double, else, enum, export, extern, false, float, for, foreach, foreach_active, foreach_tiled, foreach_unique, goto, if, **in**, inline, noinline, __vectorcall, int, int8, int16, int32, int64, launch, NULL, print, return, signed, sizeof, soa, static, struct, switch, sync, task, true, typedef, uniform, union, unsigned, varying, void, volatile, while.

The problem with just renaming in the ispc code is that in the hoc/python file we should be able to do something like:

`a = soma(0.5)._ref_in`

maybe I am missing something but I do not think that this would work if we rename the variable for ispc...

cattabiani:
Fix #424


cattabiani:
please retest

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -25,6 +25,19 @@ using symtab::syminfo::Status;
...
+    ast::AstNodeType::EIGEN_LINEAR_SOLVER_BLOCK,
+    ast::AstNodeType::WATCH_STATEMENT,
+    ast::AstNodeType::TABLE_STATEMENT};
+
+const std::unordered_set<std::string> CodegenIspcVisitor::incompatible_var_names = {
+    "cdo",           "cfor",           "cif",    "cwhile",   "foreach",      "foreach_active",
+    "foreach_tiled", "foreach_unique", "in",     "noinline", "__vectorcall", "int8",
+    "int16",         "int32",          "int64",  "launch",   "print",        "soa",
+    "sync",          "task",           "varying"};
+
```
pramodk, line 40: This seems strange formatting to me. we should disable clang-format for this block ?

test/integration/mod/ican.mod
```diff
@@ -0,0 +1,56 @@
+: simple first-order model of calcium dynamics
+
+        NEURON {
```
pramodk, line 3: I would do minimal mod file. (remove all other part of mod file which are not required to test  this ISPC keyword issue)

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -292,6 +305,31 @@ std::string CodegenIspcVisitor::ptr_type_qualifier() {
     }
 }
 
+std::string CodegenIspcVisitor::global_var_structure_type_decorator() {
+    if (wrapper_codegen) {
+        return CodegenCVisitor::global_var_structure_type_decorator();
+    } else {
+        return "uniform ";  // @note: extra space needed to separate qualifier from var name.
+    }
+}
```
pramodk, line 314: I would like to discuss this bit in detail - I am thinking isn't it would be simple if somehow we avoid execution of ISPC codegen completely and use C++ visit? Lets discuss this offline.

src/codegen/codegen_ispc_visitor.cpp
cattabiani: I find it quite nice instead!

pramodk: `beauty lies in the eye of the beholder` - I am fine with it then :)

src/codegen/codegen_ispc_visitor.cpp
cattabiani: probably it is possible. However, the ispc file would be completly empty. Does it work in that case?

pramodk: > Does it work in that case?

I hope! need to test...

cattabiani: if I do not write anything in the ispc file I get:
`Error: Premature end of file: syntax error, unexpected end of file.`

cattabiani: ok, solved

cattabiani:
please retest

nmodl/__init__.py
```diff
@@ -1,7 +1,5 @@
-import imp
 try:
-    # check if nmodl binds are built before importing
-    imp.find_module('_nmodl')
+# check if nmodl binds are built before importing
```
ohm314, line 2:
```suggestion
    # Try importing but catch exception in case bindings are not available
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1260,6 +1260,14 @@ std::string CodegenCVisitor::ptr_type_qualifier() {
     return "__restrict__ ";
 }
 
+/// Useful in ispc so that variables in the global struct get "uniform "
+std::string CodegenCVisitor::global_var_struct_type_decorator() {
```
ohm314, line 1264:
```suggestion
std::string CodegenCVisitor::global_var_struct_type_qualifier() {
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -2370,7 +2378,9 @@ void CodegenCVisitor::print_coreneuron_includes() {
  * Note that static variables are already initialized to 0. We do the
  * same for some variables to keep same code as neuron.
  */
-void CodegenCVisitor::print_mechanism_global_var_structure(bool wrapper) {
+void CodegenCVisitor::print_mechanism_global_var_structure() {
+    const auto decorator = global_var_struct_type_decorator();
```
ohm314, line 2382:
```suggestion
    const auto qualifier = global_var_struct_type_qualifier();
```
and then adjust everywhere below

src/codegen/codegen_c_visitor.hpp
```diff
@@ -882,6 +887,24 @@ class CodegenCVisitor: public visitor::AstVisitor {
...
+    /**
+     * The used global type decorator
+     *
+     * For C code generation this is empty
+     * \return ""
+     *
+     * For ispc
+     * \return "uniform "
+     */
+    virtual std::string global_var_struct_type_decorator();
```
ohm314, line 899:
```suggestion
    virtual std::string global_var_struct_type_qualifier();
```

src/codegen/codegen_c_visitor.hpp
```diff
@@ -882,6 +887,24 @@ class CodegenCVisitor: public visitor::AstVisitor {
      */
     virtual std::string ptr_type_qualifier();
 
+    /**
+     * The used global type decorator
```
ohm314, line 891:
```suggestion
     * The used global type qualifier
```

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -292,6 +305,23 @@ std::string CodegenIspcVisitor::ptr_type_qualifier() {
     }
 }
 
+std::string CodegenIspcVisitor::global_var_struct_type_decorator() {
+    if (wrapper_codegen) {
+        return CodegenCVisitor::global_var_struct_type_decorator();
+    } else {
+        return "uniform ";  // @note: extra space needed to separate qualifier from var name.
```
ohm314, line 312: put the space not here but actually where you use the qualifier

ohm314 (commented):
Very nice, @cattabiani ! Just a few minor nitpicking 😄

src/codegen/codegen_ispc_visitor.cpp
cattabiani: I am following the standard set by `ptr_type_qualifier()`

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -683,43 +551,84 @@ void CodegenIspcVisitor::print_backend_compute_routine_decl() {
...
+    emit_fallback = std::vector<bool>(BlockType::BlockTypeEnd, false);
+
     if (info.initial_node) {
         emit_fallback[BlockType::Initial] =
-            has_incompatible_nodes(*info.initial_node) ||
+            emit_fallback[BlockType::Initial] || has_incompatible_nodes(*info.initial_node) ||
             visitor::calls_function(*info.initial_node, "net_send") || info.require_wrote_conc;
     } else {
-        emit_fallback[BlockType::Initial] = info.net_receive_initial_node ||
+        emit_fallback[BlockType::Initial] = emit_fallback[BlockType::Initial] ||
```
pramodk, line 609: > `emit_fallback[BlockType::Initial] = emit_fallback[BlockType::Initial] ||` 

I might be missing something here: why do we `||` with `emit_fallback[x]` itself ?

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -683,43 +551,84 @@ void CodegenIspcVisitor::print_backend_compute_routine_decl() {
     }
 }
 
-void CodegenIspcVisitor::determine_target() {
+bool CodegenIspcVisitor::check_incompatibilities() {
     const auto& has_incompatible_nodes = [this](const ast::Ast& node) {
         return !collect_nodes(node, incompatible_node_types).empty();
     };
 
+    const auto get_name_from_symbolTypeVector = [](const SymbolType& var) -> const std::string& {
```
pramodk, line 559: nitpick : we shouldn't mix `_` and `CamelCase` for var name.

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -683,43 +551,84 @@ void CodegenIspcVisitor::print_backend_compute_routine_decl() {
...
+        // info.top_local_variables is not checked because it should be addressed by the
+        // renameIspcVisitor
+        check_incompatible_var_name<SymbolType>(info.global_variables,
+                                                get_name_from_symbolTypeVector) ||
+
+
+        check_incompatible_var_name<SymbolType>(info.constant_variables,
+                                                get_name_from_symbolTypeVector)) {
+        return true;
+    }
```
pramodk, line 591: nitpick : I don't have specific suggestion here but I would say that this code with too long `if conditions` is bit unreadable.

src/codegen/codegen_ispc_visitor.cpp
cattabiani: so that it does not override a 0 if we had a 1 before. In this way it becomes more modular and scalable

src/codegen/codegen_ispc_visitor.cpp
cattabiani: ok, I will change it. However, it becomes more wordy

pramodk (approved):
LGTM

PR #428, opened 2020-11-13
Add NMODL_ENABLE_PYTHON_BINDINGS option to disable python bindings

### LINKED ISSUE #391: Can we build partial nmodl python module?
When NMODL is build as sub-project under NEURON or CoreNEURON, build time is quite high.

image

CoreNEURON doesn't use all python bindings of NMODL but only need access to solvers. I wonder if there is way to build minimal nmodl module so that build time can be reduced.

Thoughts?

pramodk:
- NMODL_ENABLE_PYTHON_BINDINGS=ON by default
  - if python bindings are not enabled, show warning at the import
  - do not add dependency with _nmodl target
  - use imp module to test existence of package

fixes #391 (not really but I think this is sufficient for now)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 
+# =============================================================================
+# Build options for NMODL
+# =============================================================================
+option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybin11 based python bindings" ON)
 option(NMODL_ENABLE_LEGACY_UNITS "Use original faraday, R, etc. instead of 2019 nist constants" OFF)
 if(NMODL_ENABLE_LEGACY_UNITS)
   add_definitions(-DUSE_LEGACY_UNITS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 # =============================================================================
 # Build options for NMODL
 # =============================================================================
-option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybin11 based python bindings" ON)
+option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybind11 based python bindings" ON)
 option(NMODL_ENABLE_LEGACY_UNITS "Use original faraday, R, etc. instead of 2019 nist constants" OFF)
 if(NMODL_ENABLE_LEGACY_UNITS)
   add_definitions(-DUSE_LEGACY_UNITS)
--- a/nmodl/__init__.py
+++ b/nmodl/__init__.py
@@ -1,4 +1,10 @@
-from ._nmodl import NmodlDriver, to_json, to_nmodl  # noqa
-from ._nmodl import __version__
+import imp
+try:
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 
+# =============================================================================
+# Build options for NMODL
+# =============================================================================
+option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybin11 based python bindings" ON)
```
alkino, line 25: `pybind11`

src/pybind/CMakeLists.txt
```diff
@@ -43,39 +43,49 @@ else()
...
+    $<TARGET_OBJECTS:lexer_obj>
+    $<TARGET_OBJECTS:util_obj>
+    $<TARGET_OBJECTS:printer_obj>)
 
-add_dependencies(_nmodl pyastgen)
-add_dependencies(_nmodl lexer_obj)
-add_dependencies(_nmodl util_obj)
+  add_dependencies(_nmodl pyastgen)
+  add_dependencies(_nmodl lexer_obj)
+  add_dependencies(_nmodl util_obj)
```
alkino, line 67: One line?

alkino (approved):

CMakeLists.txt
pramodk: thanks! done!

src/pybind/CMakeLists.txt
pramodk: thanks! done!

src/pybind/CMakeLists.txt
```diff
@@ -43,39 +43,46 @@ else()
...
-  ${NMODL_PROJECT_SOURCE_DIR}/src/ast/ast_common.hpp
-  ${NMODL_PROJECT_SOURCE_DIR}/src/pybind/pybind_utils.hpp
-  ${NMODL_PROJECT_SOURCE_DIR}/src/pybind/pynmodl.cpp
-  ${PYBIND_GENERATED_SOURCES}
-  $<TARGET_OBJECTS:symtab_obj>
-  $<TARGET_OBJECTS:visitor_obj>
-  $<TARGET_OBJECTS:lexer_obj>
-  $<TARGET_OBJECTS:util_obj>
-  $<TARGET_OBJECTS:printer_obj>)
+# avoid _nmodl target if python bindings are disabled
```
tristan0x, line 46: Can't you avoid including `src/pybind/CMakeLists.txt` if `NMODL_ENABLE_PYTHON_BINDINGS` is disabled instead?

pramodk: @tristan0x : not really because we still use pybind11 `pybind11::initialize_interprete` interface to execute python code for Sympy solver.

nmodl/__init__.py
```diff
@@ -1,4 +1,10 @@
...
-from ._nmodl import __version__
+import imp
+try:
+    # check if nmodl binds are built before importing
+    imp.find_module('_nmodl')
+    from ._nmodl import NmodlDriver, to_json, to_nmodl  # noqa
+    from ._nmodl import __version__
+    __all__ = ["NmodlDriver", "to_json", "to_nmodl"]
+except ImportError:
+    print("[NMODL] [warning] :: Python bindings are not available")
```
cattabiani, line 9: do we want to specify which import is failing?

cattabiani (approved):
just one comment that is not blocking. As you prefer

nmodl/__init__.py
pramodk: there is only one module called `_nmodl`. So IMO it's sufficient to say bindings are not enabled.

PR #429, opened 2020-11-18
Put nrnunits.lib into share/nmodl instead of share

alkino:


cattabiani:
just a question: why?

alkino:
> just a question: why?

Currently 3 differents people try to install differents `nrnunits.lib` files: neuronsimulator/nrn, BlueBrain/nmodl and BlueBrain/mod2c

Because they are different we want to have three in the same time.
`share/nrnunits.lib` is for neuronsimulator/nrn
`share/mod2c/nrnunits.lib` for BlueBrain/mod2c
`share/nmodl/nrnunits.lib` for BlueBrain/nmodl

alkino:
I checked that it works well with Coreneuron, but in fact nmodl don't use the env var for finding nrnunits.lib, it's hardcoded inside the binary

pramodk (approved):
LGTM

PR #430, opened 2020-11-18
Don't install pycache files

alkino:
Remove `__pycache__` and install optional binaries only if compiled in Debug


pramodk (approved):
LGTM. You might need to fix cmake format.

Issue #433, opened 2020-11-19
ispc nmodl constants


cattabiani:
Right now we have in `codegen_ispc_visitor`:

```
void CodegenIspcVisitor::print_nmodl_constants() {
    printer->add_newline(2);
    printer->add_line("/** constants used in nmodl. */");
    // we use here macros to work around ispc's PI being declared in global namespace
    printer->add_line("static const uniform double FARADAY = 96485.3d;");
    printer->add_line("static const uniform double ISPC_PI = 3.14159d;");
    printer->add_line("static const uniform double R = 8.3145d;");
}
```

This does not take care of a few things:
- in case of non-legacy units faraday is a slightly different number
- in case someone wrote the mod file using, for example, `FARADAY = (faraday)  (kilocoulombs)` the value should be 96.4853, not 96485.3

PR #434, opened 2020-11-19
fixup! ispc nmodl constants

cattabiani:
Fix  issue #433

- Constants were hard-coded so no custom constants, no
 unit changes, no double_to_string handling

I leave to double_to_string to handle how to chop doubles

### DIFF
--- a/src/codegen/codegen_ispc_visitor.cpp
+++ b/src/codegen/codegen_ispc_visitor.cpp
@@ -477,12 +477,17 @@ void CodegenIspcVisitor::print_headers_include() {
 }
 
 void CodegenIspcVisitor::print_nmodl_constants() {
-    printer->add_newline(2);
-    printer->add_line("/** constants used in nmodl. */");
-    // we use here macros to work around ispc's PI being declared in global namespace
-    printer->add_line("static const uniform double FARADAY = 96485.3d;");
-    printer->add_line("static const uniform double ISPC_PI = 3.14159d;");
-    printer->add_line("static const uniform double R = 8.3145d;");
+    if (!info.factor_definitions.empty()) {
+        printer->add_newline(2);
+        printer->add_line("/** constants used in nmodl */");
+        for (auto& it: info.factor_definitions) {
+            const std::string name = it->get_node_name() == "PI" ? "ISPC_PI" : it->get_node_name();
+            std::string format_string = "static const uniform double {} = {};";
+            printer->add_line(fmt::format(format_string.c_str(),
+                                          name,
+                                          double_to_string(it->get_value()->get_value())));
+        }
+    }
 }
 
 void CodegenIspcVisitor::print_wrapper_headers_include() {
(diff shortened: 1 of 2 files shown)


ohm314 (approved):
lgtm

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -477,12 +477,17 @@ void CodegenIspcVisitor::print_headers_include() {
...
-    // we use here macros to work around ispc's PI being declared in global namespace
-    printer->add_line("static const uniform double FARADAY = 96485.3d;");
-    printer->add_line("static const uniform double ISPC_PI = 3.14159d;");
-    printer->add_line("static const uniform double R = 8.3145d;");
+    if (!info.factor_definitions.empty()) {
+        printer->add_newline(2);
+        printer->add_line("/** constants used in nmodl */");
+        for (auto& it: info.factor_definitions) {
+            const std::string name = it->get_node_name() == "PI" ? "ISPC_PI" : it->get_node_name();
+            std::string format_string = "static const uniform double {} = {};";
```
iomaganaris, line 485:
```suggestion
            std::string format_string = "static const uniform double {} = {}d;";
```
Isn't `d` also needed for `double`s on `ispc`?

iomaganaris: This is already handled by `double_to_string`, so all good!

iomaganaris (approved):

pramodk (changes requested):
Unrelated but CMake v3.19 is out and we have something to fix in out CMake:

```
CMake Error at /usr/local/Cellar/cmake/3.19.0/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:281 (message):
1124  find_package_check_version(): Cannot be used outside a 'Find Module'
1125Call Stack (most recent call first):
1126  /usr/local/Cellar/cmake/3.19.0/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:530 (find_package_check_version)
1127  cmake/FindPythonModule.cmake:62 (find_package_handle_standard_args)
1128  CMakeLists.txt:114 (find_python_module)
1129
1130
1131-- Configuring incomplete, errors occurred!
1132See also "/Users/travis/build/BlueBrain/nmodl/build/CMakeFiles/CMakeOutput.log".
1133The command "cmake .. -DPYTHON_EXECUTABLE=$(which python3) -DCMAKE_INSTALL_PREFIX=$HOME/nmodl" exited with 1.
```

cattabiani:
Ok, I will wait for another PR to solve the issue and rebase this

pramodk (approved):

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -477,12 +477,17 @@ void CodegenIspcVisitor::print_headers_include() {
...
+            printer->add_line(fmt::format(format_string.c_str(),
+                                          name,
+                                          double_to_string(it->get_value()->get_value())));
```
iomaganaris, line 488: `double_to_string` is printing all the doubles with 6 decimals, since it's using the `%f` to print them, while `mod2c` is using 6 or 18 digits depending on if we have legacy or non legacy units.
I think we should edit `double_to_string` or use a different function to have the same behaviour in this case.
Maybe this is not needed though after @pramodk 's change were all the doubles will be printed as strings.

iomaganaris (changes requested):

src/codegen/codegen_ispc_visitor.cpp
pramodk: Discussed this offline - we will fix this in next PR.

PR #435, opened 2020-11-20
Build our own cmake 3.3.2 in CI

ohm314:
especially with homebrew one has to go with whatever version they
choose, which is not testing our minimum version and can cause unrelated
failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,6 @@ addons:
     packages:
       - flex
       - bison
-      - cmake
       - dvipng
       - doxygen
       - libboost-all-dev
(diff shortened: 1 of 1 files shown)


ohm314:
Actually I'm going to switch to cmake binary packages

.travis.yml
```diff
@@ -80,9 +78,14 @@ before_install:
         brew unlink python@2;
         brew link --overwrite python@3.8;
         export SDKROOT="$(xcrun --show-sdk-path)";
+        curl -L -o cmake-3.3.2.tar.gz https://github.com/Kitware/CMake/releases/download/v3.3.2/cmake-3.3.2-Darwin-x86_64.tar.gz
```
pramodk, line 81: all statements must end with `;` !

pramodk (approved):

PR #437, opened 2020-11-25
Fix FindPythonModule for cmake >= 3.19

alkino:
find_package_handle_standard_args is no more usable outside a find_package function
Mimick it with basic if tests


ohm314 (approved):
nice fix 😄

alkino:
Please retest

PR #438, opened 2020-11-27
Use submodules

### LINKED ISSUE #88: CMake should raises an error if submodules are absent
I didn't update submodules of my working copy since eigen has been added. As a result, I experience weird compilation issues. Instead, I would have liked a nice error message  "Error: eigen submodule is missing".

This check could be made at CMake time. CMake would raise an error if a submodule is missing. The message could also provide the `git submodule update --recursive` command to execute.

alkino:
Endly fix #88


cmake/AddCLI11Submodule.cmake
```diff
@@ -0,0 +1,30 @@
...
+  if(NOT ${GIT_FOUND})
+    message(FATAL_ERROR "git not found, clone repository with --recursive")
+  endif()
+  message(STATUS "Sub-module CLI11 missing: running git submodule update --init --recursive")
+  execute_process(
+    COMMAND
+      ${GIT_EXECUTABLE} submodule update --init --recursive --
+      ${PROJECT_SOURCE_DIR}/ext/cli11
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+endif()
```
pramodk, line 28: Could we avoid duplication by keeping this in cmake function? e.g. https://github.com/neuronsimulator/nrn/blob/master/cmake/ExternalProjectHelper.cmake

CMakeLists.txt
```diff
@@ -129,27 +139,13 @@ endif()
...
-message(STATUS "INCLUDING PYBIND11")
-add_subdirectory(ext/pybind11)
-
-# =============================================================================
-# Include fmt library
-# =============================================================================
-message(STATUS "INCLUDING FMT")
-add_subdirectory(${NMODL_PROJECT_SOURCE_DIR}/ext/fmt EXCLUDE_FROM_ALL)
-set_property(TARGET fmt PROPERTY POSITION_INDEPENDENT_CODE ON)
+include_directories(SYSTEM ${THIRD_PARTY_DIRECTORY} ${THIRD_PARTY_DIRECTORY}/fmt/include
```
alkino, line 142: I think we can remove `${THIRD_PARTY_DIRECTORY}` there is nothing right there.

CMakeLists.txt
```diff
@@ -129,27 +139,13 @@ endif()
...
+                    ${THIRD_PARTY_DIRECTORY}/spdlog/include)
 
 # =============================================================================
 # Include path from external libraries
 # =============================================================================
-include_directories(${NMODL_PROJECT_SOURCE_DIR}/ext/cli11/include)
+include_directories(${THIRD_PARTY_DIRECTORY}/cli11/include)
```
alkino, line 148: We can put `cli11` in the same include_directories above as `SYSTEM`, isnt?

CMakeLists.txt
pramodk: IIRC, its needed because source include libraries as `<json/json.hpp>`

CMakeLists.txt
pramodk: yes, I see that you have done it already.

pramodk (approved):

PR #441, opened 2020-11-30
Add compare_blocks for robust sympy visitor comparisons

### LINKED ISSUE #334: 2 tests failing with the new sympy (1.6)
The release of the new [sympy](https://github.com/sympy/sympy/wiki/Release-Notes-for-1.6) broke 2 ctests. We need to investigate this otherwise CIs are gonna fail.
Thanks @iomaganaris to help me detect the problem.

cattabiani:
No more problems with new sympy versions that rearrange the
results producing equivalent but string-different results! Now we
have compare_blocks: it lets sympy itself do the heavylifting
to compare equivalent systems of equations. Noteworthy:

- sanitization: pow(dt, 3) -> dt**3
- brackets: A[0] -> A_0
- tmp variables: back substitution
- pretty printing in case of failure
- simplification
- checking of indentation etc is still preserved!
- ctest-checked with sympy=1.4.*, 1.5*, 1.6*, and 1.7.*

Fix #334


test/unit/visitor/sympy_solver.cpp
```diff
@@ -64,6 +64,92 @@ std::vector<std::string> run_sympy_solver_visitor(
...
 
+void compare_blocks(const std::string& result, const std::string& expected) {
+    using namespace pybind11::literals;
+
+    auto locals = pybind11::dict("s1"_a = result, "s2"_a = expected, "is_equal"_a = false);
+    pybind11::exec(R"(
+                    def compare_blocks(s1, s2):
+                        def sanitize(s):
+                            import re
+                            d = {'\[(\d+)\]':'_\\1',  'pow\((\w+), ?(\d+)\)':'\\1**\\2'}
```
pramodk, line 75: `a nitpick` : would it be possible to expand some variable names? i.e. to avoid single letter var acmes and expand a bit? e.g. d, d1, d2, s.

test/unit/visitor/sympy_solver.cpp
```diff
@@ -64,6 +64,92 @@ std::vector<std::string> run_sympy_solver_visitor(
...
+                                    if parse_expr(f'simplify(({v})-({d2[k]}))'):
+                                        return False
+                            except KeyError:
+                                return False
+
+                            d1.clear()
+                            d2.clear()
+                            return True
+
+                        def reduce(s):
```
pramodk, line 94: some docstrings would be nice, to explain bit of logic

test/unit/visitor/sympy_solver.cpp
```diff
@@ -950,30 +1038,22 @@ SCENARIO("LINEAR solve block (SympySolver Visitor)", "[sympy][linear]") {
...
+                ~ x + 4*c*y = -6*a
+                ~ a + x/b + z - y = 1*b*b
+                ~ 10*x + 13*y - z/(a*a*b) = 14/c
             })";
+        std::string expected_text = R"(
+        LINEAR lin {
+            x = 2*b*(39*pow(a, 3)*b+28*pow(a, 2)*b-2*a*c-3*a+2*pow(b, 2)*c)/(40*pow(a, 2)*pow(b, 2)*c-13*pow(a, 2)*pow(b, 2)+b+4*c)
+            y = (-60*pow(a, 3)*pow(b, 2)*c-14*pow(a, 2)*pow(b, 2)+a*b*c-6*a*c-pow(b, 3)*c)/(c*(40*pow(a, 2)*pow(b, 2)*c-13*pow(a, 2)*pow(b, 2)+b+4*c))
+            z = pow(a, 2)*b*(-40*a*b*pow(c, 2)-47*a*b*c-78*a*c+40*pow(b, 3)*pow(c, 2)-13*pow(b, 3)*c-14*b-56*c)/(c*(40*pow(a, 2)*pow(b, 2)*c-13*pow(a, 2)*pow(b,2)+b+4*c))
+        })";
```
pramodk, line 1051: nice!!!!

`x = (4*pow(a, 2)*pow(b, 2)*...`  looks different than `x = 2*b*(39*pow(a, 3)*b+28*....` but I trust you about they being equal :)

pramodk (commented):
Very nice! 👌

Few minor comments and we should be good to go!

cattabiani:
TODO: remove the limitation on sympy

test/unit/visitor/sympy_solver.cpp
cattabiani: I am using the names that are already in the nmodl print. We get A[0] -> A_0 for example. I have a feeling that you may be confused: the d that you see here means just digit in regex and the replacement is done out of the usual nmodl framework: this is used only in this function. However, if you think A_0 is too short I will add something, np

test/unit/visitor/sympy_solver.cpp
cattabiani: done! However, in clion they will not appear highlghted (because it is python code inside a string). I also added a doxygen explaination for the whole funciton

test/unit/visitor/sympy_solver.cpp
cattabiani: I changed the equations of the test! (from a floating point point of view). It does not change the test overall

cattabiani: I could add some unit-test of this routine so that we know that it also fails

pramodk:
> Katta promised to look later for a way to initialize the python function once :)

I assume this will be done separately. LGTM.

pramodk (approved):

PR #443, opened 2020-12-06
Fix range/state/ion/assigned variable ordering

### LINKED ISSUE #436: Wrong order of ion variables declaration in mod file
When compiling `cal_mig.mod` from https://github.com/Neurosim-lab/netpyne/blob/coreneuron/examples/coreneuron/m1/mod/cal_mig.mod the order of the `cai` and `ica` `ion` channels' variables is wrong in the generated `.cpp` file by `NMODL`.
The issue is fixed if `cai` and `ica` definition as `RANGE` variables in `line 44` is removed.
We need to check again the order of the printing of the `ion` variables in `NMODL` to make sure it's the same as `neuron nmodl`.

pramodk:
- range variabes needs to be printed in the order as it's
   done in neuron/mod2c implementation
 - existing implementation had bugs when ion variables are
   state variables. (CaDynamics_E2.mod cal_mig.mod)
 - this PR reviews implementation in mod2c and documents are
   variable order is calculated in original implementation
   and fixes variable ordering issue.

TODO 

- [x] simplify implementation / code
- [x] document some of the implementation details
- [x] add unit tests for variable ordering

fixes #436

### DIFF
--- a/src/codegen/codegen_helper_visitor.cpp
+++ b/src/codegen/codegen_helper_visitor.cpp
@@ -276,17 +276,87 @@ void CodegenHelperVisitor::find_non_range_variables() {
 /**
  * Find range variables i.e. ones that are belong to per instance allocation
  *
- * In order to be compatible with NEURON, we need to print range variables in
- * certain order. For example, range variables which are parameters comes first.
- * Also, there is difference between declaration order vs. definition order. For
- * example, POINTER variable in NEURON block is just declaration and doesn't
- * determine the order in which they will get printed. Below we query symbol table
- * and order all instance variables into certain order.
+ * In order to be compatible with NEURON, we need to print all variables in
+ * exact order as NEURON/MOD2C implementation. This is important because memory
+ * for all vatiables is allocated in single 1-D array with certain offset
+ * for each variable. The order of variables determine the offset and hence
+ * they must be in same order as NEURON.
+ *
+ * Here is how order is determined into NEURON/MOD2C implementation:
+ *
+ * First, following three lists are created
+ * - variables with parameter and range property (List 1)
+ * - variables with state and range property (List 2)
+ * - variables with assigned and range property (List 3)
+ *
+ * Once above three lists are created, we remove following variable from above
+ * three lists:
+ * - In NEURON/MOD2C implementation, we remove variables with NRNPRANGEIN
+ *   or NRNPRANGEOUT type
+ * - So who has NRNPRANGEIN and NRNPRANGEOUT type? these are such USEION read
+ *   or write variables that are not ionic currents.
+ * - This is the reason for mod files CaDynamics_E2.mod or cal_mig.mod, ica variable
+ *   is printed earlier in the list but other variables like cai, cao don't appear
+ *   in same order.
+ *
+ * Finally we create 4th list:
+ *  - variables with assigned property and not in the previous 3 lists
+ *
+ * We now print the variables in following order:
+ *
+ * - List 1 i.e. range + parameter variables are printed first
+ * - List 3 i.e. range + assigned variables are printed next
+ * - List 2 i.e. range + state variables are printed next
+ * - List 4 i.e. assigned variables not present in previous 3 lists
+ *
+ * NOTE:
+ * - State variables are also has property `assigned_definition` but these variables
+ *   are not from ASSIGNED block.
+ * - Variable can not be range as well as state, it's redeclaration error
+ * - Variable can be parameter as well as range. Without range, parameter
+ *   is considered as global variable i.e. one value for all instances.
+ * - If a variable is only defined as RANGE and not in assgined or parameter
+ *   or state block then it's not printed.
+ * - Note that if variable property is different than the variable type. For example,
+ *   if variable has range property, it doesn't mean the variable is declared as RANGE.
+ *   Other variables like STATE and ASSIGNED block variables also get range property.
+ * - Also, there is difference between declaration order vs. definition order. For
+ *   example, POINTER variable in NEURON block is just declaration and doesn't
+ *   determine the order in which they will get printed. Below we query symbol table
+ *   and order all instance variables into certain order.
  */
 void CodegenHelperVisitor::find_range_variables() {
     /// comparator to decide the order based on definition
     auto comparator = [](const SymbolType& first, const SymbolType& second) -> bool {
         return first->get_definition_order() < second->get_definition_order();
+    };
+
+    /// from symbols vector `vars`, remove all ion variables which are not ionic currents
+    auto remove_non_ioncur_vars = [](SymbolVectorType& vars, CodegenInfo& info) -> void {
+        vars.erase(std::remove_if(vars.begin(),
+                                  vars.end(),
+                                  [&](SymbolType& s) {
+                                      return info.is_ion_variable(s->get_name()) &&
+                                             !info.is_ionic_current(s->get_name());
+                                  }),
+                   vars.end());
+    };
+
+    /// if `seconday` vector contains any symbol that exist in the `primary` then remove ut
+    auto remove_var_exist = [](SymbolVectorType& primary, SymbolVectorType& secondary) -> void {
+        secondary.erase(std::remove_if(secondary.begin(),
+                                       secondary.end(),
+                                       [&primary](const SymbolType& tosearch) {
+                                           return std::find_if(primary.begin(),
+                                                               primary.end(),
+                                                               // compare by symbol name
+                                                               [&tosearch](
+                                                                   const SymbolType& symbol) {
+                                                                   return tosearch->get_name() ==
+                                                                          symbol->get_name();
+                                                               }) != primary.end();
+                                       }),
+                        secondary.end());
     };
 
     /**
@@ -299,8 +369,10 @@ void CodegenHelperVisitor::find_range_variables() {
                    | NmodlType::pointer_var
                    | NmodlType::bbcore_pointer_var
                    | NmodlType::state_var;
+
     // clang-format on
     info.range_parameter_vars = psymtab->get_variables(with, without);
+    remove_non_ioncur_vars(info.range_parameter_vars, info);
     std::sort(info.range_parameter_vars.begin(), info.range_parameter_vars.end(), comparator);
 
     /**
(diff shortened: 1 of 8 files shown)


src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
  *
- * In order to be compatible with NEURON, we need to print range variables in
- * certain order. For example, range variables which are parameters comes first.
- * Also, there is difference between declaration order vs. definition order. For
- * example, POINTER variable in NEURON block is just declaration and doesn't
- * determine the order in which they will get printed. Below we query symbol table
- * and order all instance variables into certain order.
+ * In order to be compatible with NEURON, we need to print all variables in
+ * exact order as NEURON/MOD2C implementation. This is important because memory
+ * for all vatiables is allocated in single 1-D array with certain offset
```
cattabiani, line 281:
```suggestion
 * for all variables is allocated in single 1-D array with certain offset
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ *
+ * We now print the variables in following order:
+ *
+ * - List 1 i.e. range + parameter variables are printed first
+ * - List 3 i.e. range + assigned variables are printed next
+ * - List 2 i.e. range + state variables are printed next
+ * - List 4 i.e. assigned variables not present in previous 3 lists
+ *
+ * NOTE:
+ * - State variables are also has property `assigned_definition` but these variables
```
cattabiani, line 313:
```suggestion
 * - State variables also have the property `assigned_definition` but these variables
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+                                                               // compare by symbol name
+                                                               [&tosearch](
+                                                                   const SymbolType& symbol) {
+                                                                   return tosearch->get_name() ==
+                                                                          symbol->get_name();
+                                                               }) != primary.end();
+                                       }),
+                        secondary.end());
+    };
+
```
pramodk, line 361: @tristan0x : is there any better pattern to achieve above? Or above is "fine"?

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+                                                               primary.end(),
+                                                               // compare by symbol name
+                                                               [&tosearch](
+                                                                   const SymbolType& symbol) {
+                                                                   return tosearch->get_name() ==
+                                                                          symbol->get_name();
+                                                               }) != primary.end();
+                                       }),
+                        secondary.end());
+    };
```
alkino, line 360: `auto remote_var_exist = secondary | erase_if([&](const auto& symbol) { return tosearch->get_name() == symbol->get_name();`

alkino: it's not really the right syntax

pramodk: interesting! I will see the range-v3 (for future integration)

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ * they must be in same order as NEURON.
+ *
+ * Here is how order is determined into NEURON/MOD2C implementation:
+ *
+ * First, following three lists are created
+ * - variables with parameter and range property (List 1)
+ * - variables with state and range property (List 2)
+ * - variables with assigned and range property (List 3)
+ *
+ * Once above three lists are created, we remove following variable from above
```
cattabiani, line 292:
```suggestion
 * Once created, we remove following variable from above
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ * three lists:
```
cattabiani, line 293:
```suggestion
 * three lists:
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ * - In NEURON/MOD2C implementation, we remove variables with NRNPRANGEIN
+ *   or NRNPRANGEOUT type
+ * - So who has NRNPRANGEIN and NRNPRANGEOUT type? these are such USEION read
```
cattabiani, line 296:
```suggestion
 * - So who has NRNPRANGEIN and NRNPRANGEOUT type? these are USEION read
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ * - This is the reason for mod files CaDynamics_E2.mod or cal_mig.mod, ica variable
+ *   is printed earlier in the list but other variables like cai, cao don't appear
+ *   in same order.
+ *
+ * Finally we create 4th list:
+ *  - variables with assigned property and not in the previous 3 lists
+ *
+ * We now print the variables in following order:
+ *
+ * - List 1 i.e. range + parameter variables are printed first
```
cattabiani, line 307: nitpick: shouldn't we create the lists in the same order we print them?

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ * - So who has NRNPRANGEIN and NRNPRANGEOUT type? these are such USEION read
+ *   or write variables that are not ionic currents.
+ * - This is the reason for mod files CaDynamics_E2.mod or cal_mig.mod, ica variable
+ *   is printed earlier in the list but other variables like cai, cao don't appear
+ *   in same order.
+ *
+ * Finally we create 4th list:
+ *  - variables with assigned property and not in the previous 3 lists
+ *
+ * We now print the variables in following order:
```
cattabiani, line 305: nitpick: shouldn't we create the lists in the same order we print them?

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ * - List 2 i.e. range + state variables are printed next
+ * - List 4 i.e. assigned and ion variables not present in the previous 3 lists
+ *
+ * NOTE:
+ * - State variables also have the property `assigned_definition` but these variables
+ *   are not from ASSIGNED block.
+ * - Variable can not be range as well as state, it's redeclaration error
+ * - Variable can be parameter as well as range. Without range, parameter
+ *   is considered as global variable i.e. one value for all instances.
+ * - If a variable is only defined as RANGE and not in assgined or parameter
```
cattabiani, line 318:
```suggestion
 * - If a variable is only defined as RANGE and not in assigned or parameter
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ *   or state block then it's not printed.
+ * - Note that if variable property is different than the variable type. For example,
```
cattabiani, line 320:
```suggestion
 * - Note that a variable property is different than the variable type. For example,
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+        vars.erase(std::remove_if(vars.begin(),
+                                  vars.end(),
+                                  [&](SymbolType& s) {
+                                      return info.is_ion_variable(s->get_name()) &&
+                                             !info.is_ionic_current(s->get_name());
+                                  }),
+                   vars.end());
+    };
+
+    /// if `seconday` vector contains any symbol that exist in the `primary` then remove it
```
cattabiani, line 345:
```suggestion
    /// if `secondary` vector contains any symbol that exist in the `primary` then remove it
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+    auto remove_var_exist = [](SymbolVectorType& primary, SymbolVectorType& secondary) -> void {
```
cattabiani, line 346: another implementation is to stuff a hashtable with the values of primary and then run over the second and check if they are present in the hash table. This has complexity O(N_p + N_s) on average (N_p = primary.size(), N_s = primary.size()). Yours is O(N_p * N_s). However, hashing has a big constant in front so for small vectors your version could come on top. I would prefer the most O -complexity efficient since in case of small vectors the difference is not noticible anyway. Ultimately your decision

src/codegen/codegen_helper_visitor.cpp
cattabiani: I had a suggestion for better patterns :D

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ *   and order all instance variables into certain order.
  */
 void CodegenHelperVisitor::find_range_variables() {
     /// comparator to decide the order based on definition
     auto comparator = [](const SymbolType& first, const SymbolType& second) -> bool {
         return first->get_definition_order() < second->get_definition_order();
     };
 
+    /// from symbols vector `vars`, remove all ion variables which are not ionic currents
+    auto remove_non_ioncur_vars = [](SymbolVectorType& vars, CodegenInfo& info) -> void {
```
cattabiani, line 335:
```suggestion
    auto remove_non_ioncur_vars = [](SymbolVectorType& vars, const CodegenInfo& info) -> void {
```

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -615,14 +654,14 @@ void CodegenHelperVisitor::visit_program(const ast::Program& node) {
         }
     }
     node.visit_children(*this);
+    find_ion_variables();
```
cattabiani, line 657: nitpick:
```suggestion
    find_ion_variables(); // Keep this before find_*_range_variables()
```

test/unit/codegen/codegen_helper.cpp
```diff
@@ -0,0 +1,130 @@
...
+              cai' = -(10000)*(ica*gamma/(2*FARADAY*depth)) - (cai - minCai)/decay
+            }
+        )";
+
+        THEN("ion state variable is ordered after parameter and assigned ionic current") {
+            std::string expected = "gamma;decay;depth;minCai;ica;cai;";
+            auto result = run_inline_visitor(nmodl_text);
+            REQUIRE(result == expected);
+        }
+    }
```
cattabiani, line 129: another edge-case that could be useful to test is `thalamus/mod/ican.mod`. In there there is an ion called `n` and its current could give some problems to the ion current detection system and ispc (since `in` is a reserved variable in there)

cattabiani (commented):
Very good! A few nitpicks and a suggestion for improving efficiency

test/unit/codegen/codegen_helper.cpp
pramodk: thanks, I will add that shortly.

pramodk: ican.mod is handled normally in the context of this PR : 

image

This mod file has issue due to `in` variable which is keyword in ISPC. This will be handled separately.

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -276,19 +276,89 @@ void CodegenHelperVisitor::find_non_range_variables() {
...
+ *  - variables with assigned property and not in the previous 3 lists
+ *
+ * We now print the variables in following order:
+ *
+ * - List 1 i.e. range + parameter variables are printed first
+ * - List 3 i.e. range + assigned variables are printed next
+ * - List 2 i.e. range + state variables are printed next
+ * - List 4 i.e. assigned and ion variables not present in the previous 3 lists
+ *
+ * NOTE:
```
iomaganaris, line 311: I suppose that all of the notes below apply to mod2c/neuron as well, right?

test/unit/codegen/codegen_helper.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    GIVEN("CaDynamics_E2.mod : USEION variables declared as STATE variable") {
+        std::string nmodl_text = R"(
+            NEURON  {
+              SUFFIX CaDynamics_E2
+              USEION ca READ ica WRITE cai
```
iomaganaris, line 102: I think that there are some mod files that have the same variable in `READ` and `WRITE`? What happens with those ones? Maybe a test would be useful?
Looking at the code I think that they are handled the same way. The only issue might be that they are declared 2 times?

iomaganaris (commented):
LGTM in general.
I have a couple of questions only.

test/unit/codegen/codegen_helper.cpp
pramodk: cadyn.mod unit test added. Its handled correctly.

iomaganaris (approved):

cattabiani:
I am Katta and I approve this PR :D

cattabiani (approved):

PR #444, opened 2020-12-08
Bump pybind11 to v2.6.1

alkino:
update pybind11 to better support python 3.9


alkino:
Please retest

pramodk:
Curious - is there specific in pybind v2.6.1 ?

alkino:
It does not seem, it's only that flex is installed, after that azure try to launch `brew clean` it fails and `bison` is never installed.

alkino:
pybind11 > 3.5.0 ask for a cmake >= 3.4, and we wish to stay with cmake 3.3 as long as necessary.

PR #450, opened 2020-12-11
Fix issue with double constants in ISPC generated files

### LINKED ISSUE #442: ISPC codegen backend and floating point values
ISPC backend needs to print floating point values in different form because floating point values are treated as `float` by default. From [ISPC docs](https://ispc.github.io/ispc.html):

```
Updating ISPC Programs For Changes In ISPC 1.5.0

This release adds support for double precision floating point constants.
Double precision floating point constants are floating point number with
d suffix and optional exponent part. Here are some examples:
3.14d, 31.4d-1, 1.d, 1.0d, 1d-2. Note that floating point number without
suffix is treated as single precision constant.
```

For this reason, ISPC has `double_to_string` routine to convert floating point values into ISPC compatible form:   https://github.com/BlueBrain/nmodl/blob/ac46928a78cba40fa97900835814cfc765b143ff/src/codegen/codegen_ispc_visitor.cpp#L126

But as observed in #431 and #144, this could result in discrepancies in C++ backend and ISPC backend.

As part of this ticket we need to review the implementation and see if values calculated for units is same/similar to CPP backend.

One option could be [hex representation](https://ispc.github.io/ispc.html#lexical-structure): 

```
Floating-point constants can be specified in one of three ways. First, they may be a sequence of zero or
more digits from 0 to 9, followed by a period, followed by zero or more digits from 0 to 9. (There must be
at least one digit before or after the period).

The second option is scientific notation, where a base value is specified as the first form of a floating-point
constant but is then followed by an "e" or "E", then a plus sign or a minus sign, and then an exponent.

Finally, floating-point constants may be specified as hexadecimal constants; this form can ensure a perfectly
bit-accurate representation of a particular floating-point number. These are specified with an "0x" prefix,
followed by a zero or a one, a period, and then the remainder of the mantissa in hexadecimal form, with digits
from 0-9, a-f, or A-F. The start of the exponent is denoted by a "p", which is then followed by an optional plus
or minus sign and then digits from 0 to 9. For example:

float two = 0x1p+1;  // 2.0
float pi  = 0x1.921fb54442d18p+1;  // 3.1415926535...
float neg = -0x1.ffep+11;  // -4095.

Floating-point constants can optionally have a "f" or "F" suffix (ispc currently treats all floating-point
constants as having 32-bit precision, making this suffix not currently have an effect.)
```

iomaganaris:
- Fixes double/float constant representation in ISPC generated files to be similar to `c++` backend
- Adds ISPC codegen unit test
- Small fixes in ISPC visitor to create the ISPC codegen unit test

Fixes #442 

Note: the floating point literals in `C++` (and in `NMODL` as well) are represented by digits and `[e|E]`
In NMODL we have: https://github.com/BlueBrain/nmodl/blob/85aeac62ef7a1e748f0d3e60510ded545b2b1cec/src/lexer/nmodl.ll#L42
https://github.com/BlueBrain/nmodl/blob/85aeac62ef7a1e748f0d3e60510ded545b2b1cec/src/lexer/nmodl.ll#L246


test/unit/codegen/codegen_ispc.cpp
```diff
@@ -0,0 +1,199 @@
...
+                    double v = voltage[node_id];
+                    
+                    double x, y;
+                    x = 1d-18 + FARADAY * 1.2345d;
+                    y = 1d+18 + FARADAY * 0.1234d;
+                    inst->a[id] = x * 1.012345678901234567d + y;
+                    inst->b[id] = inst->a[id] + 1.0d + 2.0d;
+                }
+            }
+        )";
```
pramodk, line 187: Overall very nice!

I am thinking about this unit test and if it could be made smaller / simple : `float_to_string` does simple job and it would be nice if it could be tested without this whole Codegen backend. And also, whenever we change how we print initial / breakpoint block then we have to change this test.

Could we move this float_to_string to some utils or codegen_helper namespace and then just unit tests those functions?

iomaganaris: That's a valid point. I also though testing only the function `float_to_string` however I thought that having a test for the whole backend would also be useful and also test this function.
If we change something in the code that changes the output file then we should be careful and also alter this test if needed, right?

src/codegen/codegen_c_visitor.cpp
```diff
@@ -440,11 +440,13 @@ int CodegenCVisitor::position_of_int_var(const std::string& name) const {
...
  * and neuron where ".0" is appended). Otherwise we print double variables as
- * they are represented in the mod file by user.
+ * they are represented in the mod file by user. If the value is in scientific
+ * representation (1e+20, 1E-15) then keep it as it is.
  */
 std::string CodegenCVisitor::double_to_string(const std::string& s_value) {
     double value = std::stod(s_value);
-    if (std::ceil(value) == value) {
+    if (std::ceil(value) == value && s_value.find('E') == std::string::npos &&
+        s_value.find('E') == std::string::npos) {
```
cattabiani, line 449: You repeat twice the same check. Maybe you wanted to check 'e'?
```suggestion
        s_value.find('e') == std::string::npos) {
```

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -124,54 +124,46 @@ void CodegenIspcVisitor::visit_local_list_statement(const ast::LocalListStatemen
...
-    if (std::ceil(value) == value) {
-        return "{:.1f}d"_format(value);
-    }
-    if ((value <= 1.0) && (value >= -1.0)) {
-        return "{:.16f}d"_format(value);
+    std::string return_string = s_value;
+    if (s_value.find('E') != std::string::npos) {
+        std::replace(return_string.begin(), return_string.end(), 'E', 'd');
+    } else if (s_value.find('e') != std::string::npos) {
+        std::replace(return_string.begin(), return_string.end(), 'e', 'd');
```
cattabiani, line 139: You can match any char with:

```suggestion
    if (s_value.find_first_of("eE") != std::string::npos) {
        std::replace(return_string.begin(), return_string.end(), 'E', 'd');
        std::replace(return_string.begin(), return_string.end(), 'e', 'd');
```

src/codegen/codegen_ispc_visitor.cpp
```diff
@@ -124,54 +124,46 @@ void CodegenIspcVisitor::visit_local_list_statement(const ast::LocalListStatemen
...
 std::string CodegenIspcVisitor::float_to_string(const std::string& s_value) {
-    float value = std::stof(s_value);
-    if (std::ceil(value) == value) {
-        return "{:.1f}"_format(value);
-    }
-    if ((value <= 1.0f) && (value >= -1.0f)) {
-        return "{:.6f}"_format(value);
+    std::string return_string = s_value;
+    if (s_value.find('E') == std::string::npos && s_value.find('e') == std::string::npos &&
+        s_value.find('.') == std::string::npos) {
```
cattabiani, line 159: as before:

```suggestion
    if (s_value.find_first_of("Ee.") == std::string::npos) {
```

test/unit/codegen/codegen_ispc.cpp
cattabiani: 2 tests? We keep this, we move float_to_string in utils and we also test that alone?

test/unit/codegen/codegen_ispc.cpp
```diff
@@ -0,0 +1,199 @@
...
+            NEURON {
+                SUFFIX unit_test
+                RANGE a, b
+            }
+            ASSIGNED {
+                a
+                b
+            }
+            UNITS {
+                FARADAY = (faraday) (coulomb)
```
cattabiani, line 122: IIRC this does not cover the whole picture. We should also test:
- R constant
- The possibility to ask for kiloculombs: FARADAY = (faraday) (kilocoulombs)
- Custom value with custom units like: R = 8.3e-3 (kilojoule/degC)
- other constants?
- Do we check with AND without legacy units?

cattabiani (commented):
Very nice! A few suggestions and a big question on unit tests

test/unit/codegen/codegen_ispc.cpp
iomaganaris: All the tests related to the units definition are done in https://github.com/BlueBrain/nmodl/blob/master/test/unit/units/parser.cpp
Here my intention was to just check if the values for the units are printed with the expected form.
I believe the other tests cover all the cases and we don't have to check this also here

test/unit/codegen/codegen_ispc.cpp
iomaganaris: Okay, I agree. I will take care of this. Thank you!

src/codegen/codegen_c_visitor.cpp
```diff
@@ -440,11 +440,13 @@ int CodegenCVisitor::position_of_int_var(const std::string& name) const {
...
  * and neuron where ".0" is appended). Otherwise we print double variables as
- * they are represented in the mod file by user.
+ * they are represented in the mod file by user. If the value is in scientific
+ * representation (1e+20, 1E-15) then keep it as it is.
  */
 std::string CodegenCVisitor::double_to_string(const std::string& s_value) {
     double value = std::stod(s_value);
-    if (std::ceil(value) == value) {
+    if (std::ceil(value) == value && s_value.find('E') == std::string::npos &&
+        s_value.find('e') == std::string::npos) {
```
cattabiani, line 449: as in other comments. Iforgot to change it here
```suggestion
    if (std::ceil(value) == value && s_value.find_first_of("Ee") == std::string::npos) {
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -453,7 +455,8 @@ std::string CodegenCVisitor::double_to_string(const std::string& s_value) {
 
 std::string CodegenCVisitor::float_to_string(const std::string& s_value) {
     float value = std::stof(s_value);
-    if (std::ceil(value) == value) {
+    if (std::ceil(value) == value && s_value.find('E') == std::string::npos &&
+        s_value.find('e') == std::string::npos) {
```
cattabiani, line 459: as before
```suggestion
    if (std::ceil(value) == value && s_value.find_first_of("Ee") == std::string::npos) {
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -440,11 +440,13 @@ int CodegenCVisitor::position_of_int_var(const std::string& name) const {
  * then it gets printed as an integer. To avoid this, we use below wrapper.
  * If user has provided integer then it gets printed as 1.0 (similar to mod2c
  * and neuron where ".0" is appended). Otherwise we print double variables as
- * they are represented in the mod file by user.
+ * they are represented in the mod file by user. If the value is in scientific
+ * representation (1e+20, 1E-15) then keep it as it is.
  */
 std::string CodegenCVisitor::double_to_string(const std::string& s_value) {
```
cattabiani, line 446: it is still a little puzzling that double_to_string gets a string and not a double for conversion. Maybe we should find a better name... dunno

src/codegen/codegen_c_visitor.cpp
iomaganaris: Yy. I already took care of this in another commit which I haven't pushed yet :+1:

src/codegen/codegen_c_visitor.cpp
iomaganaris: Same

src/codegen/codegen_c_visitor.cpp
iomaganaris: It's a bit puzzling at first, I agree. I am going to add some more documentation about that to explain why there is a string there.
I cannot think any better name for this tbh :confused:

cattabiani: reformat_stringed_double?

cattabiani (approved):

src/codegen/codegen_c_visitor.hpp
```diff
@@ -575,15 +575,15 @@ class CodegenCVisitor: public visitor::ConstAstVisitor {
 
     /**
      * Convert a given \c double value to its string representation
-     * \param value The number to convert
+     * \param value The number to convert given as string as it parsed by the modfile
```
cattabiani, line 578:
```suggestion
     * \param value The number to convert given as string as it is parsed by the modfile
```

src/codegen/codegen_c_visitor.hpp
```diff
@@ -575,15 +575,15 @@ class CodegenCVisitor: public visitor::ConstAstVisitor {
...
      * \return      Its string representation
      */
     virtual std::string double_to_string(const std::string& value);
 
 
     /**
      * Convert a given \c float value to its string representation
-     * \param value The number to convert
+     * \param value The number to convert given as string as it parsed by the modfile
```
cattabiani, line 586:
```suggestion
     * \param value The number to convert given as string as it is parsed by the modfile
```

src/printer/code_printer.hpp
```diff
@@ -47,10 +47,10 @@ class CodePrinter {
 
   public:
     CodePrinter()
-        : result(new std::ostream(std::cout.rdbuf())) {}
+        : result(std::make_shared<std::ostream>(std::cout.rdbuf())) {}
```
cattabiani, line 50: nice!

test/unit/codegen/codegen_utils.cpp
```diff
@@ -0,0 +1,218 @@
...
+            REQUIRE(nmodl_constant_result == float_constant_with_e_minus);
+            nmodl_constant_result = codegen::utils::float_to_string<CodegenIspcVisitor>(
+                float_constant_with_decimal_point_e);
+            REQUIRE(nmodl_constant_result == float_constant_with_decimal_point_e);
+            nmodl_constant_result = codegen::utils::float_to_string<CodegenIspcVisitor>(
+                float_constant_with_E);
+            REQUIRE(nmodl_constant_result == float_constant_with_E);
+        }
+    }
+}
```
cattabiani, line 218: can we test also -0 and +0? Not sure it is so important. Otherwise, very good!

cattabiani (approved):

cattabiani:
@iomaganaris how does this interact with CONSTANT?

src/codegen/codegen_c_visitor.cpp
iomaganaris: `format_double_string` and `format_float_string` may be even better?

test/unit/codegen/codegen_utils.cpp
iomaganaris: `-` and `+` are parsed by the lexer/parser and they are not passed to these functions, so `0` will be handled like `1`

iomaganaris:
> @iomaganaris how does this interact with CONSTANT?

IIRC `CONSTANT` blocks are not handled yet, so there shouldn't be any conflict with those

src/codegen/codegen_c_visitor.hpp
```diff
@@ -1888,7 +1903,7 @@ void CodegenCVisitor::print_vector_elements(const std::vector<T>& elements,
     for (auto iter = elements.begin(); iter != elements.end(); iter++) {
         printer->add_text(prefix);
         (*iter)->accept(*this);
-        if (!separator.empty() && !utils::is_last(iter, elements)) {
+        if (!separator.empty() && !nmodl::utils::is_last(iter, elements)) {
```
pramodk, line 1906: same

test/unit/codegen/codegen_utils.cpp
```diff
@@ -0,0 +1,218 @@
...
+            REQUIRE(nmodl_constant_result == float_constant_with_e);
+            nmodl_constant_result = codegen::utils::format_float_string<CodegenIspcVisitor>(
+                float_constant_with_e_minus);
+            REQUIRE(nmodl_constant_result == float_constant_with_e_minus);
+            nmodl_constant_result = codegen::utils::format_float_string<CodegenIspcVisitor>(
+                float_constant_with_decimal_point_e);
+            REQUIRE(nmodl_constant_result == float_constant_with_decimal_point_e);
+            nmodl_constant_result = codegen::utils::format_float_string<CodegenIspcVisitor>(
+                float_constant_with_E);
+            REQUIRE(nmodl_constant_result == float_constant_with_E);
```
pramodk, line 215: Just a nitpick : as this pattern is getting repeated a lot in this file, one can think of wrapper function that accepts input vector (of numbers as strings) and output vector (of numbers as string) and then you iterate on each element of vector and compare the with an element of output vector. This will reduce the code in this test ?  (we use similar approach in [the parser tests](https://github.com/BlueBrain/nmodl/blob/1089bc1509e7ae8cc9a76e0d527c3f738014c47b/test/unit/parser/parser.cpp#L136)).

pramodk (commented):
Very nice!  Just have one or two minor comments/questions. Could you take a look?

src/codegen/codegen_c_visitor.cpp
```diff
@@ -990,7 +984,7 @@ std::string CodegenCVisitor::get_parameter_str(const ParamVector& params) {
                                     std::get<1>(*iter),
                                     std::get<2>(*iter),
                                     std::get<3>(*iter));
-        if (!utils::is_last(iter, params)) {
+        if (!nmodl::utils::is_last(iter, params)) {
```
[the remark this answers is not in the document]
iomaganaris: There are 2 `utils` namespaces.
There is the `nmodl::utils` which includes general utility functions in `nmodl`.
And there is `nmodl::codegen::utils` that includes the utility functions only for the `codegen`.
So, in this case we have to explicitly use the full `namespace` name to disambiguate them.

test/unit/codegen/codegen_utils.cpp
iomaganaris: I was thinking something similar in the beginning when I didn't have a lot of checks, so I thought it wasn't worth it. Then I started adding cases though, so it seems like a good idea now :stuck_out_tongue:

iomaganaris: Done. I hope you think it's better like this

src/codegen/codegen_c_visitor.cpp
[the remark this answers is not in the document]
pramodk: Ahh didn’t realise that! Wouldn’t it be confusing ? i.e. should we rename to codegen_utils or something ? Just asking because I don’t  know what is current best practices is. @alkino?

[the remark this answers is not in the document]
alkino: The best practice is to always give full namespace path ^^

Maybe we can group functions in only one namespace?

Or not create a `is_last` and use something more "STL".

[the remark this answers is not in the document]
alkino: If you want the codegen_utils, you only want to do
`using codegen_utils = namespace nmodl::codegen::utils`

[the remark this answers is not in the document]
pramodk: Ok

pramodk (approved):
LGTM

Issue #458, opened 2020-12-15
MOD file compatibility checker as a standalone tool


pramodk:
NMODL handles most of the features we commonly required with CoreNEURON. But there can be cases where some mod files from user can't be handled by NMODL for code generation. For example [not all solvers from MOD2C are implemented](https://github.com/BlueBrain/nmodl/blob/1089bc1509e7ae8cc9a76e0d527c3f738014c47b/src/codegen/codegen_compatibility_visitor.hpp#L56) or [not all NMODL constructs are handled](https://github.com/BlueBrain/nmodl/blob/1089bc1509e7ae8cc9a76e0d527c3f738014c47b/src/codegen/codegen_compatibility_visitor.cpp#L21).

Currently, when nmodl is called via nrnivmodl-core, we [check if mod file is compatible](https://github.com/BlueBrain/nmodl/blob/1089bc1509e7ae8cc9a76e0d527c3f738014c47b/src/nmodl/main.cpp#L363) with NMODL otherwise we error / return exit code 1. With this, one can identify errors but user can easily get lost while compiling large number of mod files with `nrnivmodl-core -j`.

To avoid this, it would be helpful to have ability to launch nmodl **only to check** if mod file is compatible with code generation or not. With this, we can update nrnivmodl-core workflow to do following:
* first launch nmodl to check if all mod files provided by user are compatible.
* in case of incompatibilities, nmodl print corrections and we stop
* in the absence of errors we continue with the code generation for all mod files

What needs to be done here:

* Add new CLI option to run only codegen compatibility pass
* With this new CLI option on, we will only check if mod file is compatible with NMODL.
  * if compatible, return 0 as exit code
  * if not compatible, return 1 as exit code

cc: @iomaganaris

PR #459, opened 2020-12-15
Add CLI flag only-check-compatibility

cattabiani:
We return right after checking compatibility without generating code

Useful for having coreneuron switch to mod2c in case mod file is not compatible

Fix: #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ codegen
     --layout TEXT:{aos,soa}=soa           Memory layout for code generation
     --datatype TEXT:{float,double}=soa    Data type for floating point variables
     --force                               Force code generation even if there is any code incompatibility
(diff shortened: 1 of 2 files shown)


cattabiani:
if `--only-check-compatibility` is too long just suggest a better name and I will change it. This name has the advantage that is quite explicative. Considering that its main use is inside coreneuron and not for users I do not think it is a problem

iomaganaris (commented):
LGTM
Just add the documentation in the `README.md` of the root directory where all the `CLI` options are listed.

cattabiani:
> LGTM
> Just add the documentation in the `README.md` of the root directory where all the `CLI` options are listed.

done!

iomaganaris (approved):
Nice! Thank you

PR #460, opened 2020-12-16
Add a --quiet option to remove all logs

alkino:


iomaganaris (approved):
LGTM
Could you just please add this option in the root `README.md` to keep all the options in place?

src/nmodl/main.cpp
```diff
@@ -158,7 +161,10 @@ int main(int argc, const char* argv[]) {
     app.get_formatter()->column_width(40);
     app.set_help_all_flag("-H,--help-all", "Print this help message including all sub-commands");
 
-    app.add_flag("-v,--verbose", verbose, "Verbose logger output")->ignore_case();
+    auto v_flag = app.add_flag("-v,--verbose", verbose, "Verbose logger output")->ignore_case();
+    auto q_flag = app.add_flag("-q,--quiet", quiet, "Remove logger output")->ignore_case();
+    v_flag->excludes(q_flag);
```
pramodk, line 166: @alkino : is there a way to use existing verbose flag for the same purpose? (i.e. having different levels of verbose enabled?)

alkino: --quiet --verbose=2 --debug?
Those 3 flags?

alkino:
maybe --quiet should always print warning and above ?

src/nmodl/main.cpp
pramodk: no, just verbose with different levels and we can get rid of --debug and --quiet.

pramodk:
> maybe --quiet should always print warning and above ?

Make sense to me.

pramodk (approved):
LGTM, thanks!

PR #461, opened 2020-12-17
Attempt to fix nightly tag

ohm314:
Nightly build fails because the tag is not correctly set. This PR should fix it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,13 +2,12 @@
 # Azure Pipeline NMODL settings
 #=============================================================================
 variables:
-- name: ReleaseWheelBuild
-  value: false
-- name: NightlyWheelBuild
-  value: false
+  # defined in UI
+  #- ReleaseWheelBuild: False
+  #- NightlyWheelBuild: False
 - name: buildWheel
   value: $[or(in(variables['Build.Reason'], 'Schedule'),
-              and(in(variables['Build.Reason'], 'Manual'), eq(variables.NightlyWheelBuild, true)))]
+              and(in(variables['Build.Reason'], 'Manual'), eq(variables.NightlyWheelBuild, True)))]
 
 # Nightly build master for pypi upload
 schedules:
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -142,29 +141,32 @@ jobs:
...
         -e NMODL_NIGHTLY_TAG=$TAG \
         'bluebrain/nmodl:wheel' \
         packaging/build_wheels.bash linux $(python.version)
-    condition: and(succeeded(), or(eq(variables.buildWheel, true), eq(variables['python.version'], '3.9')))
+    condition: and(succeeded(), or(eq(variables.buildWheel, True), eq(variables['python.version'], '3.9')))
     displayName: 'Building ManyLinux Wheel'
+  #- task: PublishBuildArtifacts@1
+  #  inputs:
+  #    pathToPublish: '$(Build.SourcesDirectory)/wheelhouse'
+  #  displayName: 'Publish wheel as build artifact'
```
pramodk, line 163: I would uncomment this

azure-pipelines.yml
```diff
@@ -2,21 +2,20 @@
 # Azure Pipeline NMODL settings
 #=============================================================================
 variables:
-- name: ReleaseWheelBuild
-  value: false
-- name: NightlyWheelBuild
-  value: false
+  # defined in UI
+  #- ReleaseWheelBuild: False
+  #- NightlyWheelBuild: False
```
pramodk, line 7: I think we can remove this

ohm314: OK, I thought I leave it more for documenting, this way we know that they are defined in UI and can be used inside yaml

azure-pipelines.yml
ohm314: You want to always produce artefacts?

pramodk: yes, that's we typically do. Artefacts get deleted automatically after certain time by Azure automatically .

pramodk (approved):
LGTM

PR #466, opened 2020-12-20
Disable extra nmodl targets if used as sub-module

pramodk:
- nmodl has multiple utility lexer/parsers
  - when nmodl is used as submodule in neuron or
    coreneuron then disable these targets


pramodk:
this is required to update submodules in coreneuron and then in neuron. So will go ahead with this minor change.

PR #470, opened 2020-12-24
FunctionBlock code generation and terminator checks

georgemitenkov:
This PR adds support for LLVM code generation for `FunctionBlock`. Since the code is almost identical to `ProcedureBlock`'s codegen, it was refactored into a helper function `visit_procedure_or_function()`. The differences are:
- Procedure returns void, Function returns a dummy double 0.0 (similarly to C visitor).
- Procedure has a void return type in its signature, and function has a double return type.

Also, tests were altered to enforce the existence of the block terminator (`ret` instruction) in both procedure and function.

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -47,7 +47,18 @@ void CodegenLLVMVisitor::visit_procedure_or_function(const ast::Block& node) {
     llvm::BasicBlock* body = llvm::BasicBlock::Create(*context, /*Name=*/"", func);
     builder.SetInsertPoint(body);
 
-    // First, allocate parameters on the stack and add them to the symbol table.
+    // When processing a function, it returns a value named ret_<function_name>. Note that the
+    // renaming has been done by RenameVisitor pass and is a necessary precondition. Otherwise, the
+    // behaviour is undefined.
+    std::string return_var_name = "ret_" + name;
+    if (node.is_function_block()) {
+        llvm::Value* return_var = builder.CreateAlloca(llvm::Type::getDoubleTy(*context),
+                                                       /*ArraySize=*/nullptr,
+                                                       return_var_name);
+        named_values[return_var_name] = return_var;
+    }
+
+    // Allocate parameters on the stack and add them to the symbol table.
     unsigned i = 0;
     for (auto& arg: func->args()) {
         std::string arg_name = parameters[i++].get()->get_node_name();
@@ -67,10 +78,12 @@ void CodegenLLVMVisitor::visit_procedure_or_function(const ast::Block& node) {
 
     // Add the terminator. If visiting function, the convention is that it returns a dummy 0.0
     // value.
-    if (node.is_function_block())
-        builder.CreateRet(llvm::ConstantFP::get(llvm::Type::getDoubleTy(*context), 0.0));
-    else
+    if (node.is_function_block()) {
+        llvm::Value* return_var = builder.CreateLoad(named_values[return_var_name]);
+        builder.CreateRet(return_var);
+    } else {
         builder.CreateRetVoid();
+    }
 
     values.clear();
     // \todo: Add proper support for the symbol table.
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -21,6 +21,63 @@ namespace codegen {
 
 
 /****************************************************************************************/
+/*                            Helper routines                                           */
+/****************************************************************************************/
+
+
+void CodegenLLVMVisitor::visit_procedure_or_function(const ast::Block& node) {
+    const auto& name = node.get_node_name();
+    const auto& parameters = node.get_parameters();
+
+    // Procedure or function parameters are doubles by default.
+    std::vector<llvm::Type*> arg_types;
+    for (size_t i = 0, e = parameters.size(); i < e; ++i)
+        arg_types.push_back(llvm::Type::getDoubleTy(*context));
+
+    // If visiting a function, the return type is a double by default.
+    llvm::Type* return_type = node.is_function_block() ? llvm::Type::getDoubleTy(*context)
+                                                       : llvm::Type::getVoidTy(*context);
+
+    llvm::Function* func =
+        llvm::Function::Create(llvm::FunctionType::get(return_type, arg_types, /*isVarArg=*/false),
+                               llvm::Function::ExternalLinkage,
+                               name,
+                               *module);
+
+    llvm::BasicBlock* body = llvm::BasicBlock::Create(*context, /*Name=*/"", func);
+    builder.SetInsertPoint(body);
+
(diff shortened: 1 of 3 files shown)


src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -20,6 +20,63 @@ namespace nmodl {
...
+        if (statement->is_local_list_statement() || statement->is_expression_statement())
+            statement->accept(*this);
+    }
+
+    // Add the terminator. If visiting function, the convention is that it returns a dummy 0.0
+    // value.
+    if (node.is_function_block())
+        builder.CreateRet(llvm::ConstantFP::get(llvm::Type::getDoubleTy(*context), 0.0));
+    else
+        builder.CreateRetVoid();
```
pramodk, line 73: I think we are mixing here with PROCEDURE return type? i.e. `FUNCTION` returns `double` but `PROCEDURE` returns dummy integer value `0`. e.g. if we have

```
PROCEDURE simple_proc(x, y) {
    LOCAL z
    z = x + y
}

FUNCTION simple_fun(x, y) {
    simple_fun = x + y
}
```

then generated C code is:

```
    inline int simple_proc_procedure_test(..., double x, double y) {
        int ret_simple_proc = 0;
        double z;
        z = x + y;
        return ret_simple_proc;
    }

    inline double simple_fun_procedure_test(....., double x, double y) {
        double ret_simple_fun = 0.0;
        ret_simple_fun = x + y;
        return ret_simple_fun;
    }
```

i.e. 
* procedures get return type integer and dummy return value 0
* function has return type double and they return variable `ret_<function name>` variable. Note that we create local variable `ret_<function name>` of type double in the function.

georgemitenkov: Oh, I was confused about that! That's actually what I was asking on slack the other day - what exactly `FUNCTION` returns, if there is no return statement. My confusion came from this code in C visitor:
```
// function requires return variable declaration
    if (node.is_function_block()) {
        auto type = default_float_data_type();
        printer->add_line("{} ret_{} = 0.0;"_format(type, name));
    } else {
        printer->add_line("int ret_{} = 0;"_format(name));
    }

    print_statement_block(*node.get_statement_block(), false, false);
    printer->add_line("return ret_{};"_format(name));
```
 since I didn't see anywhere that we return a variable with name = function name.

georgemitenkov: But what happens in more complicated function with control flow, etc.? Do we just have:
```
FUNCTION foo() {
   if (...) foo = 2 else foo = 3
}
```

pramodk: > Oh, I was confused about that! That's actually what I was asking on slack the other day - what exactly FUNCTION returns, if there is no return statement.

Oops.. ok! Hopefully following will explain what's going on.

> But what happens in more complicated function with control flow, etc.? 

I added following `../test/integration/mod/procedure.mod` : 

```
FUNCTION foo() {
    if (2 > 3) {
        foo = 2
    } else {
        foo = 3
    }
}
```

and 

```
./bin/nmodl ../test/integration/mod/procedure.mod
```

gives following in `procedure.cpp`:

```
    inline double foo_procedure_test(...) {
        double ret_foo = 0.0;
        if (2.0 > 3.0) {
            ret_foo = 2.0;
        } else {
            ret_foo = 3.0;
        }
        return ret_foo;
    }
```

So what we do is:
* first run `RenameVisitor` pass on FUNCTION body to change `foo` with `ret_foo`.
* we add local variable `ret_foo = 0` at the beginning of function. Type of this is depends on node type : int for procedure and double for function.
* we add `return ret_foo` statement at the end of function

pramodk: For renaming, see this https://github.com/BlueBrain/nmodl/blob/ab27595d047ef8e43519a5a6058d19f1a9bc45ef/src/codegen/codegen_c_visitor.cpp#L1674

georgemitenkov: Yes, this explains everything now! I will change the PR accordingly.

georgemitenkov:
I fixed the return value of the function. It relies on the renaming pass - I assume that it would be ran before LLVM pass all the time? (Put a comment with this precondition).

Also, I added a quick fix to the symbol table(`named_values`): using LLVM's data structure is more convenient.

pramodk:
>  It relies on the renaming pass - I assume that it would be ran before LLVM pass all the time? (Put a comment with this precondition).

Yes, this is fine. But we have to explicitly run this ourselves. We can do this before calling this LLVM codegen visitor or you can call it while visiting `visit_function_block()` as we do in https://github.com/BlueBrain/nmodl/blob/ab27595d047ef8e43519a5a6058d19f1a9bc45ef/src/codegen/codegen_c_visitor.cpp#L1661  i.e.

```cpp
    // name of return variable
    std::string name =  node.get_node_name();
    std::string return_var  = "ret_" + name;

    // first rename return variable name
    auto& block = node.get_statement_block().get();
    RenameVisitor v(name, return_var);
    block->accept(v);

    // now do codegen
    visit_procedure_or_function(node);
```

georgemitenkov:
Good! I think it will actually be more safe to run the visitor on the function (like in C visitor and your example).

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -20,6 +21,80 @@ namespace nmodl {
...
+/****************************************************************************************/
+
+
+void CodegenLLVMVisitor::visit_procedure_or_function(const ast::Block& node) {
+    const auto& name = node.get_node_name();
+    const auto& parameters = node.get_parameters();
+
+    // Procedure or function parameters are doubles by default.
+    std::vector<llvm::Type*> arg_types;
+    for (size_t i = 0, e = parameters.size(); i < e; ++i)
```
pramodk, line 35: just nitpick : should be fine to just do `i < parameters.size();` as well

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -20,6 +21,80 @@ namespace nmodl {
...
+                               llvm::Function::ExternalLinkage,
+                               name,
+                               *module);
+
+    // Create the entry basic block of the function/procedure and point the local named values table
+    // to the symbol table.
+    llvm::BasicBlock* body = llvm::BasicBlock::Create(*context, /*Name=*/"", func);
+    builder.SetInsertPoint(body);
+    local_named_values = func->getValueSymbolTable();
+
```
pramodk, line 53: nice! didn't know about this!

pramodk (approved):
this is in good shape !

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: I think that's LLVM/MLIR style to cache upper bound. But not necessary, right!

pramodk (approved):
LGTM

PR #477, opened 2020-12-29
Function call support

### LINKED ISSUE #472: Function call support
Support function calls in NMODL
- locally defined functions
- mathematical built-ins

Question: can we call function external to the current mod file?
Question: does ordering of functions matter? (do we need declarations?)

georgemitenkov:
This patch introduces support for function call code generation, particularly:
- User-defined procedures and functions are supported
- Framework for external method has been created, currently support for `exp` and `pow` has been added

Added corresponding tests.

fixes #472

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -5,8 +5,8 @@
  * Lesser General Public License. See top-level LICENSE file for details.
  *************************************************************************/
 
-#include "ast/all.hpp"
 #include "codegen/llvm/codegen_llvm_visitor.hpp"
+#include "ast/all.hpp"
 #include "codegen/codegen_helper_visitor.hpp"
 #include "visitors/rename_visitor.hpp"
 #include "visitors/visitor_utils.hpp"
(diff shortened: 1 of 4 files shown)


src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -44,7 +44,57 @@ void CodegenLLVMVisitor::run_llvm_opt_passes() {
...
+                                              const std::string& name,
+                                              const ast::ExpressionVector& arguments) {
+    // Check that function is called with the expected number of arguments.
+    if (arguments.size() != func->arg_size()) {
+        throw std::runtime_error("Error: Incorrect number of arguments passed");
+    }
+
+    // Process each argument and add it to a vector to pass to the function call instruction. Note
+    // that type checks are not needed here as NMODL operates on doubles by default. If that was not
+    // the case, cast instructions had to be created.
```
pramodk, line 84: > If that was not the case, cast instructions had to be created.

this won't be required, at DSL level we will make sure to have correct types.

pramodk (approved):
nice! LGTM!

About Jenkins CI failure: 

> default — Build finished.

I am taking a look at this on internal CI node. There is some weird linker issue with Intel compiler. I will try to fix this by tomorrow otherwise we can merge this.

pramodk:
> About Jenkins CI failure:
> default — Build finished.
> I am taking a look at this on internal CI node. There is some weird linker issue with Intel compiler. I will try to fix this by tomorrow otherwise we can merge this.

@georgemitenkov : to fix Jenkins build issue, you have to do small change:

```
$ git diff
diff --git a/test/unit/CMakeLists.txt b/test/unit/CMakeLists.txt
index d4b29bf..f0c2e95 100644
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -98,6 +98,7 @@ if(NMODL_ENABLE_LLVM)
   add_executable(testllvm visitor/main.cpp codegen/llvm.cpp)
   target_link_libraries(
     testllvm
+    llvm_codegen
     codegen
     visitor
     symtab
@@ -105,7 +106,6 @@ if(NMODL_ENABLE_LLVM)
     util
     test_util
     printer
-    llvm_codegen
     ${NMODL_WRAPPER_LIBS}
     ${LLVM_LIBS_TO_LINK})
   set(CODEGEN_TEST testllvm)
```

The reason for this is that llvm_codegen library has dependency with codegen library (for CodegenHelperVisitor) and hence that needs to be linked after llvm_codegen (as we are building all static libraries).  This error appears with Intel compiler (only) that we use under Jernkins CI.


By the way, I assume I mentioned previously how to check Jenkins CI logs. If not, here is summary : 

* Click on the `Details` link

image

* Login via Gasper account

image

* It will take you to Jenkins build. Click on the `default` link:

image

* And then `Console Output` on left side bar

image

* And if you scroll to the end of log, you should see the error:

image

pramodk (approved):
CI now looks green!

PR #478, opened 2020-12-30
Georgemitenkov/llvm indexed names

### LINKED ISSUE #467: Support indexed values
Currently, there is no support for `IndexedName` construct, e.g. `x[i] = ...`.

georgemitenkov:
This PR addresses `IndexedName`s, adding the following:
- Initialising arrays in LOCAL blocks is now supported.
- Assigning to array elements is supported.
- Using array elements in expressions is possible now.

Note that this handles 1D arrays (same as NMODL)

Also, codegen for `DEFINE` node has been implemented: macros are currently stored in a dedicated map.

fixes #467

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -183,6 +183,7 @@ void CodegenLLVMVisitor::visit_binary_expression(const ast::BinaryExpression& no
     llvm::Value* rhs = values.back();
     values.pop_back();
     if (op == ast::BinaryOp::BOP_ASSIGN) {
+        // \todo: handle indexed name.
         auto var = dynamic_cast<ast::VarName*>(node.get_lhs().get());
         if (!var) {
             throw std::runtime_error("Error: only VarName assignment is currently supported.\n");
@@ -220,6 +221,10 @@ void CodegenLLVMVisitor::visit_boolean(const ast::Boolean& node) {
     values.push_back(constant);
 }
 
+void CodegenLLVMVisitor::visit_define(const ast::Define& node) {
+    macros[node.get_node_name()] = node.get_value()->get_value();
+}
+
 void CodegenLLVMVisitor::visit_double(const ast::Double& node) {
     const auto& constant = llvm::ConstantFP::get(llvm::Type::getDoubleTy(*context),
                                                  node.get_value());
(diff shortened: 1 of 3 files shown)


src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -165,6 +210,10 @@ void CodegenLLVMVisitor::visit_boolean(const ast::Boolean& node) {
     values.push_back(constant);
 }
 
+void CodegenLLVMVisitor::visit_define(const ast::Define& node) {
+    macros[node.get_node_name()] = node.get_value()->get_value();
```
pramodk, line 214: Same question as before : can the symtab be used for this purpose?

src/codegen/llvm/codegen_llvm_visitor.hpp
```diff
@@ -96,6 +99,35 @@ class CodegenLLVMVisitor: public visitor::ConstAstVisitor {
...
+     */
+    llvm::Value* codegen_indexed_name(const ast::IndexedName& node);
+
+    /**
+     * Returns GEP instruction to 1D array
+     * \param name 1D array name
+     * \param index element index
+     * \return GEP instruction value
+     */
+    llvm::Value* create_GEP(const std::string& name, unsigned index);
```
pramodk, line 122: nitpick : stylistically, in nmodl we would write as `create_gep` (but it's true that in LLVM method has `GEP`)

pramodk (commented):
lgtm except the comment about using macro values from global symtab.

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -25,6 +25,43 @@ namespace codegen {
...
+
+    // Check if index is within array bounds.
+    if (!check_array_bounds(node, index))
+        throw std::runtime_error("Error: Index is out of bounds");
+
+    return create_GEP(node.get_node_name(), index);
+}
+
+unsigned CodegenLLVMVisitor::get_array_index_or_length(const ast::IndexedName& indexed_name) {
+    auto integer = dynamic_cast<ast::Integer*>(indexed_name.get_length().get());
```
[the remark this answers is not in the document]
georgemitenkov: Thanks for letting me know about this!

georgemitenkov:
@pramodk I used a symbol table instead! However, while testing, I noticed that in C visitor, the macro doesn't seem to be expanded?
```
DEFINE N 2

PROCEDURE foo() {
        LOCAL x[N]
}
```
gives 
```
inline int foo_(int id, int pnodecount, _Instance* inst, double* data, const Datum* indexes, ThreadDatum* thread, NrnThread* nt, double v) {
        int ret_foo = 0;
        double x[N];
        return ret_foo;
    }
```
and there is no definition of N as well.

pramodk:
Oops..that's a bug! Didn't notice that earlier as local array is not commonly used. :unamused:

Could you create an issue with above code snippet? I will then take care of this in C codegen.

georgemitenkov:
Created an issue :) It seems like only occurring when used for array sizes.

pramodk (approved):
LGTM

PR #481, opened 2021-01-04
Implement exprelr and vexpm1

cattabiani:
Add support for exprrelr and expm1 as implemented in [arbor](https://arbor.readthedocs.io/en/latest/internals/simd_api.html?highlight=exprelr#operatorname-expm1-x)

### DIFF
--- a/src/codegen/fast_math.ispc
+++ b/src/codegen/fast_math.ispc
@@ -23,7 +23,7 @@ static inline unsigned int64 dp2uint64(double x) {
     return *((varying unsigned int64*) (&x));
 }
 
-static inline float uint322sp(int32 x) {
+static inline float int322sp(int32 x) {
     return *((varying float*) (&x));
 }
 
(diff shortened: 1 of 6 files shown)


src/codegen/fast_math.ispc
```diff
@@ -78,13 +79,9 @@ static const uniform float PX4expf = 4.1665795894E-2f;
...
 
-/// double precision exp function
-static inline double vexp(double initial_x) {
-    double x = initial_x;
-    double px = dpfloor(LOG2E * x + 0.5d);
-    const int32 n = px;
+static inline double egm1(double x, double px) {
 
     x -= px * 6.93145751953125d-1;
     x -= px * 1.42860682030941723212d-6;
```
ohm314, line 87: maybe these two constants should be also named, liek they are for the single prec version
and the two lines should be taken out of this function because they are not part of the `$e^g -1$` expression. It makes more sense the way it's onde for single prec.

ohm314 (approved):
Looks good except for some minor documentation issues

PR #486, opened 2021-01-07
Fixup! Deprecation warning for sympy 1.7

### LINKED ISSUE #485: sympy 1.7 deprecation warning
Sympy 1.7 deprecated sympy.printing.ccode. We need to remove the warning

```
importing from sympy.printing.ccode has been deprecated since SymPy 1.7. 
Use Import from sympy.printing.c instead. 
See https://github.com/sympy/sympy/issues/20256 for more info.

### PR DESCRIPTION
**Pull request by cattabiani**
Fix #485

### DIFF
--- a/nmodl/ode.py
+++ b/nmodl/ode.py
@@ -12,11 +12,15 @@ import sympy as sp
 # import known_functions through low-level mechanism because the ccode
 # module is overwritten in sympy and contents of that submodule cannot be
 # accessed through regular imports
-known_functions = import_module('sympy.printing.ccode').known_functions_C99
+major, minor = (int(v) for v in sp.__version__.split(".")[:2])
+if major >= 1 and minor >= 7:
+    known_functions = import_module('sympy.printing.c').known_functions_C99
+else:
+    known_functions = import_module('sympy.printing.ccode').known_functions_C99
 known_functions.pop('Abs')
 known_functions['abs'] = 'fabs'
 
-major, minor = (int(v) for v in sp.__version__.split(".")[:2])
+
 if not ((major >= 1) and (minor >= 2)):
     raise ImportError(f"Requires SympPy version >= 1.2, found {major}.{minor}")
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: nmodl/ode.py [hunk 1]
```diff
@@ -12,7 +12,10 @@
 # import known_functions through low-level mechanism because the ccode
 # module is overwritten in sympy and contents of that submodule cannot be
 # accessed through regular imports
-known_functions = import_module('sympy.printing.ccode').known_functions_C99
+if int(sp.__version__.replace('.','')) < 17:
+    known_functions = import_module('sympy.printing.ccode').known_functions_C99
```
pramodk, line 16: versions like `1.6.1` like possible?

cattabiani: yes, I noticed when I woke up this morning. Corrected

cattabiani:
please retest

pramodk (commented):
LGTM!

one comment though : our CI install version <1.6. I assume that is still necessary for tests? (to get same output?)

https://github.com/BlueBrain/nmodl/blob/d583ae80943b576f2498d1736dc7b0712c26fa0d/azure-pipelines.yml#L41

cattabiani:
> one comment though : our CI install version <1.6. I assume that is still necessary for tests? (to get same output?)

not at all! I fixed that with this: #441

I thought I removed that limit but, evidently, I forgot

cattabiani:
please retest

ohm314 (approved):
lgtm

nmodl/ode.py
```diff
@@ -12,11 +12,15 @@
 # import known_functions through low-level mechanism because the ccode
 # module is overwritten in sympy and contents of that submodule cannot be
 # accessed through regular imports
-known_functions = import_module('sympy.printing.ccode').known_functions_C99
+major, minor = (int(v) for v in sp.__version__.split(".")[:2])
+if major >= 1 and minor >= 7:
```
alkino, line 16: This is not correct.

2.2 will not work here.

pramodk: Stupid me! thanks Nico for pointing that!

We should convert this to integer and compare against it (`major*10+minor > 17`).

alkino: Only if minor will never be > 9

Issue #55, opened 2019-03-12
derivimplicit/sparse solvers: order of statements in derivative block


lkeegan:
Derivimplicit/sparse solvers solve a coupled set of ODEs.

However, in the derivative block the ODE statements are done separately for each variable, and other statements can be inserted between these statements.

A simple example:

```
STATE {
        m
        h
}
DERIVATIVE states {
        m' =  a
        a = a + 1
        h' = a
}
```
For derivimplicit/sparse, we currently first apply all non-derivative statements in order, i.e.  `a = a+1`, then solve the ODE system, so this would be like solving
```
DERIVATIVE states {
        m' = a+1
        h' = a+1
}
```
So the first equation is incorrect

Simplest solution:

- require that derivimplicit/sparse derivative blocks have ODEs at bottom of the block with no other statements mixed between them

or:

- require that derivimplicit/sparse derivative blocks do not mix statements and ODE's, i.e. first statements, followed by the ODE's, possibly followed by more statements
- insert ODE solution between these two sets of statements in the codegen

or (complicated):

- keep track of all intermediate statements, recursively substitute them into subsequent ODE statements before solving system

PR #487, opened 2021-01-07
sympy substitute inplace instead of reordering

### LINKED ISSUE #60: Implement "blackbox" solvers as fallback if sympy analytic fails
- would be useful for testing
- calculate Jacobian numerically with difference approx
- would also work with if statements, verbatim blocks, etc
- performance not critical

cattabiani:
Necessary to enable control flow and redefinitions

Possibly fix #60, #55


cattabiani:
after various iterations I realized that this should be done in another way

PR #489, opened 2021-01-08
Extended support for binary ops and refactoring

### LINKED ISSUE #453: Support remaining binary/unary operators
Currently, codegen supports floating-point add/div/mul/sub/neg. Scale this to support logical and comparison operators.

georgemitenkov:
This PR addresses 2 issues:
-  Adds support for more binary/unary operators
Now, there are code generation functions for all comparison and logical operators. Code generation functions are now split based on the expression "type" (assignment, arithmetic, comparison, logical). Moreover, the lhs and rhs expression results can be both double and integer. This is important for control flow code generation and for the new AST node `CodegenVarType`.

Tests for new operators will be added when the first control flow node (most likely FOR node) will land.

-  Renames LLVM test file to `codegen_llvm_ir.cpp` to follow convention.

fixes #453


src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -152,6 +152,88 @@ void CodegenLLVMVisitor::emit_procedure_or_function_declaration(const ast::Block
...
+llvm::Value* CodegenLLVMVisitor::visit_comparison_bin_op(llvm::Value* lhs,
+                                                         llvm::Value* rhs,
+                                                         unsigned op) {
+    const auto& bin_op = static_cast<ast::BinaryOp>(op);
+    llvm::Value* result;
+
+    switch (bin_op) {
+#define DISPATCH(binary_op, f_llvm_op, i_llvm_op) \
+    case binary_op:                               \
+        if (lhs->getType()->isDoubleTy())         \
```
pramodk, line 217: same

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -152,6 +152,88 @@ void CodegenLLVMVisitor::emit_procedure_or_function_declaration(const ast::Block
...
+llvm::Value* CodegenLLVMVisitor::visit_arithmetic_bin_op(llvm::Value* lhs,
+                                                         llvm::Value* rhs,
+                                                         unsigned op) {
+    const auto& bin_op = static_cast<ast::BinaryOp>(op);
+    llvm::Value* result;
+
+    switch (bin_op) {
+#define DISPATCH(binary_op, llvm_fp_op, llvm_int_op) \
+    case binary_op:                                  \
+        if (lhs->getType()->isDoubleTy())            \
```
pramodk, line 164: Should here be `lhs->getType()->isDoubleTy() || rhs->getType()->isDoubleTy()` ? 🤔

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -152,6 +152,88 @@ void CodegenLLVMVisitor::emit_procedure_or_function_declaration(const ast::Block
...
+    default:
+        return nullptr;
+    }
+}
+
+void CodegenLLVMVisitor::visit_assign_op(const ast::BinaryExpression& node, llvm::Value* rhs) {
+    auto var = dynamic_cast<ast::VarName*>(node.get_lhs().get());
+    if (!var) {
+        throw std::runtime_error("Error: only VarName assignment is currently supported.\n");
+    }
```
pramodk, line 186: this is fine. I think such "legality / semantic" check can/will happen before codegen.

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: Not really, I think. LLVM instructions, particularly binary ones, operate on the same type operands. Therefore to find out what instruction should we use (integer or floating-point) can be determined by a single lhs or rhs. 

In NMODL, I assume that double and integer values never really interact with each other? Doubles are used in computations, as arguments and return values, whereas integers are used for loops/conditions? If they do interact, then we should also emit instructions like `sitofp` to cast types. But I think you have previously mentioned that should not be the case?

Alternatively, I can add a check for both types to be the same. Relying on a single lhs is not that safe, I agree.

pramodk: >  assume that double and integer values never really interact with each other? Doubles are used in computations, as arguments and return values, whereas integers are used for loops/conditions?

thats correct

> But I think you have previously mentioned that should not be the case?

yes. I was thinking that _at some point_ we might emit `1+3` as integer operation but then I realised that people do write `1/3` in DSL and there we don't want to bring integer types.

In summary, ignore my previous comment and emit floating point instruction.

One suggestion though: for floating point type, can you put some flag in the class indicating floating point type is `double` or `float`? e.g. `bool floating_type_double = true or false;`. If we toggle this, code generation should emit either `FloatType` or `DoubleType` ? (assuming its not too much work). This would be helpful for evaluations / benchmarking later.

georgemitenkov: > One suggestion though: for floating point type, can you put some flag in the class indicating floating point type is double or float? e.g. bool floating_type_double = true or false;. If we toggle this, code generation should emit either FloatType or DoubleType ? (assuming its not too much work). This would be helpful for evaluations / benchmarking later.

Sure, will add this!

pramodk (approved):
@georgemitenkov : above change you can do in separate PR if you would like.

Once you fix clang-format, this is good to merge.

> error:improper C/C++ file formatting: /jenkins/06/workspace/hpc.nmodl/platform/bb5/src/codegen/llvm/codegen_llvm_visitor.hpp

georgemitenkov:
@pramodk I made the change here straight away - it is just a couple of lines. Now we can pass an extra flag to the visitor: `--single-precision` that indicates that we need to use 32-bit types and not the default 64-bit.

I also altered one test to check that IR is generated correctly for floats.

PR #497, opened 2021-01-21
Integrating LLVM helper into LLVM visitor

georgemitenkov:
This PR integrates the helper visitor into LLVM visitor. Particularly:

- Lowering `CodegenVarType`s to LLVM types.
- Code generation for `PROCEDURE`  and `FUNCTION` blocks has been moved to a single `visit_codegen_function()` function. Since both old blocks are still in the AST, old visitors just do nothing on invocation.
- Fixing LLVMCodegenHelper return types for procedures.

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -71,6 +71,11 @@ std::shared_ptr<ast::Expression> create_expression(const std::string& code) {
     return std::make_shared<ast::WrappedExpression>(rhs->clone());
 }
 
+CodegenFunctionVector CodegenLLVMHelperVisitor::get_codegen_functions(const ast::Program& node) {
+    const_cast<ast::Program&>(node).accept(*this);
+    return codegen_functions;
+}
+
 /**
  * \brief Add code generation function for FUNCTION or PROCEDURE block
  * @param node AST node representing FUNCTION or PROCEDURE
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -105,16 +105,23 @@ CodegenFunctionVector CodegenLLVMHelperVisitor::get_codegen_functions(const ast:
  * will have minimum logic:
  *  - Add type for the function arguments
  *  - Define variables and return variable
- *  - Add return type (void for PROCEDURE and double for FUNCTION)
+ *  - Add return type (int for PROCEDURE and double for FUNCTION)
  */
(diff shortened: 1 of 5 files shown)


pramodk:
thanks @tristan0x !

Can I suggest you to split the PR ? i.e. changes for codegen_llvm_helper_visitor and codegen_llvm_visitor. The reason being, the former is we can change easily but codegen_llvm_visitor is mostly WIP and George is working on it incrementally. To avoid conflicts in ongoing work, it would be helpful to split two.

georgemitenkov:
@pramodk I don't think splitting is needed cause I made the PR😊 This is to rebase and test the for loop + start vector codegen.

pramodk:
> @pramodk I don't think splitting is needed cause I made the PR😊

I am re-reading and laughing about this now! I saw some other notification from Tristan and clicked the email link. Somehow I was on this page and also noticed @bbpbuildbot's logo which is same as @tristan0x's slack profile pic:

image

and I was like " this must be Tristan!" :grin:

Sorry Tristan for noise! 😃  :expressionless:

@georgemitenkov : you can go ahead with the refactoring ! :D

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
```diff
@@ -98,26 +103,18 @@ std::shared_ptr<ast::Expression> create_expression(const std::string& code) {
...
     auto return_var = new ast::Name(new ast::String("ret_" + function_name));
-
-    /// return type based on node type
-    ast::CodegenVarType* ret_var_type = nullptr;
-    if (node.get_node_type() == ast::AstNodeType::FUNCTION_BLOCK) {
-        ret_var_type = new ast::CodegenVarType(FLOAT_TYPE);
-    } else {
-        ret_var_type = new ast::CodegenVarType(INTEGER_TYPE);
-    }
+    auto ret_var_type = new ast::CodegenVarType(FLOAT_TYPE);
```
pramodk, line 117: Curious why you changed the type from int to float for PROCEDURE. It's true that they currently return just value 0. But their actual goal is to return error code which we are not doing now.

pramodk (commented):
overall looks ok but I am curious about the question regarding return type.

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
georgemitenkov: There was no any specific reason - I think it was done in LLVM visitor like that before - I just found it more natural since then we do not need to specifically initialise the return variable. But with error codes it all makes sense!

georgemitenkov: I will adjust the PR.

iomaganaris (approved):
LGTM

pramodk (approved):
LGTM

PR #501, opened 2021-01-25
Added support for while statement

### LINKED ISSUE #500: Add code generation for WHILE block

georgemitenkov:
This PR adds support for `WHILE` statement code generation, as well as fixes a bug in comparison expressions (mismatched FP and integer types).

fixes #500

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -578,5 +578,36 @@ void CodegenLLVMVisitor::visit_var_name(const ast::VarName& node) {
     values.push_back(var);
 }
 
+void CodegenLLVMVisitor::visit_while_statement(const ast::WhileStatement& node) {
+    // Get the current and the next blocks within the function.
+    llvm::BasicBlock* curr_block = builder.GetInsertBlock();
+    llvm::BasicBlock* next = curr_block->getNextNode();
+    llvm::Function* func = curr_block->getParent();
+
+    // Add a header and the body blocks.
+    llvm::BasicBlock* header = llvm::BasicBlock::Create(*context, /*Name=*/"", func, next);
+    llvm::BasicBlock* body = llvm::BasicBlock::Create(*context, /*Name=*/"", func, next);
+    llvm::BasicBlock* exit = llvm::BasicBlock::Create(*context, /*Name=*/"", func, next);
+
+    builder.CreateBr(header);
+    builder.SetInsertPoint(header);
+
+    // Generate code for condition and create brunch to the body block.
+    node.get_condition()->accept(*this);
+    llvm::Value* condition = values.back();
+    values.pop_back();
+    builder.CreateCondBr(condition, body, exit);
+
+    builder.SetInsertPoint(body);
+    for (const auto& statement: node.get_statement_block()->get_statements()) {
+        if (statement->is_local_list_statement() || statement->is_expression_statement() ||
+            statement->is_while_statement())
+            statement->accept(*this);
+    }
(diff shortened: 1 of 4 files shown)


test/unit/codegen/codegen_llvm_execution.cpp
```diff
@@ -127,6 +127,22 @@ SCENARIO("Optimised arithmetic expression", "[llvm][runner]") {
             }
 
             PROCEDURE foo() {}
+
+            FUNCTION loop() {
+                LOCAL i, j, sum, result
+                    result = 0
+                    WHILE (j < 2) {
```
pramodk, line 134: j is not initialised ?

test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -474,6 +474,50 @@ SCENARIO("Unary expression", "[visitor][llvm]") {
...
+SCENARIO("While", "[visitor][llvm]") {
+    GIVEN("Procedure with a simple while loop") {
+        std::string nmodl_text = R"(
+            FUNCTION loop() {
+                LOCAL i
+                i = 0
+                WHILE (i < 10) {
+                    i = i + 1
+                }
+                loop = 0
```
pramodk, line 490: Just a comment: in my mind I was expecting ‘loop = i’ so that this function can be called via jit and we check return value as 10.

pramodk (commented):
Few minor comments otherwise it’s good.

test/unit/codegen/codegen_llvm_ir.cpp
georgemitenkov: That's just IR check for correct branches/blocks. I do JIT testing in `codegen_llvm_execition.cpp`, and there return the result of the loop which I verify.

pramodk (approved):

iomaganaris (approved):
LGTM :+1:

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -279,7 +279,8 @@ void CodegenLLVMVisitor::visit_procedure_or_function(const ast::Block& node) {
     const auto& statements = block->get_statements();
     for (const auto& statement: statements) {
         // \todo: Support other statement types.
-        if (statement->is_local_list_statement() || statement->is_expression_statement())
+        if (statement->is_local_list_statement() || statement->is_expression_statement() ||
+            statement->is_while_statement())
```
iomaganaris, line 283: Just a minor suggestion that could go in a future PR if you like. I saw that the following expression is reused multiple times.
```
statement->is_local_list_statement() || statement->is_expression_statement() ||
            statement->is_while_statement()
```
Maybe it would make sense to create a new function to replace this expression and avoid rewriting it and forgetting any new type of statements while they are added? Like `is_supported(statement)`?

georgemitenkov: Indeed! I did that in previous PR (maybe that's why submitting many PRs on a single day is not a good idea😊) with if/else support. This one have not been rebased, so yes, this `is_while_statement()` function will go into `is_supported()`.

But thanks for pointing that out! I think some refactoring is needed later as some repetitive patterns started to occur.

iomaganaris: Ah yes! I also forgot that, sorry :slightly_smiling_face:
Yes, this is something that can be done step by step. No worries :+1:

pramodk:
There is conflict with llvm/codegen_llvm_visitor.cpp that needs to be resolved.

I will close and reopen to see why `BlueBrain.nmodl` CI is not started.

PR #502, opened 2021-01-25
Added error handling for values not in scope

georgemitenkov:
A small PR to add error handling when a non-scope value is looked up. Before, such a lookup would yield a `nullptr`, therefore leading to a segfault. This PR adds a `lookup` function that wraps around symbol lookup, and throws an error with a message if `nullptr` is returned.


iomaganaris (approved):
LGTM. Good addition :+1:

pramodk (approved):

PR #507, opened 2021-01-28
Create mechanism instance struct in LLVM IR

iomaganaris:
Create a stuct declaration in LLVM IR module similar to `<mech>_Instance` struct declared in `cpp` files.

An example of such struct declaration is:
`test.mod` (based on `hh.mod`)
```
NEURON {
    USEION na READ ena WRITE ina
    USEION k READ ek WRITE ik
    NONSPECIFIC_CURRENT il
    RANGE gnabar
    RANGE minf, hinf, ninf, mtau, htau, ntau
}

STATE {
    m h n
}

PARAMETER {
    gnabar = .12 (S/cm2)    <0,1e9>
}

ASSIGNED {
    v (mV)
    celsius (degC)

    minf hinf ninf
    mtau (ms) htau (ms) ntau (ms)
}
```
`test.cpp`
```
/** all mechanism instance variables */
    struct _Instance  {
        const double* __restrict__ gnabar;
        double* __restrict__ minf;
        double* __restrict__ hinf;
        double* __restrict__ ninf;
        double* __restrict__ mtau;
        double* __restrict__ htau;
        double* __restrict__ ntau;
        double* __restrict__ m;
        double* __restrict__ h;
        double* __restrict__ n;
        double* __restrict__ ena;
        double* __restrict__ ina;
        double* __restrict__ ek;
        double* __restrict__ ik;
        double* __restrict__ Dm;
        double* __restrict__ Dh;
        double* __restrict__ Dn;
        double* __restrict__ v_unused;
        const double* __restrict__ ion_ena;
        double* __restrict__ ion_ina;
        double* __restrict__ ion_dinadv;
        const double* __restrict__ ion_ek;
        double* __restrict__ ion_ik;
        double* __restrict__ ion_dikdv;
    };
 ```
 
 In LLVM IR we generate the following:
 ```
 %unknown_Instance = type { double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double* }
 ```
 The variables that are taken into account are:
 ```
 88:  * Variable added to instance struct: gnabar
88:  * Variable added to instance struct: minf
88:  * Variable added to instance struct: hinf
88:  * Variable added to instance struct: ninf
88:  * Variable added to instance struct: mtau
88:  * Variable added to instance struct: htau
88:  * Variable added to instance struct: ntau
88:  * Variable added to instance struct: m
88:  * Variable added to instance struct: h
88:  * Variable added to instance struct: n
88:  * Variable added to instance struct: ena
88:  * Variable added to instance struct: ina
88:  * Variable added to instance struct: ek
88:  * Variable added to instance struct: ik
88:  * Variable added to instance struct: Dm
88:  * Variable added to instance struct: Dh
88:  * Variable added to instance struct: Dn
88:  * Variable added to instance struct: v_unused
88:  * Variable added to instance struct: ion_ena
88:  * Variable added to instance struct: ion_ina
88:  * Variable added to instance struct: ion_dinadv
88:  * Variable added to instance struct: ion_ek
88:  * Variable added to instance struct: ion_ik
88:  * Variable added to instance struct: ion_dikd
```

These variables include any float pointers or int pointers that are indexes to other vectors.

This PR also includes:
- Refactoring `CodegeCVisitor` and moving some functions to `CodegenInfo`
- Updating `hpc-coding-conventions`
- Small fixes in breaking unit tests (Maybe this only happened to my machine? I am using LLVM 11 on Ubuntu 20.04)


georgemitenkov:
@iomaganaris A quick question about the Mech struct. The pointers point to the arrays of these variables that define current, conductivity, etc, across different points on the membrane. (By that I mean that NEURON user can define segments within the model that have their own properties). Do we know the sizes of the arrays they point to at compile time?

iomaganaris:
> @iomaganaris A quick question about the Mech struct. The pointers point to the arrays of these variables that define current, conductivity, etc, across different points on the membrane. (By that I mean that NEURON user can define segments within the model that have their own properties). Do we know the sizes of the arrays they point to at compile time?

@georgemitenkov  As far as I know the sizes of the arrays are defined during run time according to the number of  cells and segments per cell. @pramodk can correct me if I am wrong.

pramodk:
What Ioannis mentioned is correct! we will know the trip count at runtime unless we do JIT (but that's different topic for now).

src/language/nmodl.yaml
```diff
@@ -438,6 +438,18 @@
                               is base class and defines common interface for these nodes.
 
                       children:
+                        - InstanceStruct:
+                            nmodl: "INSTANCE_STRUCT "
```
pramodk, line 442: I wanted to rename this but we will do it in separate PR where will rename it in C backend as well.

pramodk (approved):

test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -367,12 +367,12 @@ SCENARIO("Function", "[visitor][llvm]") {
             std::smatch m;
 
             // Check function signature. The return type should be the default double type.
-            std::regex function_signature(R"(define double @foo\(double %x1\) \{)");
+            std::regex function_signature(R"(define double @foo\(double %x[0-9].*\) \{)");
```
iomaganaris, line 370: Before merging this, I have a question for @georgemitenkov for this change.
I got issues on my system (Ubuntu 20.04, LLVM 11) when I run the tests because the variables had different names. Instead of `x1` for example it was `x6`, etc. Do you agree with this change in the tests?
I am not familiar with how LLVM decides which number to append to the variable name, so I was wondering if this is a reasonable change to avoid test failing from system to system.

georgemitenkov: LLVM auto-index the variables if there is a collision in the symbol table. The change is fine (maybe `%x[0-9]+`?)! the reason I had `x1` was to match the assignment of the temporary with its usage:
```
%t = ...
... = ... %t, ...
```
I do not see a way of checking this unless we "hardcode" the temporary SSA names (which is pretty bad in my opinion). I guess testing manually and using regexes like `[0-9]+` should be ok. In future, we may want to integrate FileCheck (https://llvm.org/docs/CommandGuide/FileCheck.html) - they have a nice way of matching and storing a temporary for usage checks.

iomaganaris: I see. Thanks a lot for the clarification once more! :+1:

PR #510, opened 2021-01-31
Printf support in LLVM IR codegen

georgemitenkov:
This PR adds support for `printf` code generation within LLVM IR pipeline. This will be useful for easier/better testing.

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -116,6 +116,11 @@ void CodegenLLVMVisitor::run_llvm_opt_passes() {
 
 void CodegenLLVMVisitor::create_external_method_call(const std::string& name,
                                                      const ast::ExpressionVector& arguments) {
+    if (name == "printf") {
+        create_printf_call(arguments);
+        return;
+    }
+
     std::vector<llvm::Value*> argument_values;
     std::vector<llvm::Type*> argument_types;
     for (const auto& arg: arguments) {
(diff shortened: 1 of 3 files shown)


georgemitenkov:
This is a draft PR because tests need to be added. I currently tested this with generating the LLVM IR module and using `lli` to actually run the code.

I noticed that when parsing strings in NMODL, the AST has additional quotes:
```
"String": [
    {
       "name": "\"Hello world\""
    }
 ]
```
I think this is because we parse variable/function names as `String` nodes as well, and they do not have quotes. My question is whether this should be addressed at parsing stage (or forming AST), whether I should trim the string, or just leave these extra quotes like they are now.

pramodk:
> I noticed that when parsing strings in NMODL, the AST has additional quotes:
> "name": "\"Hello world\""

As you see, the AST node is of type `String`. What is stored in the String node is `"Hello world"` (i.e. double quotes included). When this is printed to JSON format, it has form `"name": "value"`. But as value here has double quotes, they get escaped sequence and hence double `double quotes`. 

You can also check this quickly by editing `src/visitors/json_visitor.cpp` and changing `visit_string` method to:

```cpp
   void JSONVisitor::visit_string(const String& node) {
       printer->push_block(node.get_node_type_name());
       if (embed_nmodl) {
           printer->add_block_property("nmodl", to_nmodl(node));
       }
       node.visit_children(*this);
       std::stringstream ss;
       ss << node.eval();
       std::cout << "---->" << node.eval() << "\n";
       printer->add_node(ss.str());
       printer->pop_block();
   }
```

for input nmodl

```
 PROCEDURE hello_world() {
     printf("Hello World")
 }
```

it will print:

```
→ ./bin/nmodl ../test/integration/mod/procedure.mod passes --json-ast
..
---->hello_world
---->printf
---->"Hello World"
```

pramodk (approved):
LGTM

pramodk:
Just FYI:

```
[100%] Built target parser-gen
Scanning dependencies of target check-clang-format
error:improper C/C++ file formatting: /jenkins/07/workspace/hpc.nmodl/platform/bb5/src/codegen/llvm/codegen_llvm_visitor.hpp
make[3]: *** [cmake/hpc-coding-conventions/cpp/CMakeFiles/check-clang-format] Error 
```

pramodk (approved):

PR #514, opened 2021-02-15
defuse with pointers to the statement in a variable is used/defined

### LINKED ISSUE #513: defuse chain visitor returns also a pointer to the node or the statement
with the defuse visitor we currently return define-use chains where we state how many times a certain variable is used. It could be useful to add also a pointer that indicates in which node/statement a variable is used

[ud-chain wiki](https://en.wikipedia.org/wiki/Use-define_chain#:~:text=From%20Wikipedia%2C%20the%20free%20encyclopedia,some%20value%20to%20a%20variable.)

cattabiani:
In this PR we expand DUInstance to also return the ... "line" in which defuse is used/defined. In particular, we return a pointer to the outermost `binary_expression`. Why not returning an `expression_statement` pointer? In this way we handle these 2 cases:

- `IF (tau == 0)` \\ not expression statement, binary expression
- `x = x + 1`  \\ binary expression and expression statement

This is propedeutic for using the defuse visitor inside #495 .

Addresses #513 

PS: I also included a few corner cases in the unit tests for defuse that were not checked

### DIFF
--- a/test/unit/visitor/defuse_analyze.cpp
+++ b/test/unit/visitor/defuse_analyze.cpp
@@ -73,10 +73,9 @@ SCENARIO("Perform DefUse analysis on NMODL constructs") {
             REQUIRE(chains[0].to_string(true) == expected_text);
             REQUIRE(chains[0].eval() == DUState::D);
 
-            REQUIRE(to_nmodl(*chains[0].chain[0].expression_statement) == "tau = 1");
-            REQUIRE(to_nmodl(*chains[0].chain[1].expression_statement) == "tau = 1+tau");
-            REQUIRE(to_nmodl(*chains[0].chain[2].expression_statement) == "tau = 1+tau");
-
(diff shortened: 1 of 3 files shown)


test/unit/visitor/defuse_analyze.cpp
```diff
@@ -7,7 +7,7 @@
 
 #include <catch/catch.hpp>
 
-#include "ast/program.hpp"
+#include "ast/all.hpp"
```
pramodk, line 10: nitpik : if we have "few" ast classes to use then we include individual headers and avoid `all.hpp` (which is very large).

pramodk (approved):
I was initially wondering about the same:

>  In particular, we return a pointer to the outermost binary_expression. Why not returning an expression_statement pointer?

but by looking at test examples and use cases, it seems make sense.

Overall, looks nicely done with minimal changes / extensions! 👌

PR #517, opened 2021-02-18
BB5 CI: build and test only intel compiler / bump cmake-format to 0.6.13

tristan0x:
This change modifies GitLab CI/CD pipeline:

* drop gcc build in favor on intel
* bump cmake-format from 0.6.0 to 0.6.10


tristan0x:
The build fails on Jenkins because cmake-format is still tested with 0.6.0.
The Jenkins project should be disabled anyway @pramodk

ci/bb5-pr
```diff
@@ -16,7 +16,7 @@ function setup_virtualenv() {
     virtualenv venv
     . venv/bin/activate
     pip3 install "breathe<=4.12.0"
-    pip3 install "cmake-format==0.6.0"
+    pip3 install "cmake-format==0.6.10"
```
pramodk, line 19: as discussed on slack, we can bump this to 0.6.13

ci/bb5-pr
```diff
@@ -42,7 +42,7 @@ function build_with() {
              -DNMODL_FORMATTING:BOOL=ON \
              -DClangFormat_EXECUTABLE=$clang_format_exe \
              -DLLVM_DIR=/gpfs/bbp.cscs.ch/apps/hpc/jenkins/merge/deploy/externals/latest/linux-rhel7-x86_64/gcc-9.3.0/llvm-11.0.0-kzl4o5/lib/cmake/llvm
-    make -j
+    make -j2
```
pramodk, line 45: NMODL is quite slow to build. For BB5, we should be able to use 4 or 6 by default?

tristan0x: Bumped to 6

pramodk (approved):

PR #518, opened 2021-02-22
sympy var clashes

### LINKED ISSUE #515: `sympy_visitor` clash of variables of type `tmp*` and `old_*` with the user-defined ones
`sympy_visitor` creates some variables of type `tmp*` and `old_*`. Clashes with user-defined variables are not checked. This could lead to bugs (probably compilation errors)

It was already marked as `TODO` in the code. Now we also have an issue

cattabiani:
- `old_*` is checked with `suffix_random_string`
- `suffix_random_string` now also checks if the `new_string` is a prefix of another variable (already in `vars`)
- removed `_make_unique_prefix` as it was duplicated code
   - removed its tests
- added tests for name-clash prevention methods

Fix #515

Ready for review as soon as CI passes :D

### DIFF
--- a/test/unit/visitor/sympy_solver.cpp
+++ b/test/unit/visitor/sympy_solver.cpp
@@ -65,6 +65,22 @@ std::vector<std::string> run_sympy_solver_visitor(
     return results;
 }
 
+// check if in a list of vars (like LOCAL) there are duplicates
+bool is_unique_vars(std::string result) {
+    result.erase(std::remove(result.begin(), result.end(), ','), result.end());
+    std::stringstream ss(result);
+    std::string token;
+
+    std::unordered_set<std::string> old_vars;
+
+    while (getline(ss, token, ' ')) {
+        if (!old_vars.insert(token).second) {
+            return false;
+        }
+    }
+    return true;
+}
+
 
 /**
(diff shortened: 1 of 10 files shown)


cattabiani:
please retest

cattabiani:
please retest

iomaganaris (approved):
LGTM

test/unit/visitor/sympy_solver.cpp
```diff
@@ -65,6 +65,22 @@ std::vector<std::string> run_sympy_solver_visitor(
     return results;
 }
 
+// check if in a list of vars (like LOCAL) there are duplicates
+bool is_unique_vars(std::string result) {
```
pramodk, line 69: I will need bit some more info on this tomorrow - I wonder why there will be name clash.

cattabiani: if we do not replace correctly in `sympy_solver`, using the correct routine, we can get name clashes. For example if I have:
```
DERIVATIVE {
LOCAL old_x
x' = ...
y' = ...
}
```

if I forgot something it could be that 2 `old_x` appear in local

pramodk (approved):
LGTM - I think I was thinking too much about (non-existent?) corner cases. But if something I see, will report it and fix it later.

PR #526, opened 2021-02-27
Move code gen specific InstanceStruct node to codegen.yaml

pramodk:
- nmodl.yaml file is more for language constructs
  - InstanceStruct is specific for code generation and hence
    move it to codegen.yaml


iomaganaris (approved):
LGTM

PR #531, opened 2021-02-28
Generate compute kernels with correct instance variable 

pramodk:
* Improvements to codegen helper
     - instance structure now contains all global variables
     - instance structure now contains index variables for ions
     - nrn_state kernel now has all variables converted to instance
     - InstanceVarHelper added to query variable and it's location
* nmodl_to_json helper added in main.cpp
* added --vector-width CLI option
* Add argument to nrn_state_hh
*  [x] WIP progress :: add comments as TODOs to support LLVM IR generation
*  [x] : fix cmake-format
*  [x] : fix clang-format
*  [x] : fix tests

Using : 

```
make -j && ./bin/nmodl hh.mod llvm --ir --vector-width 8 passes --json-ast
```

We get now

```
INSTANCE_STRUCT {
    DOUBLE *gnabar
    DOUBLE *gkbar
    DOUBLE *gl
    DOUBLE *el
    DOUBLE *gna
    DOUBLE *gk
    DOUBLE *il
    DOUBLE *minf
    DOUBLE *hinf
    DOUBLE *ninf
    DOUBLE *mtau
    DOUBLE *htau
    DOUBLE *ntau
    DOUBLE *m
    DOUBLE *h
    DOUBLE *n
    DOUBLE *Dm
    DOUBLE *Dh
    DOUBLE *Dn
    DOUBLE *ena
    DOUBLE *ek
    DOUBLE *ina
    DOUBLE *ik
    DOUBLE *v_unused
    DOUBLE *g_unused
    DOUBLE *ion_ena
    DOUBLE *ion_ina
    DOUBLE *ion_dinadv
    DOUBLE *ion_ek
    DOUBLE *ion_ik
    DOUBLE *ion_dikdv
    INTEGER *ion_ena_index
    INTEGER *ion_ina_index
    INTEGER *ion_dinadv_index
    INTEGER *ion_ek_index
    INTEGER *ion_ik_index
    INTEGER *ion_dikdv_index
    DOUBLE *voltage
    INTEGER *node_index
    DOUBLE t
    DOUBLE dt
    DOUBLE celsius
    INTEGER secondorder
    INTEGER node_count
}

VOID nrn_state_hh(INSTANCE_STRUCT *mech){
    INTEGER id
    for(id = 0; id<mech->node_count; id = id+8) {
        INTEGER node_id, ena_id, ek_id
        DOUBLE v
        node_id = mech->node_index[id]
        ena_id = mech->ion_ena_index[id]
[...]

PR #533, opened 2021-03-02
Addressing TODOs for Instance struct

georgemitenkov:
This PR addresses certain TODOs from [2da2bc4](https://github.com/BlueBrain/nmodl/commit/2da2bc4e5829091c9b28f356731646db5209a9a7) and would help with instance struct integration (see main [PR](https://github.com/BlueBrain/nmodl/pull/531))

Scalar integration:

- [x] Add instance type for `CodegenVarType` visitor
- [x] Emit instance type prior to function declarations (NOT NEEDED)
- [x] Add instance members query via helper struct for the kernel `nrn_state` generation
- [x] Clean-up member creation

Simple direct index vectorization:
- [x] Make the kernel for loop vectorized

Other:
~- [ ] Update tests to reflect the changes and test the kernel~ Will be done in separate PR

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -159,19 +159,44 @@ void CodegenLLVMHelperVisitor::create_function_for_node(ast::Block& node) {
 }
 
 std::shared_ptr<ast::InstanceStruct> CodegenLLVMHelperVisitor::create_instance_struct() {
-    ast::CodegenVarVector codegen_vars;
+    ast::CodegenVarWithTypeVector codegen_vars;
+
+    auto add_var_with_type = [&](const std::string& name, const ast::AstNodeType type, int is_pointer, size_t index) {
+        auto var_name = new ast::Name(new ast::String(name));
+        auto var_type = new ast::CodegenVarType(type);
+        auto var_index = new ast::Integer(index, nullptr);
+        auto codegen_var = new ast::CodegenVarWithType(var_type, is_pointer, var_name, var_index);
+        codegen_vars.emplace_back(codegen_var);
+    };
+
+    // the order or index in the structure
+    size_t index = 0;
+
     /// float variables are standard pointers to float vectors
     for (auto& float_var: info.codegen_float_variables) {
-        auto name = new ast::Name(new ast::String(float_var->get_name()));
-        auto codegen_var = new ast::CodegenVar(1, name);
-        codegen_vars.emplace_back(codegen_var);
+        add_var_with_type( float_var->get_name(), FLOAT_TYPE, 1, index++);
     }
+
     /// int variables are pointers to indexes for other vectors
     for (auto& int_var: info.codegen_int_variables) {
-        auto name = new ast::Name(new ast::String(int_var.symbol->get_name()));
-        auto codegen_var = new ast::CodegenVar(1, name);
-        codegen_vars.emplace_back(codegen_var);
+        add_var_with_type(int_var.symbol->get_name(), FLOAT_TYPE, 1, index++);
     }
+
+    // for integer variables, there should be index
+    for (auto& int_var: info.codegen_int_variables) {
+        std::string var_name = int_var.symbol->get_name() + "_index";
+        add_var_with_type(var_name, INTEGER_TYPE, 1, index++);
+    }
+
+    // add voltage
+    add_var_with_type("voltage", FLOAT_TYPE, 1, index++);
+
+    // add dt, t, celsius
+    add_var_with_type(naming::NTHREAD_T_VARIABLE, FLOAT_TYPE, 0, index++);
+    add_var_with_type(naming::NTHREAD_DT_VARIABLE, FLOAT_TYPE, 0, index++);
+    add_var_with_type(naming::CELSIUS_VARIABLE, FLOAT_TYPE, 0, index++);
+    add_var_with_type(naming::SECOND_ORDER_VARIABLE, INTEGER_TYPE, 0, index++);
+
     return std::make_shared<ast::InstanceStruct>(codegen_vars);
 }
 
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -475,6 +475,26 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
     const auto& condition = create_expression("id < node_count");
     const auto& increment = create_statement_as_expression("id = id + {}"_format(vector_width));
 
+    // Now, change the vector width and 0 types in initialization and increment blocks to Integer,
+    // since NMODL considers numbers to be Double by default. \todo: This is NOT a proper way to do
+    // this, but works for now.
+    const auto& init_expression =
+        std::dynamic_pointer_cast<ast::WrappedExpression>(initialization)->get_expression();
+    const auto& init_bin_expression = std::dynamic_pointer_cast<ast::BinaryExpression>(
+        init_expression);
+    const auto& zero = std::make_shared<ast::Integer>(/*value=*/0, /*macro=*/nullptr);
+    init_bin_expression->set_rhs(zero);
+
+    // \todo: Change this as well.
+    const auto& inc_expression =
+        std::dynamic_pointer_cast<ast::BinaryExpression>(
+            std::dynamic_pointer_cast<ast::WrappedExpression>(increment)->get_expression())
+            ->get_rhs();
+    const auto& inc_bin_expression = std::dynamic_pointer_cast<ast::BinaryExpression>(
+        std::dynamic_pointer_cast<ast::WrappedExpression>(inc_expression)->get_expression());
+    const auto& width = std::make_shared<ast::Integer>(/*value=*/vector_width, /*macro=*/nullptr);
+    inc_bin_expression->set_rhs(width);
+
     /// loop body : initialization + solve blocks
     ast::StatementVector loop_def_statements;
     ast::StatementVector loop_index_statements;
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -471,9 +471,10 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
     /// create now main compute part : for loop over channel instances
 
     /// loop constructs : initialization, condition and increment
-    const auto& initialization = create_statement_as_expression("id = 0");
-    const auto& condition = create_expression("id < node_count");
-    const auto& increment = create_statement_as_expression("id = id + {}"_format(vector_width));
+    const auto& initialization = create_statement_as_expression("{} = 0"_format(INDUCTION_VAR));
+    const auto& condition = create_expression("{} < node_count"_format(INDUCTION_VAR));
+    const auto& increment = create_statement_as_expression(
+        "{} = {} + {}"_format(INDUCTION_VAR, INDUCTION_VAR, vector_width));
 
     // Now, change the vector width and 0 types in initialization and increment blocks to Integer,
     // since NMODL considers numbers to be Double by default. \todo: This is NOT a proper way to do
@@ -504,7 +505,8 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
         std::vector<std::string> double_variables{"v"};
 
         /// access node index and corresponding voltage
-        loop_index_statements.push_back(visitor::create_statement("node_id = node_index[id]"));
+        loop_index_statements.push_back(
+            visitor::create_statement("node_id = node_index[{}]"_format(INDUCTION_VAR)));
         loop_body_statements.push_back(visitor::create_statement("v = voltage[node_id]"));
 
         /// read ion variables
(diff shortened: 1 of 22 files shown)


pramodk:
@georgemitenkov : I push the change to fix the gitlab CI. But it seems like there are some minor conflicts because earlier today I push forced few things to my PR. Conflicts are minor and you can fix it.

As we have dependent PRs :  I think we can merge this your #533 PR into my  #531 and you continue working in my branch   pramodk/llvm-instance-type. As I can't merge my PR into llvm branch without your PR, this will avoid juggling changes and conflicts.

georgemitenkov:
> @georgemitenkov : I push the change to fix the gitlab CI. But it seems like there are some minor conflicts because earlier today I push forced few things to my PR. Conflicts are minor and you can fix it.

Great! I will make the changes.

> As we have dependent PRs : I think we can merge this your #533 PR into my #531 and you continue working in my branch pramodk/llvm-instance-type. As I can't merge my PR into llvm branch without your PR, this will avoid juggling changes and conflicts.

Sure, I am nearly done with simple vector code generation, so I will make this PR ready by tonight.

pramodk:
> Sure, I am nearly done with simple vector code generation, so I will make this PR ready by tonight.

Ok great! feel free to change my or @alkino's PRs to prepare one final PR ready for merge. Then I will do final check and we can merge it.

In the meantime I will start sketching something for testing aspect.

georgemitenkov:
@pramodk Regarding testing, I had one idea:

1) Generate `llvm::Module` ysing the pipeline
2) Add a new file `test_llvm_kernels.cpp` or something like that. In that file, we create Instance struct artificially, and write cpp wrappers to print contents before/after the kernel execution. Then link the wrapper `llvm::Module` with our `llvm::Module` (For my GSoC I was using a similar strategy actually, so I have an idea of how this is done with LLVM API). Then we simply feed this into `llvm_nmodl_runner` and see what are the outputs :)

georgemitenkov:
For example something like this:

```
#include <stdio.h>

// ================= LLVM kernel generated from the pipeline ======================== //

struct Bar {
  int* __restrict__ indices;
  double* __restrict__ voltage;
  int num_nodes;
};

void kernel(Bar* b) {
  double v = -1.0;
  b->voltage[b->indices[0]] = v * b->voltage[b->indices[0]];
  b->voltage[b->indices[1]] = v * b->voltage[b->indices[1]];
}

// ================= Helpers that would come from wrapper class ==================== //

void print_struct(Bar *b) {
  printf("num nodes: %d\n", b->num_nodes);
  printf("indices: ");
  for (int i = 0; i < b->num_nodes; ++i) {
    printf("%d", b->indices[i]);
    if (i < b->num_nodes - 1) printf(", "); else printf("\n");
  }
  printf("voltage: ");
  for (int i = 0; i < b->num_nodes; ++i) {
    printf("%.2f", b->voltage[i]);
    if (i < b->num_nodes - 1) printf(", "); else printf("\n");
  }
}

int main() {
  Bar b;
  b.num_nodes = 2;
  int indices[] = {0, 1};
  double voltage[] = {5.0, 10.0};
  b.indices = indices;
  b.voltage = voltage;
  printf(" == Before == ");
  print_struct(&b);
  kernel(&b);
  printf(" == After == ");
  print_struct(&b);
  return 0;
}
```

I am currently using this to verify the vectorised code.

georgemitenkov:
@pramodk @iomaganaris This PR is now ready! Now, we fully support scalar code kernel generation + compute (i.e. non-scatter/gather/control flow) vectorised code generation.

There are no tests at the moment, but with the issue just opened I think these can be added later. At the moment, I have tested by comparing to LLVM generated from C++ and its manually defined vector version. See example below :)

georgemitenkov:
Scalar code is supported fully (running `./bin/nmodl hh.mod llvm --ir`). Vector code generation is supported for the following kernel (For now, I just commented out indirect instructions):
```
    INTEGER id
    for(id = 0; id<mech->node_count; id = id+4) {
        INTEGER node_id
        DOUBLE v
        mech->m[id] = mech->m[id]+(1.0-exp(mech->dt*((((-1.0)))/mech->mtau[id])))*(-(((mech->minf[id]))/mech->mtau[id])/((((-1.0)))/mech->mtau[id])-mech->m[id])
        mech->h[id] = mech->h[id]+(1.0-exp(mech->dt*((((-1.0)))/mech->htau[id])))*(-(((mech->hinf[id]))/mech->htau[id])/((((-1.0)))/mech->htau[id])-mech->h[id])
        mech->n[id] = mech->n[id]+(1.0-exp(mech->dt*((((-1.0)))/mech->ntau[id])))*(-(((mech->ninf[id]))/mech->ntau[id])/((((-1.0)))/mech->ntau[id])-mech->n[id])
    }
}
```

The output:
```
define void @nrn_state_hh(%hh__instance_var__type* %mech1) {
  %mech = alloca %hh__instance_var__type*, align 8
  store %hh__instance_var__type* %mech1, %hh__instance_var__type** %mech, align 8
  %id = alloca i32, align 4

  // These 2 lines are not really needed atm.
  %__vec_id = alloca <4 x i32>, align 16    
  store <4 x i32> <i32 0, i32 1, i32 2, i32 3>, <4 x i32>* %__vec_id, align 16

  br label %for.cond

for.cond:                                         ; preds = %for.inc, %0
  %1 = load %hh__instance_var__type*, %hh__instance_var__type** %mech, align 8
  %2 = getelementptr inbounds %hh__instance_var__type, %hh__instance_var__type* %1, i32 0, i32 43
  %3 = load i32, i32* %2, align 4
  %4 = load i32, i32* %id, align 4
  %5 = icmp slt i32 %4, %3
  br i1 %5, label %for.body, label %for.exit

for.body:                                         ; preds = %for.cond
  %node_id = alloca i32, align 4
  %v = alloca double, align 8

  // Getting the pointer to double* m 
  %6 = load %hh__instance_var__type*, %hh__instance_var__type** %mech, align 8
  %7 = getelementptr inbounds %hh__instance_var__type, %hh__instance_var__type* %6, i32 0, i32 13
  %8 = load i32, i32* %id, align 4
  %9 = sext i32 %8 to i64
  %10 = load double*, double** %7, align 8

  // A simple bitcast is sufficient since the data is consecutive!
  %11 = bitcast double* %10 to <4 x double>*

  // Load the actual values to the vector, then we proceed with other computations.
  %12 = getelementptr inbounds <4 x double>, <4 x double>* %11, i64 %9
  %13 = load <4 x double>, <4 x double>* %12, align 32

  // Similar instructions for all ...[id], etc.
...
```

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
```diff
@@ -471,9 +471,30 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
...
+
+    // \todo: Change this as well.
+    const auto& inc_expression =
+        std::dynamic_pointer_cast<ast::BinaryExpression>(
+            std::dynamic_pointer_cast<ast::WrappedExpression>(increment)->get_expression())
+            ->get_rhs();
+    const auto& inc_bin_expression = std::dynamic_pointer_cast<ast::BinaryExpression>(
+        std::dynamic_pointer_cast<ast::WrappedExpression>(inc_expression)->get_expression());
+    const auto& width = std::make_shared<ast::Integer>(/*value=*/vector_width, /*macro=*/nullptr);
+    inc_bin_expression->set_rhs(width);
```
pramodk, line 497: Ok. Will take care separately.

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -373,6 +518,117 @@ void CodegenLLVMVisitor::visit_boolean(const ast::Boolean& node) {
...
+//      |     +------------------------+  |
+//      |     | <increment code>       |  |
+//      |      | br %cond              |  |
+//      |     +------------------------+  |
+//      |                 |               |
+//      |                 +---------------+
+//      V
+//  +---------------------------+
+//  | <code after for loop>     |
+//  +---------------------------+
```
pramodk, line 552: 👌 - I should also start using such documentation.

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -406,7 +662,7 @@ void CodegenLLVMVisitor::visit_codegen_function(const ast::CodegenFunction& node
     block->accept(*this);
 
     // If function has a void return type, add a terminator not handled by CodegenReturnVar.
-    if (node.is_void())
+    if (node.get_return_type()->get_type() == ast::AstNodeType::VOID)
```
pramodk, line 665: Ah yes! I might have missed this in code review as well - return type is of type `CodegenVarType`.

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -456,6 +712,10 @@ void CodegenLLVMVisitor::visit_codegen_var_list_statement(
 }
 
 void CodegenLLVMVisitor::visit_double(const ast::Double& node) {
+    if (is_kernel_code && vector_width > 1) {
```
pramodk, line 715: got bit confused here : In my mind `compute loop body` as well as outside loop code is part of `kernel`. May be we should rename this to `is_loop_body_code` or something similar?

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -56,10 +64,81 @@ llvm::Value* CodegenLLVMVisitor::codegen_indexed_name(const ast::IndexedName& no
...
+    auto member_indexed_name = std::dynamic_pointer_cast<ast::IndexedName>(
+        member_var_name->get_name());
+    if (!member_indexed_name->get_length()->is_name())
+        throw std::runtime_error("Error: " + member_name + " has a non-Name index!");
+
+    // Load the index variable that will be used to access the member's element. Since we index a
+    // pointer variable, we need to extend the 32-bit integer index variable to 64-bit.
+    llvm::Value* i32_index = builder.CreateLoad(
+        lookup(member_indexed_name->get_length()->get_node_name()));
+    llvm::Value* i64_index = builder.CreateSExt(i32_index, llvm::Type::getInt64Ty(*context));
```
pramodk, line 111: just curious - if index variable will be `size_t` type instead of integer, would that help?

pramodk (approved):
LGTM - I added some minor comments.

pramodk:
CI failing because of:

>  error:improper C/C++ file formatting: /gpfs/bbp.cscs.ch/home/bbprelman/gl/P994/hpc/nmodl/src/codegen/llvm/codegen_llvm_visitor.hpp

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: This is done to handle the case when we visit a function that is not a kernel/not vector

```
// Not vectorised
PROCEDURE() {
  some constant = 2
}

// Vectorised
VOID nrn_state(...) { .. }
```

georgemitenkov: What we need to think about, is that if we have a function call (e.g.`rates(voltage)` in original HH), how we proceed in this case?
- Inline function?
- Vectorise function (`rates` here?)
etc.

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: Do you mean update the type of the value in the node? In that case that does not make a difference. The sign extend is on IR level, independently of what is contained in the AST node. 

Why we cast to `i64` is an interesting question. From what I know, index bit width usually comes from the ABI information, data layout, etc. On CPU platforms this is usually 64 bit. So it is fine to use both 32 or 64 bit indices I think - I just reimplemented Clang's convention for indices, but `sext` can be dropped.

georgemitenkov: Quoting LLVM Ref

Why index struct with `i32`?
> The specific type i32 is probably just a historical artifact, however it’s wide enough for all practical purposes, so there’s been no need to change it. It doesn’t necessarily imply i32 address arithmetic; it’s just an identifier which identifies a field in a struct. Requiring that all struct indices be the same reduces the range of possibilities for cases where two GEPs are effectively the same but have distinct operand types.

Otherwise we index with `i64`. Actually this just reminded me that our indices for arrays are `i32` that is not really conventional.

pramodk:
> Other
> Update tests to reflect the changes and test the kerne

@georgemitenkov : this is ready to merge in my PR right? Just asking because there is one todo in the description .

georgemitenkov:
@pramodk It is ready for the merge. I have compared the LLVM output manually, and it looks correct. Since we do not know how to test it properly yet, the tests can be added later separately.

src/codegen/llvm/codegen_llvm_visitor.cpp
pramodk: partially that's because for the reason of memory saving (i32 is sufficient what we store).

pramodk: > Do you mean update the type of the value in the node? In that case that does not make a difference. The sign extend is on IR level, independently of what is contained in the AST node.
Yes. Ok.

src/codegen/llvm/codegen_llvm_visitor.cpp
pramodk: This is good point.

> Inline function?

We already do this at NMODL level : if you pass the flag `passes --inline` then all procedures call are inlined inside AST.

> Vectorise function (rates here?)

that's another option but for all current use cases, DSL level inlining will be sufficient. But I think this deserve a ticket!

Issue #532, opened 2021-03-01
Create epilog/remainder loop for the `FOR` node in the AST 


pramodk:
Assuming llvm branch setup is ready with #524: 

In the hh.mod (see #524), we have DERIVATIVE block:

```python
DERIVATIVE states {
     m' =  (minf-m)/mtau
     h' = (hinf-h)/htau
     n' = (ninf-n)/ntau
}
```

which gets translated into:

```cpp
VOID nrn_state_hh(INSTANCE_STRUCT *mech){
    INTEGER id
    for(id = 0; id<mech->node_count; id = id+1) {
        INTEGER node_id, ena_id, ek_id
        DOUBLE v
        node_id = mech->node_index[id]
        ena_id = mech->ion_ena_index[id]
        ek_id = mech->ion_ek_index[id]
        v = mech->voltage[node_id]
        mech->ena[id] = mech->ion_ena[ena_id]
        mech->ek[id] = mech->ion_ek[ek_id]
        mech->m[id] = mech->m[id]+(1.0-exp(mech->dt*((((-1.0)))/mech->mtau[id])))*(-(((mech->minf[id]))/mech->mtau[id])/((((-1.0)))/mech->mtau[id])-mech->m[id])
        mech->h[id] = mech->h[id]+(1.0-exp(mech->dt*((((-1.0)))/mech->htau[id])))*(-(((mech->hinf[id]))/mech->htau[id])/((((-1.0)))/mech->htau[id])-mech->h[id])
        mech->n[id] = mech->n[id]+(1.0-exp(mech->dt*((((-1.0)))/mech->ntau[id])))*(-(((mech->ninf[id]))/mech->ntau[id])/((((-1.0)))/mech->ntau[id])-mech->n[id])
    }
}
```

We are now going to vectorise the `for` loop with certain vector width e.g. say `8`. With this, we will have main vector loop with trip count of `node_count / 8` and [epilog/remainder](https://software.intel.com/content/www/us/en/develop/articles/what-are-peel-and-remainder-loops-fortran-vectorization-support.html) loop  of trip count  `node_count % 8`.

So as part of this ticket we should do following:

* `CodegenLLVMHelperVisitor` should be changed so that `FOR` loop is now duplicated for `main vector loop` and `remainder` serial loop
* vector loop is incremented by vector width parameter provided through CLI
* remainder loop is serial and hence incremented by 1
* note that each act node has `clone` method that can be used for duplication
[...]

PR #534, opened 2021-03-03
Add the remainder loop for vectorization of DERIVATIVE block

alkino:


src/codegen/llvm/codegen_llvm_helper_visitor.cpp
```diff
@@ -534,17 +529,43 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
...
+                                                                                  loop_block);
+
+        /// convert all variables inside loop body to instance variables
+        convert_to_instance_variable(*for_loop_statement_main, loop_index_var);
+
+        /// loop itself becomes one of the statement in the function
+        function_statements.push_back(for_loop_statement_main);
+    }
+
+    /// remainder loop possibly vectorized on vector_width
```
pramodk, line 551:
```suggestion
    /// remainder loop (serial)
```

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
```diff
@@ -534,17 +529,43 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
...
+        const auto& initialization = create_statement_as_expression("");
+        const auto& condition = create_expression("id < node_count");
+        const auto& increment = create_statement_as_expression("++id");
 
-    /// convert all variables inside loop body to instance variables
-    convert_to_instance_variable(*for_loop_statement, loop_index_var);
+        auto for_loop_statement_remainder = std::make_shared<ast::CodegenForStatement>(initialization,
+                                                                                  condition,
+                                                                                  increment,
+                                                                                  loop_block);
```
pramodk, line 561: `loop_block` is a pointer and hence we need to clone the loop_block (to avoid side effects - if we transform first loop then 2nd will be also modified)

pramodk (changes requested):

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
alkino: Don't we want that both loop are exactly the same?

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
alkino: what does it mean? remainder for serial?

pramodk: Here serial means non-SIMD : Let's say vector width is 8 and loop trip count is 10. The remainder loop has trip count is 2 and it's executed in serial fashion / non-simd.

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
pramodk: No. Vector loop can have extra transformations:

```
# original loop
for(....) {
  int x
  ...
}

# will become

for (...) {
  int x<8>  // width of 8
  ...
}

for (....) { // reminder 
  int x       => this shouldn't be changed!
  ...
}
```

test/unit/codegen/codegen_helper.cpp
```diff
@@ -52,6 +55,20 @@ std::string run_inline_visitor(const std::string& text) {
     return variables;
 }
 
+std::vector<std::shared_ptr<ast::Ast>> run_inline_visitor_helper(const std::string& text) {
```
pramodk, line 58:
```suggestion
std::vector<std::shared_ptr<ast::Ast>> run_codegen_helper_visitor(const std::string& text) {
```

test/unit/codegen/codegen_helper.cpp
```diff
@@ -168,3 +185,33 @@ SCENARIO("unusual / failing mod files", "[codegen][var_order]") {
...
+            }
+            BREAKPOINT {
+                SOLVE states METHOD cnexp
+            }
+            DERIVATIVE states {
+                m = (minf-m)/mtau
+            }
+        )";
+
+        THEN("CHECK IT") {
```
pramodk, line 212:
```suggestion
        THEN("check if main loop and epilog loop is generated") {
```

test/unit/codegen/codegen_helper.cpp
```diff
@@ -168,3 +185,33 @@ SCENARIO("unusual / failing mod files", "[codegen][var_order]") {
...
+            auto result = run_inline_visitor_helper(nmodl_text);
+            REQUIRE(result.size() == 2);
```
pramodk, line 214: we should check the contents of individual node as well. For example, something like:
```
auto main_loop = to_nmodl(result[0]);
// here check main_loop against expected
// same for remainder loop
```

alkino:
Fix #532

pramodk:
@alkino : 1) As `pramodk/llvm-instance-type` is already merged to llvm, you should create PR against llvm branch 2) could you see why GitLab CI failing?

pramodk (approved):
LGTM

PR #535, opened 2021-03-04
Adjust GitLab CI pipeline

olupton:
Fix GitLab CI pipeline failures:
- Don't use `cache:` to pass data around; this is not guaranteed to work and can bring in data from old jobs.
- Use `bb5_build_dir: pipeline` so all jobs run in the same working directory.
- Disable git clone for all but the first job

This is intended to be a ~minimal change to get the pipeline running again.

cc: @pramodk

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,38 +1,33 @@
-cache: &global_cache
-  key: ${CI_COMMIT_REF_SLUG}
-  paths:
-  - venv/
-  - build_intel/
+.common:
+  tags: [bb5]
+  script: ci/bb5-pr
+  only:
+    - external_pull_requests
+  variables:
+    # Just run everything in the same, persistent, directory.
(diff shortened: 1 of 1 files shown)


tristan0x (approved):

pramodk:
LGTM but can we run clang & cmake format in parallel to build? This way, I know the build as well as (stupid) clang/cmake formatting issues in one go and avoid one more PR?

image

olupton:
It would be nice to have that, but it would take a little more reorganisation.

The formatting check from `hpc-coding-conventions` has to run after the CMake configure step, which in the current CI is part of the build job. The other concern is that everything is operating on the same working directory, so if the formatting checks are not strictly read-only they could interfere with the build.

I suggest we merge this first and see how we can improve later.

pramodk:
Ok good for me. We can do that in separate PR.

PR #546, opened 2021-03-08
Add InstanceStruct  test data generation helper and unit test

iomaganaris:
- [x] Taken over work from https://github.com/BlueBrain/nmodl/tree/pramodk/test-helper
- [x] Added mechanism to instantiate the InstanceStruct with deterministic data
- [x] Compare those in a unit test
- [x] Check if typecasting works in the unit test
- [x] Review and cleanup by Pramod


test/unit/codegen/codegen_data_helper.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+std::vector<int> generate_int_data(const size_t& initial_value, const size_t& num_elements) {
+    std::vector<int> data(num_elements);
+
+    for (size_t i = 0; i < num_elements; i++) {
+        data[i] = initial_value + i;
+    }
+
+    return data;
+}
```
pramodk: As code is repeated, it could be templated with specialisation on type.

pramodk:
@georgemitenkov : going to do some cleanup but this overall gives an idea of what creating data data. See `test/unit/codegen/codegen_llvm_instance_struct.cpp`.

georgemitenkov:
@pramodk @iomaganaris looks great! On LLVM/JIT side, I am working on the following:
1. Generating a wrapper LLVM function that takes a void pointer and casts it to the struct LLVM pointer type.
2. Calling a kernel.

Additionally I am thinking of creating functions that can be lowered to LLVM that use `CodegenInstanceData` to print/compare the results.

pramodk:
> Additionally I am thinking of creating functions that can be lowered to LLVM that use CodegenInstanceData to print/compare the results.

I am curious what you meant by `lowered to LLVM` here. If we add such functions in `CodegenInstanceData`, would that fit the purpose?

georgemitenkov:
@pramodk actually you are right, and we do not strictly need it I think. We can always define these on the higher level in `CodegenInstanceData`. I will make changes to the JIT first, and then we will see if we really need something like this:)

georgemitenkov:
@pramodk @iomaganaris I added a sample test case in #549 to show how this would look on LLVM side. I think I have all functions I need in this PR - for the data generation and for member comparisons.

The printing functions are not really necessary - I guess we may need them for visual interpretation only :) So i would say that `printf`s are sufficient for that now!

test/unit/codegen/codegen_llvm_instance_struct.cpp
```diff
@@ -0,0 +1,172 @@
...
+            ASSIGNED {
+                v (mV)
+                celsius (degC)
+                ena (mV)
+                minf
+                mtau
+            }
+
+            BREAKPOINT {
+                SOLVE states METHOD cnexp
```
georgemitenkov, line 97: @iomaganaris @pramodk I have noticed that when the NEURON does not have current, `SOLVE states METHOD cnexp` block is added to the kernel for loop:

```
for ( ... ) {
   SOLVE states METHOD cnexp
   ...
}
```

This does not affect code generation at all, but I am wondering if we should raise a ticket for this :)

test/unit/codegen/codegen_llvm_instance_struct.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    codegen::CodegenLLVMVisitor llvm_visitor(/*mod_filename=*/"test",
+                                             /*output_dir=*/".",
+                                             opt,
+                                             use_single_precision);
+    llvm_visitor.visit_program(*ast);
+    llvm_visitor.print_module();
+    const auto& generated_instance_struct = llvm_visitor.get_instance_struct_ptr();
+    auto codegen_data = codegen::CodegenDataHelper(ast, generated_instance_struct);
+    auto instance_data = codegen_data.create_data(num_elements, 42);
```
georgemitenkov, line 53: nit: static constexpr ... ?

test/unit/codegen/codegen_data_helper.cpp
```diff
@@ -0,0 +1,190 @@
...
+// scalar variables with default values
+const double default_nthread_dt_value = 0.025;
+const double default_nthread_t_value = 100.0;
+const double default_celsius_value = 34.0;
+const int default_second_order_value = 0;
+
+std::vector<double> generate_double_data(size_t initial_value, size_t num_elements) {
+    std::vector<double> data(num_elements);
+    for (size_t i = 0; i < num_elements; i++) {
+        data[i] = initial_value + (i + 1) * 1e-15;
```
georgemitenkov, line 20: nit: const double double_precision = 1e-15 ? Same below?

pramodk:
@iomaganaris : could you look at CI fixes?

iomaganaris:
> @iomaganaris : could you look at CI fixes?

Yy. I am looking at that atm :+1:

test/unit/codegen/codegen_llvm_instance_struct.cpp
pramodk: @georgemitenkov : I also so that somewhere and its a bug. Could you add ticket with some details?

PR #547, opened 2021-03-08
Unit test for scalar state kernel LLVM code generation

georgemitenkov:
This PR adds a unit test to check LLVM instructions generated for the
scalar kernel, particularly:
- For loop blocks.
- Induction variable increments and comparisons.
- Correct loads through GEPs from the struct.

Test for vectorised code generation would be added in a separate PR
or when full vectorisation support (indirect indexing) would land.


pramodk (approved):
LGTM

PR #549, opened 2021-03-09
Running a kernel with NMODL-LLVM JIT

georgemitenkov:
This PR handles the LLVM side of testing and the NMODL kernels.
The following is added:

- A utility in LLVM visitor to create a wrapper function for kernel
execution. It takes a `void*` data, casts it to appropriate instance struct
type and calls the kernel.

- JIT runner now supports functions without arguments and with a single
argument. The latter is useful to verify `FUNCTION` results and pass the
instance struct data to the kernel.

- A simple scalar kernel test  is added.

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -923,5 +923,41 @@ void CodegenLLVMVisitor::visit_while_statement(const ast::WhileStatement& node) 
     builder.SetInsertPoint(exit);
 }
 
+void CodegenLLVMVisitor::wrap_kernel_function(const std::string& kernel_name) {
+    // Get the kernel function and the instance struct type.
+    auto kernel = module->getFunction(kernel_name);
+    if (!kernel)
+        throw std::runtime_error("Kernel " + kernel_name + " is not found!");
+
+    if (kernel->getNumOperands() != 0)
+        throw std::runtime_error("Kernel " + kernel_name + " has too many arguments!");
+
+    auto instance_struct_ptr_type = llvm::dyn_cast<llvm::PointerType>(kernel->getArg(0)->getType());
+    if (!instance_struct_ptr_type)
+        throw std::runtime_error("Kernel " + kernel_name +
+                                 " does not have an instance struct pointer argument!");
+
+    // Create a wrapper void function that takes a void pointer as a single argument.
+    llvm::Type* void_type = llvm::Type::getVoidTy(*context);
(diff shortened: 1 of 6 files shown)


src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -919,5 +919,41 @@ void CodegenLLVMVisitor::visit_while_statement(const ast::WhileStatement& node)
     builder.SetInsertPoint(exit);
 }
 
+void CodegenLLVMVisitor::wrap_kernel_function(const std::string& kernel_name) {
+    // Get the kernel function and the instance struct type.
+    auto kernel = module->getFunction(kernel_name);
+    if (!kernel)
+        throw std::runtime_error("Kernel " + kernel_name + " is not found!");
+
+    if (kernel->getNumOperands() != 0)
```
iomaganaris, line 928: I am wondering how this is working. Does the `getNumOperands()` give the number of arguments of `llvn::Function`? Is yes, shouldn't this be one (the `InstanceStruct*`)?

georgemitenkov: That's a mistake! Thanks! `getNumOperands()` comes from `llvm::Value` class, so in case of functions this turns out to be true. What I meant here was the number of arguments (basically `args()`) which indeed should be 1.

pramodk:
@georgemitenkov : fyi, #546 is now merged. Let me know if anything you need to complete this PR and move towards benchmarking aspects.

test/unit/codegen/codegen_llvm_execution.cpp
```diff
@@ -189,23 +236,101 @@ SCENARIO("Optimised arithmetic expression", "[llvm][runner]") {
...
+        std::unique_ptr<llvm::Module> module = llvm_visitor.get_module();
+        Runner runner(std::move(module));
+
+        THEN("Values in struct have changed according to the formula") {
+            runner.run_with_argument<int, void*>("__nrn_state_test_wrapper",
+                                                 instance_data.base_ptr);
+            std::vector<double> x_expected = {4.0, 3.0, 2.0, 1.0};
+            REQUIRE(check_instance_variable(instance_info, x_expected, "x"));
+        }
+    }
```
pramodk, line 335: very nice! this kind of tests would be useful for C++ code generation backends as well (cc: @alkino  @iomaganaris )

pramodk (approved):
LGTM

iomaganaris (approved):
LGTM as well.
I agree that these test could be used with the other backends as well. I believe it's a base for the benchmarking framework as well.

PR #553, opened 2021-03-16
Add log10 function to ISPC

### LINKED ISSUE #464: log10 not defined in ISPC
`iccr.mod` coming from https://github.com/Neurosim-lab/netpyne/tree/coreneuron/examples/coreneuron/m1/mod uses `log10` function which doesn't exist in `ISPC`.
```
x86_64/corenrn/mod2c/iccr.ispc:83:39: Error: Undeclared symbol "log10". Did you mean 
        "log"? 
                vs = v_in_0 + 40.0d * log10(1000.0d * cai_in_0);
                                      ^^^^^
```

iomaganaris:
Some mod files use `log10` function which is not defined by `ISPC`.
This PR takes care of:
- [x] Add `log10` functions in `ISPC`
- [x] Add `log10` functions in `fast_math.hpp` (`static` cannot be used because of clash of declaration with `std::log10`)
- [x] Add tests for the new functions

Fixes #464

### DIFF
--- a/src/codegen/fast_math.hpp
+++ b/src/codegen/fast_math.hpp
@@ -217,11 +217,11 @@ static inline float exprelr(float x) {
 }
 
 /// double precision log10 function
-inline double log10(double f) {
+inline double log10(double f) noexcept {
     return log(f) / log(10.0);
 }
 
 /// single precision log10 function
-inline float log10(float f) {
+inline float log10(float f) noexcept {
     return log(f) / log(10.0f);
 }
--- a/src/codegen/fast_math.hpp
+++ b/src/codegen/fast_math.hpp
@@ -215,3 +215,13 @@ static inline float exprelr(float x) {
 
     return x / vexpm1(x);
 }
+
+/// double precision log10 function
+inline double log10(double f) {
+    return log(f) / log(10.0);
+}
+
+/// single precision log10 function
+inline float log10(float f) {
+    return log(f) / log(10.0f);
+}
--- a/src/codegen/fast_math.hpp
+++ b/src/codegen/fast_math.hpp
@@ -18,6 +18,9 @@
  * \brief Implementation of different math functions
  */
 
+namespace nmodl {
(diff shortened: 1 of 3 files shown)


cattabiani (approved):
LGTM!

src/codegen/fast_math.hpp
```diff
@@ -214,4 +214,14 @@ static inline float exprelr(float x) {
     }
 
     return x / vexpm1(x);
-}
\ No newline at end of file
+}
+
+/// double precision log10 function
+inline double log10(double f) noexcept {
+    return log(f) / log(10.0);
```
castigli, line 221: for efficiency I would write `1/log(10.0)` as a `constexpr` and remove the runtime division

iomaganaris: I thought that later actually and I was planning to do that. Thanks for suggestion! :+1:

test/unit/fast_math/fast_math.cpp
```diff
@@ -96,6 +96,20 @@ SCENARIO("Check fast_math") {
     GIVEN("exprelr (float)") {
         auto test = check_over_span(exprelr_ref, exprelr, low_limit_f, high_limit_f, npoints);
 
+        THEN("error inside threshold") {
+            REQUIRE(test);
+        }
+    }
+    GIVEN("log10 (double)") {
+        auto test = check_over_span(std::log10, log10, low_limit, high_limit, npoints);
```
castigli, line 104: the limits should be different for `log10`, I believe the range should ~ ]std::numeric_limits<T>::min(), std::numeric_limits<T>::max()[

castigli (changes requested):
Apart from possibly fixing an efficiency detail and the tested range, it looks good!

src/codegen/fast_math.hpp
cattabiani: +1

castigli (approved):

PR #554, opened 2021-03-16
Always initialize return variable in function block

### LINKED ISSUE #530: Return variable is not initialised in the AST transformation
If we have mod file procedure like:

```
 PROCEDURE loop_proc(v, t) {
     LOCAL i
     i = 0
     WHILE(i < 10) {
         printf("Hello World")
         i = i + 1
     }
 }
```

then if we run AST transformations from LLVM branch with `./bin/nmodl hh.mod llvm --ir` then we get:

```
INTEGER loop_proc(DOUBLE arg_v, DOUBLE t){
    INTEGER ret_loop_proc
    DOUBLE i
    i = 0
    WHILE (i<10) {
        printf("Hello World")
        i = i+1
    }
    return ret_loop_proc
}
```

Here `ret_loop_proc` is not initialised to 0.

alkino:
Fix #530 

It was already fixed when generating llvm IR. FIx it earlier, so when looking at intermediate AST the variable is added in the same time that the initialization.

Delete the code that fix it in IR

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -39,7 +39,8 @@ static ast::VarName* create_varname(const std::string& varname) {
  *         including Integer. See #542.
  */
 static std::shared_ptr<ast::Expression> int_initialization_expression(
-    const std::string& induction_var, int value = 0) {
+    const std::string& induction_var,
+    int value = 0) {
     // create id = 0
     const auto& id = create_varname(induction_var);
     const auto& zero = new ast::Integer(value, nullptr);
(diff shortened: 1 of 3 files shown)


pramodk (commented):
LGTM !

@alkino : I wonder if we could add one test in `test/unit/codegen/codegen_helper.cpp` ? e.g. similar to what you did for remainder loop but now you add simple PROCEDURE and after transformation check if it's return value is correctly initialised?

alkino:
The test is already here.

https://github.com/BlueBrain/nmodl/blob/llvm/test/unit/codegen/codegen_llvm_ir.cpp#L670-L691

pramodk:
@alkino : true. But LLVM build is optional and hence better to also have NMODL only test.

pramodk (approved):
Discussed this offline. Transformation related test will be added separately.

PR #555, opened 2021-03-16
Improved log10 by using multiplication

iomaganaris:


src/codegen/fast_math.hpp
```diff
@@ -78,8 +78,8 @@ static constexpr float PX6expf = 5.0000001201E-1f;
 
 static constexpr float LOG2EF = 1.44269504088896341f;  // 1/ln(2)
 
-static constexpr double LOG10E = 2.302585092994046;  // log(10)
-static constexpr float LOG10F = 2.302585f;           // log(10)
+static constexpr double LOG10E = 0.4342944819032518;  // 1/log(10)
+static constexpr float LOG10F = 0.4342945f;           // 1/log(10)
```
pramodk, line 82: I will ask @castigli about how constants are typically specified considering precision.

castigli: that is a good point,  you should store the constants with enough digit to preserve the exact representation, i.e. `1.0/std::log(10) == LOG10E` . 
https://godbolt.org/z/n4GKf5
Long story short you need to add another digit for double or let the compiler do the job
```suggestion
static constexpr double LOG10E = 1/std::log(10);  // 1/log(10)
```

castigli: (I didn't check the float)

castigli (approved):
forget my previous comments, the constant is good

src/codegen/fast_math.hpp
castigli: I guess it is nice to write in scientific notation, but the value itself if good (forget my previous comments!)

iomaganaris: Sorry for late reply here. I didn't have much time to look at this yesterday. The godbolt program was very informative and I was taking some time to try and understand some of the parts.
Just FYI the way I generated the constants was: https://godbolt.org/z/EaG1Mn
I agree it would be also nice to have the scientific notation but I just forgot about it :facepalm:

PR #556, opened 2021-03-19
Move python file to a new directory called python instead of nmodl

alkino:
This way, it's more clear and there is no more bug when importing `nmodl` inside a python script.


setup.py
```diff
@@ -110,7 +110,7 @@ def _config_exe(exe_name):
     author_email="bbp-ou-hpc@groupes.epfl.ch",
     description="NEURON Modeling Language Source-to-Source Compiler Framework",
     long_description="",
-    packages=["nmodl"],
+    packages=["python"],
```
pramodk, line 113: Question : should we really change package name in setup.py? does this used somewhere else? e.g. pip show or pip list command?

alkino:
This it too much complicate for the moment because setup.py wait files are in a specific directory.

PR #559, opened 2021-03-25
Add flexibility when using or not legacy units

### LINKED ISSUE #416: nmodl needs same units configuration flexibility as mod2c
nmodl should have the same configuration flexibility (not dynamic) for unit selection as the current configuration flexibility
of mod2c (see BlueBrain/mod2c#53).
I believe this would close all issues in coreneuron and nmodl that involve units.

alkino:
Fix #416


pramodk (approved):
LGTM

GitLab CI is failing with clang-format

PR #561, opened 2021-03-25
Try to use pgi as compiled in ci

alkino:


.gitlab-ci.yml
```diff
@@ -29,5 +34,30 @@ test intel: &test_config
...
+
+test pgi:
+  extends: .common
+  stage: test
+  resource_group: test
+  variables:
+    # We cloned in .pre
+    GIT_STRATEGY: none
+
+# test pgi: &test_config
```
pramodk, line 54: This should be removed?

pramodk (approved):
Based on gitlab CI status, this looks reasonable to me

PR #562, opened 2021-03-29
Add support for ConstantBlock

### LINKED ISSUE #323: CONSTANT block not handled
Currently when there is a `CONSTANT` block defined in a mod file there is an error generated and the code generation stops. Maybe we should transform the definitions of the `CONSTANT` block to `UNITS` inside `NMODL`.
Example mod file: https://github.com/Neurosim-lab/netpyne/blob/development/examples/M1detailed/mod/cadad.mod

alkino:
Fix #323

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -723,7 +723,7 @@ bool CodegenCVisitor::ion_variable_struct_required() const {
 /**
  * \details This can be override in the backend. For example, parameters can be constant
  * except in INITIAL block where they are set to 0. As initial block is/can be
- * executed on c/cpu backend, gpu/cuda backend can mark the parameter as constnat.
+ * executed on c/cpu backend, gpu/cuda backend can mark the parameter as constant.
  */
 bool CodegenCVisitor::is_constant_variable(const std::string& name) const {
     auto symbol = program_symtab->lookup_in_scope(name);
@@ -2009,7 +2009,7 @@ std::string CodegenCVisitor::process_shadow_update_statement(ShadowUseStatement&
 void CodegenCVisitor::print_nmodl_constants() {
     if (!info.factor_definitions.empty()) {
         printer->add_newline(2);
-        printer->add_line("/** constants used in nmodl */");
+        printer->add_line("/** constants used in nmodl from UNITS */");
         for (const auto& it: info.factor_definitions) {
 #ifdef USE_LEGACY_UNITS
             const std::string format_string = "static const double {} = {:g};";
@@ -2019,6 +2019,19 @@ void CodegenCVisitor::print_nmodl_constants() {
             printer->add_line(fmt::format(format_string,
                                           it->get_node_name(),
                                           stod(it->get_value()->get_value())));
+        }
+    }
+    if (!info.constants_definitions.empty()) {
+        printer->add_newline(2);
+        printer->add_line("/** constants used in nmodl from CONSTANTS */");
+        for (const auto& it: info.constants_definitions) {
+            const std::string format_string = "static const {} {{}} = "_format(
+                it->get_value()->get_node_type() == AstNodeType::DOUBLE ? "double" : "int");
+            printer->add_indent();
+            printer->add_text(fmt::format(format_string, it->get_name()->get_value()->get_value()));
+            it->get_value()->accept(*this);
+            printer->add_text(";");
+            printer->add_newline();
         }
     }
 }
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -2021,19 +2021,6 @@ void CodegenCVisitor::print_nmodl_constants() {
                                           stod(it->get_value()->get_value())));
         }
     }
-    if (!info.constants_definitions.empty()) {
-        printer->add_newline(2);
-        printer->add_line("/** constants used in nmodl from CONSTANTS */");
-        for (const auto& it: info.constants_definitions) {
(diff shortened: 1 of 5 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -2021,6 +2021,19 @@ void CodegenCVisitor::print_nmodl_constants() {
...
     }
+    if (!info.constants_definitions.empty()) {
+        printer->add_newline(2);
+        printer->add_line("/** constants used in nmodl from CONSTANTS */");
+        for (const auto& it: info.constants_definitions) {
+            const std::string format_string = "static const {} {{}} = "_format(
+                it->get_value()->get_node_type() == AstNodeType::DOUBLE ? "double" : "int");
+            printer->add_indent();
+            printer->add_text(fmt::format(format_string, it->get_name()->get_value()->get_value()));
+            it->get_value()->accept(*this);
```
alkino, line 2032: no nicier way?

alkino:
BTW my patch seems useless.

CONSTANTS are already handled in a really bad manner

src/codegen/codegen_c_visitor.cpp
cattabiani: usually I use the proper visit function instead of accept. However, this works as well. For the printing lines I let @pramodk talk because I always pick the wrong way to print stuff. I would check how the range variables are declared

cattabiani:
may you point me to where they are handled?

src/codegen/codegen_c_visitor.cpp
cattabiani: maybe we can use constexpr?

cattabiani (commented):
I did not investigate thoroughly, I just put here what could be a corner case: what happens if we re-define a variable like FARADAY in CONSTANTS?

alkino:
> may you point me to where they are handled?

I don't really know but in the outside cpp we got:
```    /** all global variables */
    struct cadad_Store {
        int ca_type;
        double cai0;
        int reset;
        int mech_type;
        **double FARADAY;**
        int* slist1;
        int* dlist1;
        ThreadDatum* ext_call_thread;
    };
```

cattabiani:
I have found the problem I had with units. in cagk.mod in hippocampus there is this block:

```
UNITS {
	FARADAY = (faraday)  (kilocoulombs)
	R = 8.313424 (joule/degC)
}
```

I wonder:

is it possible to have the same thing with CONSTANTS? I mean, is it still a valid mod file? If yes, how do we deal with ispc? Because I fear these constants get declared twice

iomaganaris:
> I have found the problem I had with units. in cagk.mod in hippocampus there is this block:
> 
> ```
> UNITS {
> 	FARADAY = (faraday)  (kilocoulombs)
> 	R = 8.313424 (joule/degC)
> }
> ```
> 
> I wonder:
> 
> is it possible to have the same thing with CONSTANTS? I mean, is it still a valid mod file? If yes, how do we deal with ispc? Because I fear these constants get declared twice

I agree that this is an issue and something that `NEURON` for example doesn't take care of.
I think we could print some error in this case from `NMODL` though :+1:

src/codegen/codegen_c_visitor.cpp
pramodk: > no nicier way?

we don't need to use `accept` to visit the element. Not short but bit straightforward is:

```cpp
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -2025,13 +2025,16 @@ void CodegenCVisitor::print_nmodl_constants() {
         printer->add_newline(2);
         printer->add_line("/** constants used in nmodl from CONSTANTS */");
         for (const auto& it: info.constants_definitions) {
-            const std::string format_string = "static const {} {{}} = "_format(
-                it->get_value()->get_node_type() == AstNodeType::DOUBLE ? "double" : "int");
-            printer->add_indent();
-            printer->add_text(fmt::format(format_string, it->get_name()->get_value()->get_value()));
-            it->get_value()->accept(*this);
-            printer->add_text(";");
-            printer->add_newline();
+            const auto& name = it->get_name()->get_node_name();
+            std::string format_string;
+            if(it->get_value()->is_double()) {
+                const auto& value = std::dynamic_pointer_cast<ast::Double>(it->get_value());
+                format_string = "static const double {} = {}"_format(name, value->get_value());
+            } else {
+                const auto& value = std::dynamic_pointer_cast<ast::Integer>(it->get_value());
+                format_string = "static const int {} = {}"_format(name, value->get_value());
+            }
+            printer->add_line(format_string);
         }
```

pramodk:
> I did not investigate thoroughly, I just put here what could be a corner case: what happens if we re-define a variable like FARADAY in CONSTANTS?

I wouldn't about this in code generation pass. This is `todo` in semantic analysis pass where such error should happen (there is no such pass but that is for future task).

cattabiani:
> > I did not investigate thoroughly, I just put here what could be a corner case: what happens if we re-define a variable like FARADAY in CONSTANTS?
> 
> I wouldn't about this in code generation pass. This is `todo` in semantic analysis pass where such error should happen (there is no such pass but that is for future task).

in units it IS handled. We need to remember to remove/modify the code when we do the semantic visitor

pramodk (approved):
LGTM!

Discussed offline - `ConstantBlock` was already handled. It's just that compatibility visitor put us in wrong direction. Oversight on my part!

iomaganaris (approved):
LGTM now. Oversight from my part as well. All should be good now

PR #565, opened 2021-03-30
Remove support of AoS layout

### LINKED ISSUE #563: Remove support for AoS layout from NMODL
NMODL currently has support for AoS layout. This was added for compatibility and testing with MOD2C and CoreNEURON.
I think we should remove AoS support due to following reasons:

* In practice, we never use AoS layout for CPU vectorisation or GPUs
* This part remains untested because of the same reason
* The code generation code could become simple and streamlined

But what if we need this for NEURON or some other purpose?

* The changes are only at the final level i.e. code generation stage 
* NEURON will have different code generation printing class. So there we could easily add AoS related logic

What needs to be done as part of this Issue?

* Remove CLI option in nmodl for layout 
* Remove all logic in src/codegen where we we are using `LayoutType::aos`. Simplify code assuming `LayoutType::soma` is default.

alkino:
Consider that we always use SoA.

NMODL currently has support for AoS layout. This was added for compatibility and testing with MOD2C and CoreNEURON.
We remove AoS support due to following reasons:

    In practice, we never use AoS layout for CPU vectorisation or GPUs
    This part remains untested because of the same reason
    Simplify the code

What have been done:
    Remove CLI option in nmodl for layout
    Remove all logic in src/codegen where we we are using LayoutType::aos.

Fix #563

### DIFF
--- a/src/nmodl/main.cpp
+++ b/src/nmodl/main.cpp
@@ -152,9 +152,6 @@ int main(int argc, const char* argv[]) {
     /// true if symbol table should be printed
     bool show_symtab(false);
 
-    /// memory layout for code generation
-    std::string layout("soa");
-
     /// floating point data type
     std::string data_type("double");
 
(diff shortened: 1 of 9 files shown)


src/nmodl/main.cpp
```diff
@@ -247,14 +244,6 @@ int main(int argc, const char* argv[]) {
...
 
     auto codegen_opt = app.add_subcommand("codegen", "Code generation options")->ignore_case();
-    codegen_opt->add_option("--layout",
-        layout,
-        "Memory layout for code generation",
-        true)->ignore_case()->check(CLI::IsMember({"aos", "soa"}));
-    codegen_opt->add_option("--datatype",
-        layout,
-        "Data type for floating point variables",
-        true)->ignore_case()->check(CLI::IsMember({"float", "double"}));
```
alkino, line 257: What do you think of it?

pramodk: do you mean removing data type option as well?

alkino: Yes, because the output of the option is ... `layout`, copy/paste error, so I guess that no one use it? ^^

pramodk: @alkino : its true but data type will be needed for sure in the near future. So I will keep it for now.

pramodk (approved):
LGTM


@alkino : One separate todo on coreneuron side: if nmodl is enabled and is AoS layout is asked then we should give error from CMake.

PR #566, opened 2021-04-01
Move main.cpp at the root of src

alkino:
For an easier understanding of sources put main.cpp at the root of src and create there a CMakeLists.txt delegate to build all the sources.


alkino:
Please retest

pramodk (approved):
LGTM

PR #567, opened 2021-04-05
Loop epilogue fix for LLVM visitor helper

georgemitenkov:
LLVM IR code generation relies on unique SSA value names
within function blocks. In order to improve readability, certain
variable names are preserved. This poses a problem when two
for loops declare local variables with the same name.

This PR introduces basic renaming to LLVM helper visitor. All
loop epilogue local variable declarations are prefixed with
`__epilogue__` to avoid symbolic conflicts.

Also, this PR fixes the following issues:
- Loop epilogue is NOT generated for the scalar case anymore

- Vector width within LLVM visitor is adjusted to reflect the scalar
case

- Vectorised loop trip count is now set to
`node_count - vector_width + 1` to prevent accesses out of bounds

The tests for the `nrn_state` to loop transformation have been
refactored and a test for scalar case has been added.

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -26,7 +26,7 @@ const std::string CodegenLLVMHelperVisitor::NODECOUNT_VAR = "node_count";
 const std::string CodegenLLVMHelperVisitor::VOLTAGE_VAR = "voltage";
 const std::string CodegenLLVMHelperVisitor::NODE_INDEX_VAR = "node_index";
 
-static constexpr const char epilogue_variable_prefix[] = "__epilogue__";
+static constexpr const char epilogue_variable_prefix[] = "epilogue_";
 
 /// Create asr::Varname node with given a given variable name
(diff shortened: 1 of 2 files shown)


src/codegen/llvm/codegen_llvm_helper_visitor.cpp
```diff
@@ -611,17 +642,30 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
...
+        // \todo: Change RenameVisitor to take a vector of names to which it would append a single
+        // prefix.
+        for (const auto& name: int_variables) {
+            visitor::RenameVisitor v(name, epilogue_variable_prefix + name);
+            loop_statements->accept(v);
+        }
+        for (const auto& name: double_variables) {
+            visitor::RenameVisitor v(name, epilogue_variable_prefix + name);
+            loop_statements->accept(v);
+        }
```
pramodk, line 673: @georgemitenkov : could you remind myself why we need to rename variables? I think mentioned it but I vaguely remember - variables need to have unique names across different basic blocks?

pramodk (approved):
I did very minor nitpicks. LGTM.

PR #568, opened 2021-04-06
Gather support and vectorisation fixes for LLVM code generation

georgemitenkov:
This PR fixes vector code generation, particularly now:

- The pointer to the element in the array inside instance struct
id bitcasted to vector *after* the GEP instruction.
- Local variables in the loop are vectorised, and the array
indexing can now be performed with a vector of indices.
- Gather instruction support has been added.

The tests for IR and vectorised code execution are created
accordingly.

- [x] TODO: add IR tests for gather instruction

### DIFF
--- a/test/unit/codegen/codegen_llvm_ir.cpp
+++ b/test/unit/codegen/codegen_llvm_ir.cpp
@@ -902,6 +902,61 @@ SCENARIO("Scalar state kernel", "[visitor][llvm]") {
 }
 
 //=============================================================================
+// Gather for vectorised kernel
+//=============================================================================
+
+SCENARIO("Vectorised simple kernel", "[visitor][llvm]") {
+    GIVEN("An indirect indexing of voltage") {
+        std::string nmodl_text = R"(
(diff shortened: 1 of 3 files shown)


pramodk:
I will see why CI is failing with below error:
image

test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -901,6 +901,61 @@ SCENARIO("Scalar state kernel", "[visitor][llvm]") {
...
+            ASSIGNED {
+                v (mV)
+            }
+
+            BREAKPOINT {
+                SOLVE states METHOD cnexp
+                i = 2
+            }
+
+            DERIVATIVE states {}
```
pramodk, line 927: Just a note : in future PR, we could add `var_x = v` in the DERIVATIVE block and that could be used as a test for gather.

pramodk (approved):
Not intimately familiar with codegen_llvm_visitor.cpp  but overall LGT

PR #569, opened 2021-04-06
Fix ci for old pip version

alkino:
sphinx ask for docutils > 0.12
sphinx-rtd-theme for docutils < 0.17

if sphinx-rtd-theme is installed first it seems that sphinx install a upper version of docutils.

With newer version of python or pip (I don't know), it takes care of it, but not on ubuntu 18.04 and/or macos


pramodk (approved):
looks reasonable (or, docutil could be added explicitly in the list)

PR #571, opened 2021-04-07
Fix documentation to reduce warnings

alkino:
This patch try to reduce the number of warnings and error when building documentation.


setup.py
```diff
@@ -121,6 +121,7 @@ def _config_exe(exe_name):
     zip_safe=False,
     setup_requires=[
         "jinja2>=2.9.3",
+        "jupyter-client==6.1.12",
```
pramodk, line 124: In principle, building NMODL doesn't require Jupyter. Once we will upload source wheels to pypi.org, building via pip will also install whole Jupyter stuff. So we should abvoid it from setup.py?

pramodk (changes requested):
I have one comment/question about setup.py

I assume you have checked the style/format of generated oxygen doc.

setup.py
alkino: jupyter is already a depends of nmodl for building notebooks in documentation

pramodk: discussed offline - jupyter is already installed. This merely installs different jupyter-client version due to bug.

pramodk (approved):

PR #574, opened 2021-04-08
Simplify output of azure ci for unbuilded wheel

alkino:
On master only the wheel for python 3.9 is generated, so disable matrix for
other versions of python.
During the nightly generate wheel for all versions.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,8 +5,8 @@ variables:
   # defined in UI
   #- ReleaseWheelBuild: False
   #- NightlyWheelBuild: False
-  buildWheel: $[or(in(variables['Build.Reason'], 'Schedule'),
-              and(in(variables['Build.Reason'], 'Manual'), eq(variables.NightlyWheelBuild, True)))]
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -181,22 +184,27 @@ jobs:
...
+        Python37:
+          python.version: '3.7'
+          python.org.version: '3.7.7'
+        Python38:
+          python.version: '3.8'
+          python.org.version: '3.8.2'
+        Python39:
+          python.version: '3.9'
+          python.org.version: '3.9.0'
+      ${{ if ne(variables.buildWheel, True) }}:
```
ohm314, line 200:
```suggestion
      ${{ if eq(variables.buildWheel, False) }}:
```
maybe just a matter of taste.. but this seems a bit more readable

ohm314 (approved):
Really cool! Thanks for figuring this out :)

alkino:
Please retest

alkino:
Please retest

PR #576, opened 2021-04-08
 jupyter-client is already a depends of jupyter

alkino:
jupyter-client is already a depends of jupyter

We no more need to fix the version because the wrong package (6.1.13) has been
yanked in PyPi

jupyter/jupyter_client#637


iomaganaris:
please retest

alexsavulescu (approved):
Great, they fixed this upstream.

PR #578, opened 2021-04-09
Automatically enable sympy_analytics pass if there is a sparse solver

alkino:
* Those kind of solvers always need a sympy pass because they contains ODE.
* Enable it for user if he didn't ask for.


pramodk (approved):
If CI passes, this LGTM

PR #582, opened 2021-04-12
Verification and file utilities for LLVM codegen

georgemitenkov:
This PR is a minor improvement of the current pipeline infrastructure.
Particularly, it addresses the following:

- The generated IR module is now verified after running the visitor
- The kernel is checked if it can be vectorised or not
- The generated IR can be dumped to `.ll` file with `-o <filename>`
- Printing LLVM IR is moved to debug mode

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -9,13 +9,17 @@
 
 #include "ast/all.hpp"
 #include "visitors/rename_visitor.hpp"
+#include "visitors/visitor_utils.hpp"
 
+#include "llvm/IR/AssemblyAnnotationWriter.h"
 #include "llvm/IR/BasicBlock.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Function.h"
 #include "llvm/IR/LLVMContext.h"
 #include "llvm/IR/Type.h"
 #include "llvm/IR/ValueSymbolTable.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/ToolOutputFile.h"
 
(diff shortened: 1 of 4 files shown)


georgemitenkov:
This is also a pre-benchamrk PR :)

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -28,12 +32,31 @@ static constexpr const char instance_struct_type_name[] = "__instance_var__type"
...
+/// A utility to check for supported Statement AST nodes.
 static bool is_supported_statement(const ast::Statement& statement) {
     return statement.is_codegen_var_list_statement() || statement.is_expression_statement() ||
            statement.is_codegen_for_statement() || statement.is_codegen_return_statement() ||
            statement.is_if_statement() || statement.is_while_statement();
 }
 
+/// A utility to check of the kernel body can be vectorised.
+static bool can_vectorise(const ast::CodegenForStatement& statement, symtab::SymbolTable* sym_tab) {
+    // Check that function calls are made to external methods only.
```
pramodk, line 44: Just a note - this will be expanded bit further in future but don't hesitate to comment out certain constructs from DERIVATIVE blocks if you see something not compatible :) (e.g. `if statements` are ok to comment for benchmarking purpose)

pramodk (approved):
LGTM

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: Yes! This is just a useful precondition to catch unhandled cases rather than figure out what went wrong in IR:) (If not commented out)

PR #583, opened 2021-04-14
Benchmarking LLVM code generation

georgemitenkov:
This PR is to introduce the benchmarking for LLVM code
generation pipeline. For that, new options have been added:

```
benchmark
  LLVM benchmark option
  Options:
    --run                                           Run LLVM benchmark (false)
    --instance-size INT                   Instance struct size (10000)
    --repeat INT                               Number of experiments for benchmarking (100)
    --backend TEXT:{avx2, default, sse2}         Target's backend (default)
```

The `backend` option does not work yet (as LLVM can
automatically set up the target/data layout information).

Example:
```bash
$ nmodl hh.mod llvm --ir --vector-width 1 benchmark --run --instance-size 100000 --repeat 10 --backend default

Created LLVM IR module from NMODL AST in 0.006765817

Benchmarking kernel 'nrn_state_hh'
Experiment 0: compute time = 0.013977749
Experiment 1: compute time = 0.004847989
Experiment 2: compute time = 0.004739051
Experiment 3: compute time = 0.004801505
Experiment 4: compute time = 0.005540785
Experiment 5: compute time = 0.004965192
Experiment 6: compute time = 0.004840146
Experiment 7: compute time = 0.00485995
Experiment 8: compute time = 0.005134093
Experiment 9: compute time = 0.004844469
Average compute time = 0.0058550929
```

### DIFF
--- a/src/codegen/llvm/jit_driver.hpp
+++ b/src/codegen/llvm/jit_driver.hpp
@@ -63,8 +63,9 @@ class JITDriver {
         return result;
     }
 
-    /// Set the target triple on the module.
-    static void set_target_triple(llvm::Module* module);
+    /// A wrapper around llvm::createTargetMachine to turn on/off certain CPU features.
+    std::unique_ptr<llvm::TargetMachine> create_target(llvm::orc::JITTargetMachineBuilder* builder,
+                                                       const std::string& features);
 };
 
 /**
--- a/src/codegen/llvm/jit_driver.hpp
+++ b/src/codegen/llvm/jit_driver.hpp
@@ -85,7 +85,7 @@ class Runner {
     }
 
     Runner(std::unique_ptr<llvm::Module> m, std::string features)
-            : module(std::move(m)) {
+        : module(std::move(m)) {
         driver->init(features);
     }
 
--- a/src/codegen/llvm/jit_driver.hpp
+++ b/src/codegen/llvm/jit_driver.hpp
@@ -79,12 +79,7 @@ class Runner {
     std::unique_ptr<JITDriver> driver = std::make_unique<JITDriver>(std::move(module));
 
   public:
-    Runner(std::unique_ptr<llvm::Module> m)
-        : module(std::move(m)) {
-        driver->init(/*features=*/"");
-    }
-
-    Runner(std::unique_ptr<llvm::Module> m, std::string features)
+    Runner(std::unique_ptr<llvm::Module> m, std::string features = "")
         : module(std::move(m)) {
         driver->init(features);
     }
--- a/src/codegen/llvm/jit_driver.hpp
+++ b/src/codegen/llvm/jit_driver.hpp
@@ -37,7 +37,7 @@ class JITDriver {
         : module(std::move(m)) {}
 
     /// Initialize the JIT.
-    void init();
+    void init(std::string features);
 
     /// Lookup the entry-point without arguments in the JIT and execute it, returning the result.
     template <typename ReturnType>
@@ -81,7 +81,12 @@ class Runner {
   public:
     Runner(std::unique_ptr<llvm::Module> m)
         : module(std::move(m)) {
-        driver->init();
+        driver->init(/*features=*/"");
+    }
+
+    Runner(std::unique_ptr<llvm::Module> m, std::string features)
+            : module(std::move(m)) {
+        driver->init(features);
     }
 
     /// Run the entry-point function without arguments.
--- a/src/codegen/llvm/llvm_benchmark.cpp
+++ b/src/codegen/llvm/llvm_benchmark.cpp
@@ -9,7 +9,7 @@
 #include "codegen/llvm/codegen_llvm_visitor.hpp"
 #include "codegen/llvm/jit_driver.hpp"
 
-#include "../test/unit/codegen/codegen_data_helper.hpp"
+#include "test/unit/codegen/codegen_data_helper.hpp"
 
 #include <chrono>
 #include <fstream>
--- a/src/codegen/llvm/llvm_benchmark.cpp
+++ b/src/codegen/llvm/llvm_benchmark.cpp
@@ -7,6 +7,9 @@
 
 #include "llvm_benchmark.hpp"
 #include "codegen/llvm/codegen_llvm_visitor.hpp"
+#include "codegen/llvm/jit_driver.hpp"
+
+#include "../test/unit/codegen/codegen_data_helper.hpp"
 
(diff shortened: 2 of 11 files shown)


georgemitenkov:
@pramodk I had to change some Cmake dependencies to get CodegenDataHelper from `tests`. I am thinking of moving this utility to `src` instead, but unsure where to place it (I guess other backends would use it as well). Any ideas?

pramodk:
Copying Slack conversation here:

image

pramodk:
GitLab CI is failing with errors like:

image

Seems like include path needs to be adjusted. 

I can fix that if you can't reproduce that locally. Apart from that, is this ready for review?

pramodk:
Another question @georgemitenkov : on your laptop, does this work with SSE and AVX? i.e. vector width 1, 2 and 4? Just to be sure before I start testing :)

georgemitenkov:
@pramodk So far I have an issue with large kernels (size > 1000000), that fail. Apart from that, the code is more or less ready for review. There are 2 outstanding issues:

- [ ] Fix include path
- [ ] Propagate avx/sse info to JIT (Currently only default target triple is supported, and I am figuring out how to propagate it there)

pramodk:
With last fixes, CI should be good. 

FYI : if you add new library (`llvm_benchmark`) and if those cpp files are using AST headers then that library needs to be compiled *after* AST header files are generated. In order to achieve that, we have add dependency with the target that generate AST headers .e.g `add_dependencies(llvm_benchmark lexer)`.

I would like to make some cleanup/rearrangements - e.g. I am thinking benchmark related files could be outside src/codegen. For example, separate top level directory benchmark. But lets see this tomorrow.

georgemitenkov:
> With last fixes, CI should be good.
> 
> FYI : if you add new library (llvm_benchmark) and if those cpp files are using AST headers then that library needs to be compiled after AST header files are generated. In order to achieve that, we have add dependency with the target that generate AST headers .e.g add_dependencies(llvm_benchmark lexer).
Thank you! I see, I suppose in my case it never failed because I had AST headers cached and didn't do a clean build.

> I would like to make some cleanup/rearrangements - e.g. I am thinking benchmark related files could be outside src/codegen. For example, separate top level directory benchmark. But lets see this tomorrow.
Yes, that's a good point. We can discuss this more later.

Regarding the bug, I have found that it is `mech->node_count` that has a weird value (due to misaligned access I suppose).

src/nmodl/main.cpp
```diff
@@ -587,7 +617,16 @@ int main(int argc, const char* argv[]) {
             }
 
 #ifdef NMODL_LLVM_BACKEND
-            if (llvm_ir) {
+
+            if (run_benchmark) {
+                logger->info("Running LLVM benchmark");
+                benchmark::LLVMBuildInfo info{llvm_vec_width, llvm_opt_passes, llvm_float_type};
+                benchmark::LLVMBenchmark bench(
+                    modfile, output_dir, info, repeat, instance_size, backend);
```
iomaganaris, line 625: Maybe it would make sense to set the `llvm_vec_width` based on `backend` if vector width value is not defined by the user?

src/codegen/llvm/llvm_benchmark.cpp
```diff
@@ -0,0 +1,157 @@
...
+    } else if (backend == "sse2") {
+        // Disable AVX.
+        *log_stream << "Disabling features:\n";
+        disable("avx", features);
+        disable("avx2", features);
+    }
+
+    std::string features_str = llvm::join(features.begin(), features.end(), ",");
+    std::unique_ptr<llvm::Module> m = visitor.get_module();
+    runner::Runner runner(std::move(m), features_str);
```
iomaganaris, line 106: I'm not sure if it would make sense to also measure the `JIT` compilation time as well

iomaganaris (approved):
LGTM
I just added some suggestions that could also go into another PR if you'd like

src/codegen/llvm/jit_driver.hpp
```diff
@@ -80,7 +81,12 @@ class Runner {
   public:
     Runner(std::unique_ptr<llvm::Module> m)
         : module(std::move(m)) {
-        driver->init();
+        driver->init(/*features=*/"");
+    }
+
+    Runner(std::unique_ptr<llvm::Module> m, std::string features)
+        : module(std::move(m)) {
+        driver->init(features);
```
iomaganaris, line 89:
```suggestion
    Runner(std::unique_ptr<llvm::Module> m, std::string features = "")
        : module(std::move(m)) {
        driver->init(features);
```

src/codegen/llvm/llvm_benchmark.cpp
georgemitenkov: @iomaganaris This is something to look at for sure. I think that it is not significant anyway (we just compile < 1000 instructions), but may still be worth considering. (I will add it/look at it when adding more metrics, such as min/max times, etc.)

src/nmodl/main.cpp
georgemitenkov: The default vector width is 1 in our visitor, but I see what you mean! That is a good point, however, I think that for benchmarking we want a greater control of parameters (e.g. looking at different vector widths for a specific backend: LLVM can split long vectors into shorter ones if they are not supported by the target, and that can be interesting to analyse). 

@iomaganaris I guess we can have a (warning?) message saying that there is a vector length and target ISA mismatch?

iomaganaris: I see. I think that for having greater flexibility you may want to leave it as it is now and maybe add a `todo` to fix it later :+1:

src/codegen/llvm/llvm_benchmark.cpp
iomaganaris: Yes, that's what I had in mind. In case we had to measure the JIT compilation time only measuring the construction of `runner::Runner` (and more explicitly the `JITDriver::init()`) would be enough, right?

src/nmodl/main.cpp
pramodk: LGTM

src/codegen/llvm/llvm_benchmark.cpp
pramodk: I plan to add my "favourite table" and thats where we could log this.

pramodk (approved):
LGTM.
thanks @iomaganaris for detailed review!

@georgemitenkov : feel free to squash & merge this as well.

PR #590, opened 2021-04-16
Removed allocations in loops for LLVM codegen

georgemitenkov:
When generating kernel code, local variables to the loop
body were allocated inside:
```
FOR ( ... ) {
  INTEGER node_id
  ...
}
```
For large sizes of instance structs, this was causing
stack overflows. This PR pulls local variable allocations
outside the loop bodies:
```
INTEGER node_id
FOR ( ... ) {
  ...
}
INTEGER epilogue_node_id
FOR ( ... ) {
  ...
}
```

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -562,15 +562,16 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
     function_statements.push_back(
         create_local_variable_statement(induction_variables, INTEGER_TYPE));
 
+    /// create vectors of local variables that would be used in compute part
+    std::vector<std::string> int_variables{"node_id"};
+    std::vector<std::string> double_variables{"v"};
+
     /// create now main compute part : for loop over channel instances
 
     /// loop body : initialization + solve blocks
     ast::StatementVector loop_def_statements;
     ast::StatementVector loop_index_statements;
     ast::StatementVector loop_body_statements;
-
-    std::vector<std::string> int_variables{"node_id"};
-    std::vector<std::string> double_variables{"v"};
     {
         /// access node index and corresponding voltage
         loop_index_statements.push_back(
@@ -597,6 +598,7 @@ void CodegenLLVMHelperVisitor::visit_nrn_state_block(ast::NrnStateBlock& node) {
         /// add breakpoint block if no current
         if (info.currents.empty() && info.breakpoint_node != nullptr) {
             auto block = info.breakpoint_node->get_statement_block();
+            // \todo this automatically adds `SOLVE states METHOD ...`
             append_statements_from_block(loop_body_statements, block);
         }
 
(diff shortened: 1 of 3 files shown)


pramodk:
@georgemitenkov : This is good change. But just curious - our expectation was that this would be avoided/optimised by some LLVM pass. Are we running corresponding LLVM passes on the generated module?

georgemitenkov:
> @georgemitenkov : This is good change. But just curious - our expectation was that this would be avoided/optimised by some LLVM pass. Are we running corresponding LLVM passes on the generated module?

That would be optimised by `-mem2reg` (or `-sroa` which is now overtaking `-mem2reg`). The problem was that we were executing unoptimised kernel (which should still give same results)

pramodk:
bq. The problem was that we were executing unoptimised kernel (which should still give same results)

Oh ok. Understood now. We have changed our code generation and its fine. But I wonder what happens in below case:

```
 DERIVATIVE states {
         ....

         {
              LOCAL x, y
              x = ...
              y = ...
              alpha = x + y
          }     
}
```

i.e. user can declare `LOCAL` variables inside any NMODL block. With this, will we have same problem? (I didn't look at `.ll` file to find out if it's using `alloca`). 

But what I am trying to ask is - even with unoptimised code, should there be some basic LLVM passes should be run?  (if yes then we can do it later and in separate PR is perfectly fine)

georgemitenkov:
`LOCAL` variables are fine in any NMODL block unless it is not a loop (i.e. repeated allocation): `FOR` (Not real NMODL node => more or less handled) and `WHILE`. What LLVM does, it promotes all `alloca`s to the start of the function, no matter what scoping was used. In your example, `x` and `y` would have their `alloca`s defined just after the function signature.

> should there be some basic LLVM passes should be run?

It should be working even without `mem2reg`, etc, and I am not sure that LLVM has a special pass for that (I would say that Clang handles `alloca` creation) - but I will check that. The issue raised here is definitely important and needs handling.

I guess options are:
- Gather information from NMODL to create `alloca`s from the beginning (may require renaming)
- Use LLVM pass (will check if there is one)
- Transform LLVM IR manually to move allocations from loop bodies, etc.

pramodk:
thanks @georgemitenkov for detailed explanation! We can discuss this next week in detail and approaches you mentioned make sense

> LOCAL variables are fine in any NMODL block unless it is a loop (i.e. repeated allocation): 

Just to clarify my example : the `LOCAL x, y` are under `DERIVATIVE` block. That means, implicitly whole block is added inside the for loop i.e.

```cpp
void nrn_state(....) {
     ....
     for (id = 0; id < ........) {
          ...
         {
              double x, y
              x = ...
              y = ...
              alpha = x + y
          }   
     }

}

```

pramodk (approved):
LGTM. This is good to merge!

georgemitenkov:
> thanks @georgemitenkov for detailed explanation! We can discuss this next week in detail and approaches you mentioned make sense

Sure!

> Just to clarify my example : the LOCAL x, y are under DERIVATIVE block. That means, implicitly whole block is added inside the for loop i.e.

Right! But this is the case where we can't blindly put this into for loop, but also have to pull some things out (Like `mech->dt` or as it turned out - allocations)

PR #593, opened 2021-04-18
Minor benchmarking improvement

pramodk:
- allocate instance data only once
 - store memory size with instance data
 - print memory size while running benchmarking kernel

### DIFF
--- a/src/codegen/llvm/llvm_benchmark.cpp
+++ b/src/codegen/llvm/llvm_benchmark.cpp
@@ -107,14 +107,15 @@ void LLVMBenchmark::run_benchmark(codegen::CodegenLLVMVisitor& visitor,
 
     // Benchmark every kernel.
     for (const auto& kernel_name: kernel_names) {
-        *log_stream << "Benchmarking kernel '" << kernel_name << "'\n";
+        // Initialise the data.
+        auto instance_data = codegen_data.create_data(instance_size, /*seed=*/1);
+
+        double size_mbs = instance_data.num_bytes / (1024.0 * 1024.0);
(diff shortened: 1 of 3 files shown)


src/codegen/llvm/llvm_benchmark.cpp
```diff
@@ -107,14 +107,15 @@ void LLVMBenchmark::run_benchmark(codegen::CodegenLLVMVisitor& visitor,
 
     // Benchmark every kernel.
     for (const auto& kernel_name: kernel_names) {
-        *log_stream << "Benchmarking kernel '" << kernel_name << "'\n";
+        // Initialise the data.
+        auto instance_data = codegen_data.create_data(instance_size, /*seed=*/1);
```
georgemitenkov, line 111: Don't we want to allocate for every experiment to avoid caching for further iterations?

georgemitenkov (approved):
Posted a question as a comment. LGTM!

src/codegen/llvm/llvm_benchmark.cpp
pramodk: Two aspects:
* In practice, we perform benchmark in two ways :   1) allocate small enough data so that it fits into particular cache level 2) allocate large enough data to avoid it fitting in particular cache level. For (2), we allocate dataset much greater than last level cache (LLC) so that we when we run multiple iterations we know that the data is streaming from DRAM.
* By the way - even if we allocate data on every iteration, the `initialisation of instance`  is  brining that into cache (if its small). So reallocation on each iteration doesn't really solve the caching aspect.

PR #595, opened 2021-04-19
Bug fix in codegen helper: delete LOCAL statement

### LINKED ISSUE #594: Error: incorrect IR has been generated : "Stored value type does not match pointer operand type"
A simple mod file:

```python
NEURON {
    SUFFIX hh
    NONSPECIFIC_CURRENT il
    RANGE minf, mtau, gl, el
}

STATE {
    m
}

ASSIGNED {
    v (mV)
    minf
    mtau (ms)
}

BREAKPOINT {
    SOLVE states METHOD cnexp
    il = gl*(v - el)
}

DERIVATIVE states {
     LOCAL x
     x = 12
     :m' =  (minf-m)/mtau
}
```

and IR generation fails:

```bash
$ ./bin/nmodl ../test.mod --output tmp llvm --ir --vector-width 8

    INTEGER id
    INTEGER node_id
    DOUBLE v
    for(id = 0; id<mech->node_count-7; id = id+8) {
        DOUBLE x
        v = mech->voltage[node_id]
        LOCAL x
        x = 12
    }
    INTEGER epilogue_node_id
    DOUBLE epilogue_v
    for(; id<mech->node_count; id = id+1) {
        DOUBLE x
        epilogue_v = mech->voltage[epilogue_node_id]
        LOCAL x
        x = 12
    }
}
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error: incorrect IR has been generated!
Stored value type does not match pointer operand type!
  store double 1.200000e+01, <8 x double>* %x, align 8
 <8 x double>
Aborted
```

pramodk:
- LOCAL statement was not deletected correctly
- TODO: IR generation still fails with

@georgemitenkov : I fixed LOCAL node deletion but IR generation still fails with below error:

```bash
$ ./bin/nmodl ../test.mod --output tmp llvm --ir --vector-width 8 passes
...
    INTEGER id
    INTEGER node_id
    DOUBLE v
    for(id = 0; id<mech->node_count-7; id = id+8) {
        DOUBLE yyy
        v = mech->voltage[node_id]
        yyy = 12
    }
    INTEGER epilogue_node_id
    DOUBLE epilogue_v
    for(; id<mech->node_count; id = id+1) {
        DOUBLE yyy
        epilogue_node_id = mech->node_index[id]
        epilogue_v = mech->voltage[epilogue_node_id]
        yyy = 12
    }
}
terminate called after throwing an instance of 'std::runtime_error'
  what():  Error: incorrect IR has been generated!
Stored value type does not match pointer operand type!
  store double 1.200000e+01, <8 x double>* %yyy, align 8
 <8 x double>
Aborted
```

If you have quick fix, feel free to push to this branch.

fixes #594

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -473,12 +473,13 @@ void CodegenLLVMHelperVisitor::convert_local_statement(ast::StatementBlock& node
         }
 
         /// remove local list statement now
-        const auto& statements = node.get_statements();
-        node.erase_statement(statements.begin());
+        std::unordered_set<nmodl::ast::Statement*> to_delete({local_statement.get()});
+        node.erase_statement(to_delete);
 
         /// create new codegen variable statement and insert at the beginning of the block
         auto type = new ast::CodegenVarType(FLOAT_TYPE);
         auto statement = std::make_shared<ast::CodegenVarListStatement>(type, variables);
+        const auto& statements = node.get_statements();
         node.insert_statement(statements.begin(), statement);
     }
 }


georgemitenkov:
@pramodk I think that we might want to avoid local allocations for now in kernels? (By that I mean doing a separate PR that fixes this bug and allocations all together - I think these issues are quite linked). My reasoning is the following:

1. Local allocations should be pulled out of the loops (as mentioned with the benchmark bug: I guess we can discuss this in more detail tomorrow during the sync)
2. Local allocations must also be renamed in the remainder of the loop: here we have `yyy` that is vector in the first loop and scalar in the second. Since the name conflicts, the second (scalar)`alloca` has a name `yyy1` due to LLVM’ renaming. This means that storing to `yyy` is now storing to the vector and is wrong!

This can be avoided by:
- Finding all local variables on the loop
- Renaming remainder loop ones if there is an epilogue loop
- Pulling the allocations outside

My personal view is that it’s better done on NMODL level, since we just move `CodegenVarList` blocks and do renaming on them too. What do you think?

Alternatively, we can just add this to this PR as you suggested😊

pramodk:
> By that I mean doing a separate PR that fixes this bug and allocations all together - I think these issues are quite linked

Now I remember discussing this. Agree - we can split these issues.

We can discuss this tomorrow but my question would be - how this is handled in LLVM? I don't think LLVM pass pulls this outside? i.e. there must some pass to get these local variables optimised out?

PR #597, opened 2021-04-19
Symbol table now show functions and procedures

alkino:
* Create new variable for count of calls so that callables don't increment read count.
* Show internal and external callables.


pramodk:
@alkino : did you check failing CI test?

alkino:
Yes I want to talk about that with you.

I think there is an other small bug here.

PR #598, opened 2021-04-19
Fixed bug with replacing external function when in-lining

iomaganaris:
An example, to test this is invlfire.mod.
in the function net_receive_IntervalFire, there is the call of artcell_net_move(firetime())


pramodk (approved):
LGTM!

alkino (approved):
The code generated for invlfire is now correct

PR #600, opened 2021-04-20
Fix computation of indexes

alkino:
* This bug has been introduced by 520f60e4bcd5e
* By simplifying code we forget to expand resolution of indexes


pramodk (approved):
LGTM

@alkino : this fixes your test?

iomaganaris (approved):

PR #601, opened 2021-04-20
Fix use of artcell_net_move to match the signature of CoreNeuron and nrn

alkino:
* Catched by computing netstim.mod


pramodk (approved):
if you & Ioannis has compared with mod2c code then its good to go!

PR #603, opened 2021-04-20
LLVM 13 compatibility and fixing void* type

georgemitenkov:
This PR fixes:

1. cmake LLVM libraries to be compatible with
LLVM 13 (trunk).

2. `void*` type is illegal in LLVM, and hence is
replaced with `i8*`.


pramodk (approved):
LGTM

PR #609, opened 2021-04-22
Using shared libraries in LLVM JIT

georgemitenkov:
This PR adds support for using shared libraries in the JIT runner.
The current implementation uses the benchmark as the entry-point:
```
$ bin/nmodl file.mod llvm --ir --vector-width 4 --veclib SVML benchmark --run --libs <...>.dylib <...>.dylib <...>.dylib
```

Also, the benchmark also takes the vector library into account.

### DIFF
--- a/src/codegen/llvm/jit_driver.cpp
+++ b/src/codegen/llvm/jit_driver.cpp
@@ -111,7 +111,7 @@ std::unique_ptr<llvm::TargetMachine> JITDriver::create_target(
 }
 
 void JITDriver::set_triple_and_data_layout(const std::string& features) {
-    // Get the deafult target triple for the host.
+    // Get the default target triple for the host.
     auto target_triple = llvm::sys::getDefaultTargetTriple();
     std::string error_msg;
     auto* target = llvm::TargetRegistry::lookupTarget(target_triple, error_msg);
--- a/src/codegen/llvm/jit_driver.cpp
+++ b/src/codegen/llvm/jit_driver.cpp
@@ -11,9 +11,11 @@
 #include "llvm/ExecutionEngine/JITEventListener.h"
 #include "llvm/ExecutionEngine/ObjectCache.h"
 #include "llvm/ExecutionEngine/Orc/CompileUtils.h"
+#include "llvm/ExecutionEngine/Orc/Core.h"
 #include "llvm/ExecutionEngine/Orc/IRCompileLayer.h"
 #include "llvm/ExecutionEngine/Orc/JITTargetMachineBuilder.h"
 #include "llvm/ExecutionEngine/Orc/LLJIT.h"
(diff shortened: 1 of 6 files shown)


src/codegen/llvm/jit_driver.cpp
```diff
@@ -22,27 +24,57 @@
...
+        auto layer = std::make_unique<llvm::orc::RTDyldObjectLinkingLayer>(session, []() {
+            return std::make_unique<llvm::SectionMemoryManager>();
+        });
+        for (const auto& lib_path: lib_paths) {
+            // For every library path, create a corresponding memory buffer.
+            auto memory_buffer = llvm::MemoryBuffer::getFile(lib_path);
+            if (!memory_buffer) {
+                logger->warn("Unable to create memory buffer for " + lib_path);
+                continue;
+            }
```
pramodk, line 48: If this fails, i.e. `memory_buffer == null`, then better to fail? the reason is that when we run tests / benchmarks under CI, it's better to have failures than things running with warning that we don't notice?

src/codegen/llvm/jit_driver.cpp
```diff
@@ -22,27 +24,57 @@
...
+            // Create a new JIT library instance for this session and resolve symbols.
+            auto& jd = session.createBareJITDylib(std::string(lib_path));
+            auto loaded =
+                llvm::orc::DynamicLibrarySearchGenerator::Load(lib_path.data(),
+                                                               data_layout.getGlobalPrefix());
+            if (!loaded) {
+                logger->warn("Unable to load " + lib_path);
+                continue;
+            }
```
pramodk, line 57: same

src/codegen/llvm/jit_driver.cpp
```diff
@@ -80,5 +112,26 @@ std::unique_ptr<llvm::TargetMachine> JITDriver::create_target(
...
+    std::unique_ptr<llvm::TargetMachine> tm(
+        target->createTargetMachine(target_triple, cpu, features, {}, {}));
+    if (!tm)
+        throw std::runtime_error("Error: could not create the target machine\n");
+
+    // Set data layout and the target triple to the module.
+    module->setDataLayout(tm->createDataLayout());
+    module->setTargetTriple(target_triple);
+    return false;
+}
```
pramodk, line 135: return type is not used in `JITDriver::init`. What `false` mean here?

pramodk (commented):
looking good. I have two minor questions/comments.

src/codegen/llvm/jit_driver.cpp
georgemitenkov: Didn't remove it from the previous implementation. Fixed now!

pramodk (approved):
LGTM

PR #617, opened 2021-04-27
Azure wheels : build only for schedule or manual triggers

pramodk:
* User defined variables can not be referred in compile
    time evaluation using `${{ }}` syntax
  * Also, user defined variables are not available at compile
    time evaluation
  * Stick to use of "predefined variables".
  * With this PR, all wheels are built if CI is launched via
    schedule or mnually in web UI of Azure.
  * More complex scheme based on branch name etc. can be implemented

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -5,8 +5,7 @@ variables:
   # defined in UI
   #- ReleaseWheelBuild: False
   #- NightlyWheelBuild: False
-  buildWheel: ${{ or(in(variables['Build.Reason'], 'Schedule'),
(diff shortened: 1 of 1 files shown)


pramodk:
PR CI triggered has: https://dev.azure.com/pramodskumbhar/nmodl/_build/results?buildId=2812&view=results

image

Manually triggered CI on branch via Azure Web UI has: https://dev.azure.com/pramodskumbhar/nmodl/_build/results?buildId=2814&view=results

image

alkino (approved):

PR #620, opened 2021-04-30
Refactoring of runners' infrastructure and dumping object files

### LINKED ISSUE #611: Dumping LLVM JIT module to assembly
Currently, we not see the assembly generated from LLVM IR module when passing it to JIT. We need to find a way (see #592) to verify that the instructions are produced for intended platforms.

georgemitenkov:
This PR addresses the following:

1. Dumping object files in JIT.

A functionality to dump (enabled by default) the generated from LLVM IR file binar to `.o` has been added to benchmarking. Now, in addition to logs, a `v<vector_width>_<mod_filename>.o` is generated. The reasons it is an object file and not an assembly (hence not included in logs) are the following:

- LLVM does not have library functions that take the object and turn back into assembly, but rather `object -> file -> assembly` path. It also has a `llvm-objdump` tool, but it is intended as a command-line utility and does not have a well-defined API.
- Writing custom functions to produce a readable assembly is not a priority. Also, mimicking `objdump` functionality would be difficult.
- Both `objdump` and `llvm-objdump` can be used to isnpect the `.o` file manually.

2. Refactoring of `Runner` class.
In addition to the support of dumping the binary, `Runner` and `JITDriver` classes were refactored to have a nicer OOP-style.

fixes #611

### DIFF
--- a/src/codegen/llvm/jit_driver.cpp
+++ b/src/codegen/llvm/jit_driver.cpp
@@ -15,6 +15,7 @@
 #include "llvm/ExecutionEngine/Orc/IRCompileLayer.h"
 #include "llvm/ExecutionEngine/Orc/JITTargetMachineBuilder.h"
 #include "llvm/ExecutionEngine/Orc/LLJIT.h"
+#include "llvm/ExecutionEngine/Orc/ObjectTransformLayer.h"
 #include "llvm/ExecutionEngine/Orc/RTDyldObjectLinkingLayer.h"
 #include "llvm/ExecutionEngine/SectionMemoryManager.h"
 #include "llvm/Support/Host.h"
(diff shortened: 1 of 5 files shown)


pramodk (approved):
Nice! LGTM!

Quickly tested locally and looks good!

By the way - unrelated to PR - one thing I realised while testing is that we forgot to add/update README about the all options we are adding in LLVM branch. Every time I test something, I try to grab from history esp now new options about vec-replace, svml loading , benchmarking etc etc. 😅  I will try to help with documenting this with examples...

georgemitenkov:
That’s a good point! I will make an issue for the README. I was postponing this to code refactoring/more comments (some places are not fully consistent and well documented (for people outside LLVM)). But we will definitely do this☺️

PR #621, opened 2021-04-30
Extend GitLab CI

olupton:
Trigger CVF pipeline on changes to NMODL.

Also:
 - Use `gitlab-pipelines` defaults for when to run jobs.
 - Remove `resource_group` keywords and add more `needs` to allow more parallelism.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,6 +20,12 @@ setup virtualenv:
   stage: .pre
   before_script:
     - export
+  variables:
+    # This might pull in more than needed, but it should be enough to allow the
+    # two build jobs to run in parallel safely. Otherwise they will race each
+    # other to lazily update the submodules during the CMake phase, which can
+    # cause errors.
+    GIT_SUBMODULE_STRATEGY: recursive
 
 trigger cvf:
   stage: .pre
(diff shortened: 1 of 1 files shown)


pramodk (approved):
LGTM

.gitlab-ci.yml
```diff
@@ -19,10 +21,21 @@ setup virtualenv:
...
+trigger cvf:
+  stage: .pre
+  variables:
+    # Tell the CVF pipeline to use the current ref of NMODL, this works because
+    # the CVF CI uses the `gitlab-pipelines` helper components.
+    SPACK_PACKAGE_REF_nmodl: "commit='${CI_COMMIT_SHA}'"
+  trigger:
+    project: hpc/cvf
+    # Make the NMODL CI status depend on the CVF CI status
+    strategy: depend
```
pramodk, line 33: can we avoid trigger for specific branch like `llvm`?

alkino (approved):

olupton:
I checked that [the pipeline triggered when I pushed a change with the target branch changed to `llvm`](https://bbpgitlab.epfl.ch/hpc/nmodl/-/pipelines/5100) did not try and run the CVF child pipeline.

.gitlab-ci.yml
olupton: With ecb3b68c4896e50082ad18e9f34935d82da10c2c I added a rule that stops the CVF part being run on PRs to the `llvm` branch. We already only run the CI on PR events, so that should be enough.

olupton:
@pramodk if you're happy with the `llvm` change I think this is ready to go.

pramodk (approved):
LGTM

PR #623, opened 2021-05-02
Optimisation levels for benchmarking

### LINKED ISSUE #616: Question : should we add `--opt-level X` flag for benchmarking purpose?
This is more as a question - for benchmarking via JIT, should we add CLI option to run default optimisation passes like `-O0, -O1, -O2, or -O3`? Seems like [llvm::PassManagerBuilder](https://llvm.org/doxygen/classllvm_1_1PassManagerBuilder.html) has most of the infrastructure to easily add default optimisation passes and it's used in tools like opt, lli (see usage of `populateFunctionPassManager` in [opt.cpp](https://github.com/llvm/llvm-project/blob/6fca189532511da1b48e8c0d9aad8ff2edca382d/llvm/tools/opt/opt.cpp#L416)).

@georgemitenkov : I can add this into NMODL and it might be useful for benchmarking? Before that, I want to check if you saw any reason not to have such option.

georgemitenkov:
This PR adds two flags to the benchmarking pipeline:

1. `--opt-level-ir`: This flag is used to run `-On` passes on the generated LLVM IR module.

2. `--opt-level-codegen`: This flag is used for setting optimisation level for machine code generation inside the JIT target machine.

Example:
```bash
$ ./nmodl file.mod \
     llvm --ir --vector-width 1 \
     benchmark --run --instance-size 10000000 --repeat 20 --opt-level-ir 2 --opt-level-codegen 2
```

*Note: `--opt` flag in LLVM visitor has not been removed as we may want to experiment with some specific passes, etc.*

Also, some benchmarking and JIT driver functions have been refactored from class members and static utilities. More documentation comments were added.

fixes #616


pramodk:
@georgemitenkov : looks good to me but I have two questions:

* `--opt-level-ir 2` flag should **not** be in the `benchmark` section but in `llvm` section. This way, I can use it to compare optimisations at IR level?  (wouldn't mind if you move `--opt-level-codegen 2` also in codegen if you prefer to keep them together)
* when we dump `.ll` file to output directory using `-o <dirname>`, we should dump IR after optimisation pass.

pramodk (approved):
LGTM

PR #624, opened 2021-05-03
Fix conversion to int if the double is bigger than INT_MAX

alkino:
* Due to gap between number, if number is big, std::ceil(value) == value, but it cannot fit an int
* Test before printing and use long long to take the biggest integer possible


pramodk (approved):
LGTM

PR #627, opened 2021-05-07
Add constructor

alkino:


pramodk:
Could you check CI failures ?

pramodk:
@alkino : Based on  #651, can you see what changes required here? We should get this in today.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -3315,9 +3323,20 @@ void CodegenCVisitor::print_nrn_init(bool skip_init_check) {
...
+        print_statement_block(*block.get(), false, false);
+    }
+    printer->end_block(1);
+}
+
+
 void CodegenCVisitor::print_nrn_destructor() {
     printer->add_newline(2);
-    print_global_function_common_code(BlockType::Destructor);
+    CodegenCVisitor::print_global_function_common_code(BlockType::Destructor);
```
iomaganaris, line 3339: Maybe something like
```
diff --git a/src/codegen/codegen_ispc_visitor.cpp b/src/codegen/codegen_ispc_visitor.cpp
index b2822f10..7646001f 100644
--- a/src/codegen/codegen_ispc_visitor.cpp
+++ b/src/codegen/codegen_ispc_visitor.cpp
@@ -357,6 +357,9 @@ void CodegenIspcVisitor::print_procedure(const ast::ProcedureBlock& node) {
 
 
 void CodegenIspcVisitor::print_global_function_common_code(BlockType type) {
+    if (wrapper_codegen) {
+        return CodegenCVisitor::print_global_function_common_code(type);
+    }
     std::string method = compute_method_name(type);
 
     auto params = get_global_function_parms(ptr_type_qualifier());
```
makes more sense? Otherwise if we need to call in another point `print_global_function_common_code` again we would have to remember to call the one from `CodegenCVisitor`

alkino: ok

alkino: done

pramodk (approved):
LGTM

Issue #612, opened 2021-04-23
Adding debug symbols and metadata for profiling kernels


georgemitenkov:
Issue: #592

We want to have some debug symbols that would allow us to profile the LLVM IR kernels. Consider adding these.

pramodk:
Just a minor note: for #548, we need to add comment/annotation to the loop body:

```asm
$ → cat hh.1.s
....
# LLVM-MCA-BEGIN Body
LBB0_2:                                 ## %for.body
                                        ## =>This Inner Loop Header: Depth=1
	movq	%rsp, %rax
	addq	$-16, %rax
	movq	%rax, %rsp
	movq	%rsp, %rax
	addq	$-16, %rax
	movq	%rax, %rsp
	movq	-16(%rbp), %rax
	movslq	-4(%rbp), %rcx
	movq	(%rax), %rdx
	movq	8(%rax), %rsi
	movq	16(%rax), %rax
	vmovsd	(%rdx,%rcx,8), %xmm0            ## xmm0 = mem[0],zero
	vsubsd	(%rax,%rcx,8), %xmm0, %xmm0
	vdivsd	(%rsi,%rcx,8), %xmm0, %xmm0
	vmovsd	%xmm0, (%rax,%rcx,8)
	incl	-4(%rbp)
	movq	-16(%rbp), %rax
	movl	-4(%rbp), %ecx
	cmpl	92(%rax), %ecx
	jl	LBB0_2
# LLVM-MCA-END
```

In case this is minor/easy to do with few lines. Otherwise don't worry for now.

georgemitenkov:
Do you mean annotation that it is a loop? In this case, I have actually made a ticket for that yesterday: #607. (Also: #606). In LLVM there is `llvm.loop` metadata that Clang emits, so that on I we know that the basic blocks are from the loop 👍 

Or you mean debug symbols/metadata in general for profiling?

pramodk:
May be I shouldn't have mixed the topic in this ticket. Let's get `debug symbols/metadata in general for profiling` for JIT kernels, as we discussed. I will add separate ticket regarding my comment above (this is very minor/low priority thing)

PR #628, opened 2021-05-07
Adding function debug information

### LINKED ISSUE #592: Adding debug information for JIT compiled kernel?
(This is low priority ticket but creating this so that it's on our list)

For benchmarking purpose I was running llvm backend benchmarks under VTune with the intention of finding `nrn_state` hotspot then can check type of assembly instructions generated and check various performance metrics. But it seems like due to lack of debug symbols (?), we don't see information about name of kernel.

For example, on Ubuntu box that we got access to:

```bash
. /opt/intel/oneapi/setvars.sh

vtune -collect hotspots -knob sampling-mode=hw -knob sampling-interval=0.5 ./bin/nmodl ../test.mod llvm --ir --vector-width 1 benchmark --run --instance-size 100000000 --repeat 15 --backend default
```

and then looking at generated result:

```
vtune-gui r000hs/
```
(see top functions with `Outside any known module`)

image

Or, with call stack information:

```
vtune -collect hotspots -call-stack-mode all -- ./bin/nmodl ../test.mod llvm --ir --vector-width 1 benchmark --run --instance-size 100000000 --repeat 30 --backend default
vtune-gui r000hs/
```
image

cc: @georgemitenkov  @iomaganaris  @castigli

georgemitenkov:
This PR introduces debug support to LLVM code generation pipeline. Currently, only basic support was added:

**What is supported**

- Debug information about functions (name)
- Debug information about module

**What has been changed and added**

1. A new class `DebugBuilder` was created. It is used as a wrapper around LLVM's `DIBuilder` and holds important information such as `LLVMContext`, debug file and compile unit. It also wraps `DIBuilder`'s functionality into a more suitable API.

2. A temporary `Location` struct has been added. It encapsulates the location of the source AST construct and reflects `ModToken` on LLVM code generation level. It is only used if the location of the source NMODL function is known.

3. LLVM visitor know takes an extra `add_debug_information` flag and handles debug information creation. Fore readability, `IRBuilder` was renamed to `ir_builder`.

4. JIT runner is now able to listen for GDB, perf (build LLVM with `-DLLVM_USE_PERF=ON`) and VTune (build LLVM with `-DLLVM_USE_INTEL_JITEVENTS=ON`) events.

**How to generate debug information**

Debug information is attached to every function, procedure or artificially created kernel (and corresponding wrappers). For benchmarking, debug information is enabled by default. Debug information is enable by default, so to turn it off use ` --disable-debug-info` flag. For example, the given NMODL
```nmodl
1   FUNCTION func(x) {
2     func = x
3   }
4
5   PROCEDURE proc() {}
```
is transformed (running `./bin/nmodl <filename>.mod llvm --ir
`) into
```llvm
define double @func(double %x1) !dbg !4 {
  ; ...
}

define i32 @proc() !dbg !6 {
  ; ...
}

!llvm.dbg.cu = !{!0}
!llvm.module.flags = !{!3}

!0 = distinct !DICompileUnit(language: DW_LANG_C, file: !1, producer: "NMODL-LLVM", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, enums: !2)
!1 = !DIFile(filename: "foo", directory: ".")
!2 = !{}
!3 = !{i32 2, !"Debug Version", i32 3}
!4 = distinct !DISubprogram(name: "func", linkageName: "func", scope: null, file: !1, line: 1, type: !5, scopeLine: 1, spFlags: DISPFlagDefinition | DISPFlagOptimized, unit: !0, retainedNodes: !2)
!5 = !DISubroutineType(types: !2)
!6 = distinct !DISubprogram(name: "proc", linkageName: "proc", scope: null, file: !1, line: 5, type: !5, scopeLine: 5, spFlags: DISPFlagDefinition | DISPFlagOptimized, unit: !0, retainedNodes: !2)
```

fixes #592 #612


CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybind11 based python bindings" OFF)
 option(NMODL_ENABLE_LEGACY_UNITS "Use original faraday, R, etc. instead of 2019 nist constants" OFF)
 option(NMODL_ENABLE_LLVM "Enable LLVM based code generation" ON)
+option(NMODL_ENABLE_JIT_EVENT_LISTENERS "Enable JITEventListener for Perf and Vtune" OFF)
```
pramodk, line 28: @georgemitenkov : see last two commits I did. Did some minor changes & CI fixes for Linux & BB5.

src/nmodl/main.cpp
```diff
@@ -319,6 +322,9 @@ int main(int argc, const char* argv[]) {
     llvm_opt->add_flag("--ir",
         llvm_ir,
         "Generate LLVM IR ({})"_format(llvm_ir))->ignore_case();
+    llvm_opt->add_flag("--add-debug-info",
+                       add_debug_information,
+                       "Add debug information ({})"_format(add_debug_information))->ignore_case();
```
pramodk, line 327: Wondering if we should always add debug info (or have it ON)

src/codegen/llvm/llvm_debug_builder.hpp
```diff
@@ -0,0 +1,70 @@
...
+/// some may not have it as they were artificially introduced (e.g.
+/// CodegenForStatement). This simple wrapper suffices for now, but in future
+/// we may want to handle this properly.
+struct Location {
+    /// Line in the file.
+    int line;
+
+    /// Column in the file.
+    int column;
+};
```
pramodk, line 31: to understand - once we add proper `Token` to the CodegenFunction nodes then this won't be needed?

pramodk (approved):
Quickly went through this and overall looks ok to me.

(See my changes in 651b257 and 3af2c1f to make it work on BB5 & linux)

src/codegen/llvm/jit_driver.cpp
```diff
@@ -158,8 +158,10 @@ void JITDriver::init(std::string features,
     // with listeners on (e.g. -DLLVM_USE_PERF=ON).
     if (benchmark_info) {
         gdb_event_listener = llvm::JITEventListener::createGDBRegistrationListener();
+#if defined(NMODL_HAVE_JIT_EVENT_LISTENERS)
```
georgemitenkov, line 161: It is actually safe to not use #if since if flag for listeners not set in LLVM build, these will be `nullptr`. But it is better not to rely on that, I agree!

src/nmodl/main.cpp
georgemitenkov: We can, yes, then a flag `--disable-debug-info` can be used to turn it off instead. What do you think?

src/codegen/llvm/jit_driver.cpp
pramodk: But, note that if `createPerfJITEventListener()` and `createIntelJITEventListener()` are **used** but we are not linking to libraries or don't have them then we get undefined reference error. This is why added it.

src/nmodl/main.cpp
pramodk: 👍 : the reason I mentioned to make it ON is that during development we need this all the time. So we don't need to add one more flag to CLI. (at least for now)

src/codegen/llvm/llvm_debug_builder.hpp
georgemitenkov: I am not sure here. So far I pass proper tokens from `ProcedureBlock` and `FunctionBlock`, so that the lines are known. But if the node is created artificially (e.g. in the kernel), there is no line/column information.

My idea for that class was to have it as a wrapper around location, so that `Token` is used on the lexer/parser level, and for code generation a `Location` wrapper is used (So we separate these two and can add some more functionality to location, such as `loc.getNextLocation()`, etc. Hence, knowing the location of the function start, we can know roughly locations of the inner AST nodes.) But that's not very defined to me yet - so I made this struct for now:)

src/nmodl/main.cpp
georgemitenkov: Agree! Will change that!

src/codegen/llvm/jit_driver.cpp
georgemitenkov: I see, that's right!

src/codegen/llvm/llvm_debug_builder.hpp
pramodk: Ok understood. Having bit separation with a wrapper not a bad idea. 👍

pramodk (approved):

PR #634, opened 2021-05-11
IR builder redesign for LLVM IR code generation pipeline

### LINKED ISSUE #538: LLVM Visitor clean-up
It's been a while and the LLVM visitor requires a clean-up. To address after vectorisation.

georgemitenkov:
This PR is intended to improve the code structure for the LLVM code generation pipeline. It adds the following:

**New IR builder**

Before, LLVM visitor just simply used `llvm::IRBuilder<>` class to generate instructions. Recently, this (as well as adding the functionality to the visitor on the go) had led to several challenges:

- There was no common builder utilities, leading to code duplication.
- Visitor class had to keep track of only builder-related information, such as the current function, the value stack, whether the code is being for the kernel, etc.
- It became hard to introduce more features nicely.

Hence, a special `IRBuilder` class is now used. This class is a wrapper around `llvm::IRBuilder<>`, that keeps track of certain IR generation specific fields (that are unrelated to the visitor!), and it:

1. Defines an API that LLVM visitor can use to generate LLVM IR.
2. Combines the IR generation functionality in a nice and understandable way for other contributors.

Also, this IR builder has been designed to be nearly fully-independent from NMODL AST nodes. this allows it to be more generic and to be more extensible.

**Visitor clean-up**

There are a number of changes in LLVM visitor, mostly due to integration of the new builder class. In addition to this, the functions have been refactored and renamed to better reflect the functionality and provide encapsulation.

**Scatter preparation**

The functionality of the generating code for `CodegenInstanceVar` node have been extended (see `read_from_or_write_to_instance(...)`), so that now it also takes into account the case when we store to it. While no scatter support is there yet, this allows a nice and easy scatter and predication integration into the visitor.

fixes #538


georgemitenkov:
@pramodk There are a lot of changes in this PR :) but I am unsure how to split it nicely? My guess would be to land IR builder separately, and then the remaining refactoring?

Posting the full change here for now.

pramodk:
> * Before, LLVM visitor just simply used llvm::IRBuilder<> class to generate instructions. Recently, this (as well as adding the functionality to the visitor on the go) had led to several challenges
> * There are a number of changes in LLVM visitor, mostly due to integration of the new builder class. In addition to this, the functions have been refactored and renamed to better reflect the functionality and provide encapsulation.

@georgemitenkov : this is great! 👌 

> @pramodk There are a lot of changes in this PR :) but I am unsure how to split it nicely? My guess would be to land IR builder separately, and then the remaining refactoring?

I wouldn't mind merging this as one PR. Not that large and these changes together make sense.

I will review it tomorrow morning and it can be merged.

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -62,458 +60,352 @@ static bool can_vectorise(const ast::CodegenForStatement& statement, symtab::Sym
...
-    const auto& instance_name = node.get_instance_var()->get_node_name();
-    const auto& member_name = member_node->get_node_name();
-
-    if (!instance_var_helper.is_an_instance_variable(member_name))
-        throw std::runtime_error("Error: " + member_name + " is not a member of the instance!");
+void CodegenLLVMVisitor::create_external_function_call(const std::string& name,
+                                                       const ast::ExpressionVector& arguments) {
+    if (name == "printf") {
+        create_printf_call(arguments);
+        return;
```
pramodk, line 72: just to clarify : `printf` works only for serial execution, right?
(don't need to fix it here, it's used for debugging purpose only)

pramodk (approved):
LGTM

Not much to add - separation of builder is very nice! I didn't cross check every routine but as this is reorganisation/refactoring PR, overall looks good to me.

@georgemitenkov : go ahead with merge if there is nothing else pending.

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: Yes, `printf` does not check that input types are coherent with the provided string (e.g. "%d", etc.). So it works well for scalar code, but if vectorized it prints some random things - I think vector is treated as an int or something similar.

It would be nice to fix it however, and should be rather easy actually, just replicate printf calls for every element.

Example:
```c++
printf("%d", i) // i is vector
```
would yield
```c++
<i1, i2, i3, i4>  // < can be use or some other bracket-like symbol
```

The caveat here is that the format string needs to be broken apart, for additional printf calls, etc. (I don't think you can printf with llvm vector)

georgemitenkov: Actually, what we can do is to define an extern C interface that prints llvm vectors: and replace/insert these printing function if needed:

Something like:
```c++
extern "C" void print_some_type(SomeType<int8_t> *T) {
  std::cout << "<";
  // Deal with type here
  std::cout << ">";
}
```

pramodk: Yes, thats one option. There is something similar has been done in ISPCP https://github.com/ispc/ispc/. I will try to find out what's been done.

Anyway, this is future enhancement and we don't have to worry about it now. I will mark it resolve.

PR #642, opened 2021-05-13
Fixed initialisation of `CodegenAtomicStatement`

georgemitenkov:
Before, `CodegenAtomicStatement` was initialised with lhs as `""{}[{}]"_format(ion_varname, index_varname)"`. Hence, instead of an `IndexedName`, a `Name` node was created. Moreover, the lhs variable was not recognised as `CodegenInstanceVariable`.

This PR fixes this issue, making the `CodegenAtomicStatement` consistent with other variables of `CodegenInstance` node.


pramodk (approved):
LGTM

PR #646, opened 2021-05-14
Print last line in parser error

ohm314:
Small improvement to parser error handling. Assuming we have a little typo on line 89 of netstim.mod, turn this error reporting

```
bin/nmodl netstim.mod
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: NMODL Parser Error : syntax error, unexpected . [Location : 89.50]
```

into this

```
bin/nmodl netstim.mod
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: NMODL Parser Error : syntax error, unexpected . [Location : 89.50]
    event = start + invl(interval) - interval*(f1. - noise)
-------------------------------------------------^
```

- We Store the current line being lexed in the NmodlLexer instance.
- We keep pointers to the current scanner and parser active in the driver to access them from the `NmodlDriver::parse_error` callback.
- in `parse_error` we print the last line from scanner along with the location pointer.

## Caveats

This seems to work quite nicely, but the printing is only as precise as the `location` parameter reports it from the parser. Additionally, `parse_error` is called in some cases directly from the driver. I'm not quite sure if the output would be correct - maybe there should be two parse_error callbacks?

### DIFF
--- a/src/lexer/nmodl_lexer.hpp
+++ b/src/lexer/nmodl_lexer.hpp
@@ -80,6 +80,8 @@ class NmodlLexer: public NmodlFlexLexer {
      */
     int lexical_context = 0;
 
+    std::string last_line;
+
   public:
     /// location of the parsed token
     location loc;
@@ -143,6 +145,9 @@ class NmodlLexer: public NmodlFlexLexer {
 
     /// Return last scanned unit as ast::String
     ast::String* get_unit();
+
+    /// Return current line as string
+    std::string get_curr_line();
 
     /// Enable debug output (via yyout) if compiled into the scanner.
     void set_debug(bool b);
--- a/src/lexer/nmodl_lexer.hpp
+++ b/src/lexer/nmodl_lexer.hpp
@@ -80,7 +80,7 @@ class NmodlLexer: public NmodlFlexLexer {
      */
     int lexical_context = 0;
 
-    std::string last_line;
+    std::string cur_line;
 
   public:
     /// location of the parsed token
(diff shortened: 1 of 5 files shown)


src/lexer/nmodl_lexer.hpp
```diff
@@ -80,6 +80,8 @@ class NmodlLexer: public NmodlFlexLexer {
      */
     int lexical_context = 0;
 
+    std::string last_line;
```
pramodk, line 83: `pure nitpick` : if you named function as get current line, variable could be cur_line rather than last line? :)

src/parser/nmodl_driver.cpp
```diff
@@ -21,12 +21,18 @@ NmodlDriver::NmodlDriver(bool strace, bool ptrace)
 
 /// parse nmodl file provided as istream
 std::shared_ptr<ast::Program> NmodlDriver::parse_stream(std::istream& in) {
-    NmodlLexer scanner(*this, &in);
-    NmodlParser parser(scanner, *this);
+    scanner = new NmodlLexer(*this, &in);
+    parser = new NmodlParser(*scanner, *this);
```
pramodk, line 25: could these be smart pointers?

pramodk (approved):
LGTM - I just left comment about raw pointer -> smart pointer. That could be changed if there are no caveats?

src/lexer/nmodl_lexer.hpp
ohm314: yeah.. you're right - couldn't really decide :)

src/parser/nmodl_driver.cpp
ohm314: Probably - and you're right, they should. Unfortunately converting to `unique_ptr` causes some test cases to segfault again...

ohm314: This was a very bad idea altogether! I have a better implementation now.

pramodk: 👌  (I must admit I didn't look at the code locally while reviewing PR. Just looked at the diff and thought reasonable!)

pramodk (approved):
LGTM

PR #647, opened 2021-05-16
Benchmarking code re-organisation and minor improvements

pramodk:
* Move benchmark + JIT related code from src/codegen/llvm
  to test/benchmark
* Common execution of CodegenLLVMVisitor for llvm --ir and
  benchmark option. With this, ast transformed for LLVM
  code generation is dumped to file.
* Previous object file is removed (if exist) so that output
  file name is same / deterministic
* Benchmark output is always printed to stdout via common
  logger object
* Remove unnecessary LLVMBuildInfo struct

@georgemitenkov : I wanted to do other small modifications to print benchmark performance numbers but thought better to re-organise benchmark/JIT related parts from IR generation.


test/benchmark/jit_driver.cpp
```diff
@@ -247,6 +248,13 @@ void JITDriver::init(std::string features,
 
     // Optionally, dump the binary to the object file.
     if (benchmark_info) {
+        std::string object_file = benchmark_info->filename + ".o";
+        if (utils::file_exists(object_file)) {
```
georgemitenkov, line 252: Nice! Good to automatically remove files from previous runs

test/benchmark/llvm_benchmark.hpp
```diff
@@ -10,26 +10,21 @@
...
-};
-
 /**
  * \class LLVMBenchmark
  * \brief A wrapper to execute MOD file kernels via LLVM IR backend, and
  * benchmark compile-time and runtime.
  */
 class LLVMBenchmark {
   private:
+    /// llvm visitor
```
georgemitenkov, line 25:
```suggestion
    /// LLVM visitor.
```

georgemitenkov (approved):
LGTM!

Issue #606, opened 2021-04-21
Alias analysis & attributes in LLVM


georgemitenkov:
This is a placeholder ticket to gather information on aliasing/function attributes in LLVM. Particularly we are interested in `noalias` for struct members.

georgemitenkov:
Consider a kernel:
```
VOID nrn_state_hh(INSTANCE_STRUCT *mech){
    INTEGER id
    INTEGER node_id
    DOUBLE v
    for(id = 0; id<mech->node_count; id = id+1) {
        node_id = mech->node_index[id]
        v = mech->voltage[node_id]
        mech->m[id] = (mech->minf[id]+v)/mech->mtau[id]
    }
}
```

**Experiments**

Now, consider 3 cases:
1. Generate LLVM IR via NMODL and then run `-O3`
2. Generate LLVM IR via Clang (For corresponding C++) with `-O0`, copy IR to separate file and run `-O3`
3. Generate LLVM IR via Clang (For corresponding C++) with `-O3`

Corresponding links:
- NMODL  --> LLVM IR [https://godbolt.org/z/6zv9aqq4n](https://godbolt.org/z/6zv9aqq4n)
- C++ --> LLVM IR (no previous C++ info)--> LLVM IR (O3) [https://godbolt.org/z/ae1zrse7b](https://godbolt.org/z/ae1zrse7b)
- C++ --> LLVM IR (O3) [https://godbolt.org/z/a4GKWc9av](https://godbolt.org/z/a4GKWc9av)

**Comparison and conclusions**

1. We generate IR that is identical (up to the order of evaluation of expressions (*todo: that is something for future considerations*) and overflow flags (`nsw`/`nuw`)) to IR generated by Clang with optimisations off.
2. Optimisations on IR generate the same result for cases 1 and 2, since in case 2 no C++ info is available
3. Case 3 is different, observations are:

1.  `__restrict` has no effect - Clang/LLVM is able to deduce no aliasing quite well.
2.  Clang unrolls the loop (2 iterations), hence having a case to handle odd `node_count`: this may mean there is `loop` metadata attached to branches/basic blocks => more profit from optimisations
3. Clang pulls out `load double*, double**`: due to TBAA. Fix: add `noalias` keyword as we know the kernel code does not break aliasing.

Exanple:
Changing
```llvm
define void @nrn_state_hh(%test__instance_var__type* %mech1) {
```
to
```llvm
[...]

PR #648, opened 2021-05-16
Added attributes and metadata to LLVM IR compute kernels

### LINKED ISSUE #607: Add `noalias` and `llvm.loop` metadata
As described in #606:

1. LLVM's TBAA is unable to see that there is no aliasing in the IR. Hence, we need to add `noalias` metadata to the struct argument.

2. Generating IR for loops, we do not keep any data that says that these basic blocks are part of the loop. This can be solved by adding `llvm.loop` metadata (See https://llvm.org/docs/TransformMetadata.html#transformation-metadata for more detail).

Other metadata may also be inserted (to be discussed further)

georgemitenkov:
Previously, there was no metadata and attributes associated with the instance struct pointer, compute kernels or loops. This patch fixes this.

**Instance struct attributes**

Since all pointers contained in the instance struct do not alias, we add a `noalias` (LLVM's `__restrict` alternative) attribute to it. In addition, we add `nocapture` (No capturing occurs in the function) and `readonly` (Struct pointer is not written to) attributes.

This means that some load instructions can be moved out from the loop body! Example:
```llvm
; BEFORE
for.body.lr.ph:                                   ; preds = %0
  %5 = getelementptr inbounds %avx__instance_var__type, %avx__instance_var__type* %mech1, i64 0, i32 1
  %6 = getelementptr inbounds %avx__instance_var__type, %avx__instance_var__type* %mech1, i64 0, i32 3
  %7 = getelementptr inbounds %avx__instance_var__type, %avx__instance_var__type* %mech1, i64 0, i32 2
  br label %for.body

for.body:                                         ; preds = %for.body.lr.ph, %for.body
  ; ...
  %15 = load double*, double** %5, align 8
  ; ...
  %19 = load double*, double** %6, align 8

; AFTER
for.body.lr.ph:                                   ; preds = %0
  %5 = getelementptr inbounds %avx__instance_var__type, %avx__instance_var__type* %mech1, i64 0, i32 1
  %6 = load double*, double** %5, align 8
  %7 = getelementptr inbounds %avx__instance_var__type, %avx__instance_var__type* %mech1, i64 0, i32 3
  %8 = load double*, double** %7, align 8
  %9 = getelementptr inbounds %avx__instance_var__type, %avx__instance_var__type* %mech1, i64 0, i32 2
  %10 = load double*, double** %9, align 8
  br label %for.body
```

**Loop metadata**

Also, loop metadata is added to scalar kernels, specifying that no vectorization is needed. The reason for this is because we want to benchmark truly scalar kernels, and disable LLVM's vectorization if necessary.

Note that for vector loop epilogue there is no metadata that disables vectorization.

fixes #607

### DIFF
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -528,9 +528,6 @@ void CodegenLLVMVisitor::visit_codegen_for_statement(const ast::CodegenForStatem
     llvm::BasicBlock* for_inc = llvm::BasicBlock::Create(*context, /*Name=*/"for.inc", func, next);
     llvm::BasicBlock* exit = llvm::BasicBlock::Create(*context, /*Name=*/"for.exit", func, next);
 
-    // Save the vector width.
-    int tmp_vector_width = vector_width;
-
     // Check if the kernel can be vectorised. If not, generate scalar code.
     if (!can_vectorize(node, sym_tab)) {
         logger->info("Cannot vectorise the for loop in '" + ir_builder.get_current_function_name() +
@@ -540,14 +537,12 @@ void CodegenLLVMVisitor::visit_codegen_for_statement(const ast::CodegenForStatem
         ir_builder.generate_scalar_code();
     }
 
-    // First, initialise the loop in the same basic block. This block is optional. Also, reset
-    // vector width to 1 if processing the remainder of the loop.
-    if (node.get_initialization()) {
+    // First, initialise the loop in the same basic block. This block is optional. Also, generate
+    // scalar code if processing the remainder of the loop.
+    if (node.get_initialization())
         node.get_initialization()->accept(*this);
-    } else {
-        vector_width = 1;
+    else
         ir_builder.generate_scalar_code();
-    }
 
     // Branch to condition basic block and insert condition code there.
     ir_builder.create_br_and_set_insertion_point(for_cond);
@@ -566,11 +561,9 @@ void CodegenLLVMVisitor::visit_codegen_for_statement(const ast::CodegenForStatem
     // Process increment.
     node.get_increment()->accept(*this);
 
-    // Create a branch to condition block, then generate exit code out of the loop. Restore the
-    // vector width.
+    // Create a branch to condition block, then generate exit code out of the loop.
     ir_builder.create_br(for_cond);
     ir_builder.set_insertion_point(exit);
-    vector_width = tmp_vector_width;
     ir_builder.generate_vectorized_code();
     ir_builder.start_vectorization();
 }
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -549,7 +549,8 @@ void CodegenLLVMVisitor::visit_codegen_for_statement(const ast::CodegenForStatem
 
     // Extract the condition to decide whether to branch to the loop body or loop exit.
     llvm::Value* cond = accept_and_get(node.get_condition());
-    ir_builder.create_cond_br(cond, for_body, exit);
+    llvm::BranchInst* loop_br = ir_builder.create_cond_br(cond, for_body, exit);
+    ir_builder.set_loop_metadata(loop_br);
 
     // Generate code for the loop body and create the basic block for the increment.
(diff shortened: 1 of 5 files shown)


pramodk:
> readonly (Struct pointer is not written to) attributes.

I assume when we add `readonly` to instance struct then it doesn't apply to members like `noalias`, right? (I assume yes but just to be sure...!)

Would there be a benefit if we can say which struct members are read-only/constant and which one are not?

georgemitenkov:
> I assume when we add readonly to instance struct then it doesn't apply to members like noalias, right? (I assume yes but just to be sure...!)

Yes, that only concerns  the struct pointer, not members (unlike `noalias`)

> Would there be a benefit if we can say which struct members are read-only/constant and which one are not?

I am not sure LLVM supports this - there is no way to annotate the type with attributes, only function arguments. Other option would be to unpack the instance struct and call kernels with the data pointers directly. But attributes/metadata are something to consider in future work for sure!

pramodk:
Ok, sounds good to me!

(just for reference for myself or other members who will check about [nocapture semantics](https://lists.llvm.org/pipermail/llvm-dev/2017-March/111566.html))

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -239,6 +237,26 @@ int CodegenLLVMVisitor::get_num_elements(const ast::IndexedName& node) {
     return static_cast<int>(*macro->get_value());
 }
 
+bool CodegenLLVMVisitor::is_kernel_function(const std::string& function_name) {
```
pramodk, line 240: Just as a note : do you think we should add some property / meta data for our kernel function to find kernels directly? Or, we can store their names? I am mentioning this because finding kernels by return type and arg type is not that robust. 

If you think that is worth doing in future, just add `\todo` comment.

src/codegen/llvm/llvm_ir_builder.cpp
```diff
@@ -174,6 +174,52 @@ void IRBuilder::create_intrinsic(const std::string& name,
...
+/****************************************************************************************/
+/*                                LLVM metadata utilities                               */
+/****************************************************************************************/
+
+void IRBuilder::set_loop_metadata(llvm::BranchInst* branch) {
+    llvm::LLVMContext& context = builder.getContext();
+    MetadataVector loop_metadata;
+
+    // Add nullptr for loop's metadata self-reference.
+    loop_metadata.push_back(nullptr);
```
pramodk, line 203: for my understanding as I don't know this - how this `nullptr` for loop's metadata self-reference work? if you could point me some reference / link, I will read about.

pramodk (approved):
LGTM

georgemitenkov:
@pramodk I have also added some links in the issue (regarding aliasing/attributes/metadata): see #606 and a bit of #607.

src/codegen/llvm/llvm_ir_builder.cpp
georgemitenkov: I think my comment is misleading! I add nullptr to "reserve the first place" for the self-reference. And then later in the code few lines below I call `metadata->replaceOperandWith(0, metadata)` that does the trick.

georgemitenkov: I will adjust the comment specifying what I wrote above.

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: I think so! I think I mentioned that partly when discussing GPU code generation.

> I am mentioning this because finding kernels by return type and arg type is not that robust.

I absolutely agree! I guess a flag `isKernel` can be sufficient - but then we need to have metadata/attributes on the AST or symbol level. I guess storing the names during AST transformations in the helper visitor would be the best option?

I will add a todo for the function as well (saying it's a temporary thing).

src/codegen/llvm/llvm_ir_builder.cpp
pramodk: ok, thanks!

pramodk:
> @pramodk I have also added some links in the issue (regarding aliasing/attributes/metadata): see #606 and a bit of #607.

Ah ok! thanks for the references! I didn't see those.

Once you update comments, go ahead with merge.

PR #651, opened 2021-05-17
Do not open data presence block in destructors

olupton:
Without this change then with the OpenACC backend a closing brace is omitted:
```c++
void nrn_destructor_hh(NrnThread* nt, Memb_list* ml, int type) {
  #pragma acc data present(nt, ml, hh_global)
  {
    int nodecount = ml->nodecount;
    // ...
  } // this one was not printed
}
```
This breaks compilation of NEURON + CoreNEURON + NMODL with GPU support enabled. This change stops the `#pragma acc ...` block being opened.

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -3230,7 +3230,11 @@ void CodegenCVisitor::print_global_function_common_code(BlockType type) {
 
     print_global_method_annotation();
     printer->start_block("void {}({})"_format(method, args));
-    print_kernel_data_present_annotation_block_begin();
+    if (type != BlockType::Destructor) {
+        // We do not (currently) support DESTRUCTOR (and, eventually,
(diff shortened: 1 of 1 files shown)


alkino (approved):

pramodk (changes requested):
@alkino  : `CONSTRUCTOR` & `DESTRUCTOR` blocks are mostly for allocation/initialisation & deallocation/finalization. They don't have anything to run on GPU. So for these two `BlockType` , we can skip calling `print_kernel_data_present_annotation_block_begin()` in `void CodegenCVisitor::print_global_function_common_code()`.

pramodk:
@olupton : do you want to update the PR wit above suggestion? (it's straightforward otherwise I can do as well!)

olupton:
I pushed the change, note that there is not a `Constructor` value for `BlockType` at the moment.

pramodk (approved):

PR #652, opened 2021-05-17
Predication support for LLVM backend

### LINKED ISSUE #539: Address vector code generation for edge cases
There are a number of edge cases that have to be considered and supported:

- [ ] Scatters

- [ ] Gathers

- [ ] Control flow

This ticket is due to be updated after the general vector code generation is supported.

georgemitenkov:
This PR adds support for vector predication. Currently, we support a very basic predication:
```c++
IF (/*condition*/) {
  // code here
} ELSE {
  // code here
} 
```
**What has been changed and added**

1. Removed vectorization check

Before, in the `FOR` statement visitor we were checking whether the code can be vectorized. After refactoring `llvm::IRBuilder<>` into a separate class, there is no interface to reset the builder's vector width. Hence, this check leads to visitor having scalar vector width of 1, and builder having the same vector width.
```c++
if (!can_vectorize(node, sym_tab)) {
    vector_width = 1;
     ir_builder.generate_scalar_code();
}
```
In order to avoid any issues, this check is simply removed and will be added in the separate PR.

2. Predication support

- `can_vectorize` has been changed to support a single `IF` or `IF/ELSE` pair.
- A special vectorized `IF` AST node visitor has been added.
- If generating code within `IF` AST node, instructions are masked.

Example:
```c++
VOID nrn_state_test(INSTANCE_STRUCT *mech){
    INTEGER id
    INTEGER node_id
    DOUBLE v
    for(id = 0; id<mech->node_count-7; id = id+8) {
        v = mech->voltage[node_id]
        IF (mech->y[id]<0) {
            mech->y[id] = mech->y[id]+7
        } ELSE {
            mech->y[id] = v
        }
    }
    INTEGER epilogue_node_id
    DOUBLE epilogue_v
    for(; id<mech->node_count; id = id+1) {
        epilogue_node_id = mech->node_index[id]
        epilogue_v = mech->voltage[epilogue_node_id]
        IF (mech->y[id]<0) {
            mech->y[id] = mech->y[id]+7
        } ELSE {
            mech->y[id] = epilogue_v
        }
    }
}
```
is converted to
```llvm
define void @nrn_state_test(%test__instance_var__type* noalias nocapture readonly %mech1) #0 !dbg !4 {
  %mech = alloca %test__instance_var__type*, align 8
  store %test__instance_var__type* %mech1, %test__instance_var__type** %mech, align 8
  %id = alloca i32, align 4
  %node_id = alloca <8 x i32>, align 32
  %v = alloca <8 x float>, align 32
  br label %for.cond

for.cond:                                         ; preds = %for.inc, %0
  %1 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %2 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %1, i32 0, i32 10
  %3 = load i32, i32* %2, align 4
  %4 = sub i32 %3, 7
  %5 = load i32, i32* %id, align 4
  %6 = icmp slt i32 %5, %4
  br i1 %6, label %for.body, label %for.exit

for.body:                                         ; preds = %for.cond
  %7 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %8 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %7, i32 0, i32 5
  %9 = load i32, i32* %id, align 4
  %10 = sext i32 %9 to i64
  %11 = load i32*, i32** %8, align 8
  %12 = getelementptr inbounds i32, i32* %11, i64 %10
  %13 = bitcast i32* %12 to <8 x i32>*
  %14 = load <8 x i32>, <8 x i32>* %13, align 32
  store <8 x i32> %14, <8 x i32>* %node_id, align 32
  %15 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %16 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %15, i32 0, i32 4
  %17 = load <8 x i32>, <8 x i32>* %node_id, align 32
  %18 = sext <8 x i32> %17 to <8 x i64>
  %19 = load float*, float** %16, align 8
  %20 = getelementptr inbounds float, float* %19, <8 x i64> %18
  %21 = call <8 x float> @llvm.masked.gather.v8f32.v8p0f32(<8 x float*> %20, i32 1, <8 x i1> <i1 true, i1 true, i1 true, i1 true, i1 true, i1 true, i1 true, i1 true>, <8 x float> undef)
  store <8 x float> %21, <8 x float>* %v, align 32
  %22 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %23 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %22, i32 0, i32 0
  %24 = load i32, i32* %id, align 4
  %25 = sext i32 %24 to i64
  %26 = load float*, float** %23, align 8
  %27 = getelementptr inbounds float, float* %26, i64 %25
  %28 = bitcast float* %27 to <8 x float>*
  %29 = load <8 x float>, <8 x float>* %28, align 32
  %30 = fcmp olt <8 x float> %29, zeroinitializer
  %31 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %32 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %31, i32 0, i32 0
  %33 = load i32, i32* %id, align 4
  %34 = sext i32 %33 to i64
  %35 = load float*, float** %32, align 8
  %36 = getelementptr inbounds float, float* %35, i64 %34
  %37 = bitcast float* %36 to <8 x float>*
  %38 = call <8 x float> @llvm.masked.load.v8f32.p0v8f32(<8 x float>* %37, i32 1, <8 x i1> %30, <8 x float> undef)
  %39 = fadd <8 x float> %38, <float 7.000000e+00, float 7.000000e+00, float 7.000000e+00, float 7.000000e+00, float 7.000000e+00, float 7.000000e+00, float 7.000000e+00, float 7.000000e+00>
  %40 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %41 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %40, i32 0, i32 0
  %42 = load i32, i32* %id, align 4
  %43 = sext i32 %42 to i64
  %44 = load float*, float** %41, align 8
  %45 = getelementptr inbounds float, float* %44, i64 %43
  %46 = bitcast float* %45 to <8 x float>*
  call void @llvm.masked.store.v8f32.p0v8f32(<8 x float> %39, <8 x float>* %46, i32 1, <8 x i1> %30)
  %47 = xor <8 x i1> %30, <i1 true, i1 true, i1 true, i1 true, i1 true, i1 true, i1 true, i1 true>
  %48 = call <8 x float> @llvm.masked.load.v8f32.p0v8f32(<8 x float>* %v, i32 1, <8 x i1> %47, <8 x float> undef)
  %49 = load %test__instance_var__type*, %test__instance_var__type** %mech, align 8
  %50 = getelementptr inbounds %test__instance_var__type, %test__instance_var__type* %49, i32 0, i32 0
  %51 = load i32, i32* %id, align 4
  %52 = sext i32 %51 to i64
  %53 = load float*, float** %50, align 8
  %54 = getelementptr inbounds float, float* %53, i64 %52
  %55 = bitcast float* %54 to <8 x float>*
  call void @llvm.masked.store.v8f32.p0v8f32(<8 x float> %48, <8 x float>* %55, i32 1, <8 x i1> %47)
  br label %for.inc

; ... rest of the kernel
```

3. Added execution and IR tests

fixes #539

### DIFF
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -251,7 +251,8 @@ static void append_statements_from_block(ast::StatementVector& statements,
     for (const auto& statement: block_statements) {
         const auto& expression_statement = std::dynamic_pointer_cast<ast::ExpressionStatement>(
             statement);
-        if (!expression_statement->get_expression()->is_solve_block())
+        if (!expression_statement ||
+            expression_statement && !expression_statement->get_expression()->is_solve_block())
             statements.push_back(statement);
     }
 }
--- a/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_helper_visitor.cpp
@@ -251,8 +251,7 @@ static void append_statements_from_block(ast::StatementVector& statements,
     for (const auto& statement: block_statements) {
         const auto& expression_statement = std::dynamic_pointer_cast<ast::ExpressionStatement>(
             statement);
-        if (!expression_statement ||
-            expression_statement && !expression_statement->get_expression()->is_solve_block())
+        if (!expression_statement || !expression_statement->get_expression()->is_solve_block())
             statements.push_back(statement);
     }
 }
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -54,11 +54,15 @@ static bool can_vectorize(const ast::CodegenForStatement& statement, symtab::Sym
             return false;
     }
 
-    // Check there is no control flow in the kernel.
-    const std::vector<ast::AstNodeType> unsupported_nodes = {ast::AstNodeType::IF_STATEMENT};
-    const auto& collected = collect_nodes(statement, unsupported_nodes);
+    // Check for simple supported control flow in the kernel (single if/else statement).
+    const std::vector<ast::AstNodeType> supported_control_flow = {ast::AstNodeType::IF_STATEMENT};
+    const auto& supported = collect_nodes(statement, supported_control_flow);
 
-    return collected.empty();
+    // Check for unsupported control flow statements.
+    const std::vector<ast::AstNodeType> unsupported_nodes = {ast::AstNodeType::ELSE_IF_STATEMENT};
+    const auto& unsupported = collect_nodes(statement, unsupported_nodes);
+
+    return unsupported.empty() && supported.size() <= 1;
 }
 
 llvm::Value* CodegenLLVMVisitor::accept_and_get(const std::shared_ptr<ast::Node>& node) {
@@ -160,6 +164,27 @@ void CodegenLLVMVisitor::create_printf_call(const ast::ExpressionVector& argumen
     argument_values.reserve(arguments.size());
     create_function_call_arguments(arguments, argument_values);
     ir_builder.create_function_call(printf, argument_values, /*use_result=*/false);
+}
+
+void CodegenLLVMVisitor::create_vectorized_control_flow_block(const ast::IfStatement& node) {
+    // Get the true mask from the condition statement.
+    llvm::Value* true_mask = accept_and_get(node.get_condition());
+
+    // Process the true block.
+    ir_builder.set_mask(true_mask);
+    node.get_statement_block()->accept(*this);
+
+    // Note: by default, we do not support kernels with complicated control flow. This is checked
+    // prior to visiting 'CodegenForStatement`.
+    const auto& elses = node.get_elses();
+    if (elses) {
+        // If `else` statement exists, invert the mask and proceed with code generation.
+        ir_builder.invert_mask();
+        elses->get_statement_block()->accept(*this);
+    }
+
+    // Clear the mask value.
+    ir_builder.clear_mask();
 }
 
 void CodegenLLVMVisitor::find_kernel_names(std::vector<std::string>& container) {
@@ -676,6 +701,12 @@ void CodegenLLVMVisitor::visit_function_call(const ast::FunctionCall& node) {
 }
 
 void CodegenLLVMVisitor::visit_if_statement(const ast::IfStatement& node) {
+    // If vectorizing the compute kernel with control flow, process it separately.
+    if (vector_width > 1) {
+        create_vectorized_control_flow_block(node);
+        return;
+    }
+
     // Get the current and the next blocks within the function.
     llvm::BasicBlock* curr_block = ir_builder.get_current_block();
     llvm::BasicBlock* next = curr_block->getNextNode();
--- a/src/codegen/llvm/codegen_llvm_visitor.cpp
+++ b/src/codegen/llvm/codegen_llvm_visitor.cpp
@@ -547,8 +547,8 @@ void CodegenLLVMVisitor::visit_codegen_atomic_statement(const ast::CodegenAtomic
 //  | <code after for loop>     |
 //  +---------------------------+
 void CodegenLLVMVisitor::visit_codegen_for_statement(const ast::CodegenForStatement& node) {
-    // Disable vector code generation for condition and increment blocks.
-    ir_builder.stop_vectorization();
+    // Condition and increment blocks must be scalar.
+    ir_builder.generate_scalar_ir();
 
     // Get the current and the next blocks within the function.
     llvm::BasicBlock* curr_block = ir_builder.get_current_block();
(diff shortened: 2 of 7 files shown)


pramodk:
Perfect! 👌  (seems like more work for @castigli to look into performance and @iomaganaris to keep up with LLVM IR 😅)

Just curious - **without looking** into the code and implementation details, I have high level question:

 ```cpp
we support a very basic predication:

IF (/*condition*/) {
  // code here
} ELSE {
  // code here
} 
```

Is there significant difference between implementation for simple `if-else` vs. `nested if-else`?  From my experience of DSL IR implementations (not LLVM), most of the patterns are recursive. So once we implement basic case, the nested one get naturally handled. (of course with additional details like we pop stack and handle next else level etc etc.)

I wonder if there is significant difference.

**Comment by georgemitenkov** [comment]
There is no much difference in our case - I guess just `and`ing two masks together is fine. So naturally, we support that. But the performance aspects with this straightforward lowering would be quite bad I guess😅 cause of the control flow divergence in the SIMD lane.

Here would come cost metrics, if-conversions, replacements of if/else with `select`, etc. I guess it is better to leave that for now - and just stick with a single if/else pair. (Even there the cost model would be quite useful)

Speaking of the cost model (out of curiosity): in HH kernel the control flow is on the voltage. Given that voltage for the membrane has a specific "distribution" (with action potentials, -65 mV, etc) - I am wondering if we can somehow use that to estimate  the cost of vectorization, predict branches statically (in LLVM you can add branch metadata), etc.

**Comment by pramodk** [comment]
>  But the performance aspects with this straightforward lowering would be quite bad I guess😅 cause of the control flow divergence in the SIMD lane.

Make sense. Thanks for the clarifications!

> Given that voltage for the membrane has a specific "distribution" (with action potentials, -65 mV, etc) - I am wondering if we can somehow use that to estimate the cost of vectorization, predict branches statically (in LLVM you can add branch metadata), etc.

Yeah, there is scope for this. BUT, most of the compute kernels (DERIVATIVE/BREAKPOINT) can be re-written without use of IF-ELSE. So I wouldn't worry much about the complex cases and performance issues there. So we can keep life bit simple until we finish other things 😄

File: src/codegen/llvm/codegen_llvm_helper_visitor.cpp [hunk 1]
```diff
@@ -251,7 +251,8 @@ static void append_statements_from_block(ast::StatementVector& statements,
     for (const auto& statement: block_statements) {
         const auto& expression_statement = std::dynamic_pointer_cast<ast::ExpressionStatement>(
             statement);
-        if (!expression_statement->get_expression()->is_solve_block())
+        if (!expression_statement ||
+            expression_statement && !expression_statement->get_expression()->is_solve_block())
```
pramodk, line 255: Isn't `expression_statement` in  `expression_statement && ....` redundant? (because we have `!expression_statement ||`)

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -54,11 +54,15 @@ static bool can_vectorize(const ast::CodegenForStatement& statement, symtab::Sym
             return false;
     }
 
-    // Check there is no control flow in the kernel.
-    const std::vector<ast::AstNodeType> unsupported_nodes = {ast::AstNodeType::IF_STATEMENT};
-    const auto& collected = collect_nodes(statement, unsupported_nodes);
+    // Check for simple supported control flow in the kernel (single if/else statement).
+    const std::vector<ast::AstNodeType> supported_control_flow = {ast::AstNodeType::IF_STATEMENT};
+    const auto& supported = collect_nodes(statement, supported_control_flow);
```
pramodk, line 59: what happens if there is `if` within `if`? (I should check myself! )

pramodk (approved):
Overall LGTM.

src/codegen/llvm/codegen_llvm_helper_visitor.cpp
georgemitenkov: It is! Should be `!expression_statement || expression_statement->get_expression()->is_solve_block()`

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: IR would be wrong I think😅 I have removed the use of the vectorisation check in the for loop, so this function is not really used anymore.

I think a more robust version is needed where we check for nested ifs

georgemitenkov:
@pramodk This is good to merge then? For the check, as I said I plan it to be the next patch (together with nested `if` support: this is not hard to implement, just need to keep stack of masks)

pramodk:
@georgemitenkov : yes! plz go ahead!

PR #654, opened 2021-05-18
Fixed a bug when loaded value is not added to the stack

georgemitenkov:
This PR fixes the case, where loaded value was taken from the stack, but was never actually put there.


iomaganaris:
Closed and reopened PR to clean up CI states

iomaganaris (approved):
LGTM

georgemitenkov:
@iomaganaris I think the CI is intended for master, not for llvm. I will just make the PR again :)

PR #656, opened 2021-05-18
Trying to ignore llvm branches from GHA

iomaganaris:


iomaganaris:
It seems like this is not needed and new PR for `llvm` branch don't launch GHA

PR #661, opened 2021-05-20
Improvements for LLVM code generation and benchmarking

georgemitenkov:
This PR has several improvements for the current LLVM infrastructure.

1. A nicer (I guess) approach is used to check the LLVM version in `.cpp` files: instead of defining a variable, we directly check `LLVM_VERSION MAJOR`.

2. More external methods are supported now, particularly: `ceil`, `cos`, `fabs`, `floor`, `log`, `log10`, `sin`, `sqrt`. Code generation for `pow` was also fixed, and corresponding IR  tests were added. Moreover, `^` is also now supported and transformed into `pow` intrinsic.

3. Benchmarking has been also improved. Firstly, the data is initialised for every experiment (based on the Slack discussion this is the preferred way to avoid traps, unexpected timings for large number of experiments). Secondly, minimum and maximum compute times, as well as compute time variance were added to benchmarking metrics.

### DIFF
--- a/test/benchmark/llvm_benchmark.cpp
+++ b/test/benchmark/llvm_benchmark.cpp
@@ -141,7 +141,7 @@ void LLVMBenchmark::run_benchmark(const std::shared_ptr<ast::Program>& node) {
         // Log the average time taken for the kernel.
         double time_mean = time_sum / num_experiments;
         logger->info("Average compute time = {:.6f}", time_mean);
-        logger->info("Compute time variance = {:.6f}",
+        logger->info("Compute time variance = {:g}",
                      time_squared_sum / num_experiments - time_mean * time_mean);
         logger->info("Minimum compute time = {:.6f}", time_min);
         logger->info("Minimum compute time = {:.6f}\n", time_max);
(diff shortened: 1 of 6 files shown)


test/benchmark/llvm_benchmark.cpp
```diff
@@ -126,10 +132,19 @@ void LLVMBenchmark::run_benchmark(const std::shared_ptr<ast::Program>& node) {
...
             time_sum += diff.count();
+            time_squared_sum += diff.count() * diff.count();
+            time_min = std::min(time_min, diff.count());
+            time_max = std::max(time_max, diff.count());
         }
         // Log the average time taken for the kernel.
-        logger->info("Average compute time = {:.6f} \n", time_sum / num_experiments);
+        double time_mean = time_sum / num_experiments;
+        logger->info("Average compute time = {:.6f}", time_mean);
+        logger->info("Compute time variance = {:.6f}",
```
iomaganaris, line 144: With some testing I did for small/large datasets this was always zero. Maybe it would be more meaningful if it was printed as `{:g}` for example?

iomaganaris (approved):
LGTM
Just a small optional suggestion

test/benchmark/llvm_benchmark.cpp
georgemitenkov: I fully agree, thanks for suggestion!

PR #662, opened 2021-05-20
Fast math flags for LLVM backend

georgemitenkov:
This PR adds support for fast math flags in LLVM backend. Currently, the user can specify them via command-line (this approach was chosen for easier benchmarking). The specified flags are named exactly  the same as in LLVM. 

Example:
```c++
// fma.mod
FUNCTION fma(a, b, c) {
    fma = (a * b) + c
}
```
```bash
$ ./nmodl fma.mod --verbose debug llvm --ir --fmf nnan contract afn --opt
```
```llvm
define double @fma(double %a, double %b, double %c) {
  %1 = fmul nnan contract afn double %a, %b
  %2 = fadd nnan contract afn double %1, %c
  ret double %2
}
```

This is useful to enable previously unsafe FP-math optimizations. For example, fused-multiply-add instructions can now be generated when lowering LLVM IR to assembly/executing it via JIT.

For specification of FMF see LLVM LangRef: https://llvm.org/docs/LangRef.html#fastmath

### DIFF
--- a/test/unit/codegen/codegen_llvm_ir.cpp
+++ b/test/unit/codegen/codegen_llvm_ir.cpp
@@ -37,7 +37,8 @@ std::string run_llvm_visitor(const std::string& text,
                              bool opt = false,
                              bool use_single_precision = false,
                              int vector_width = 1,
-                             std::string vec_lib = "none") {
+                             std::string vec_lib = "none",
+                             std::vector<std::string> fast_math_flags = {}) {
     NmodlDriver driver;
     const auto& ast = driver.parse_string(text);
 
@@ -50,7 +51,9 @@ std::string run_llvm_visitor(const std::string& text,
                                              opt,
                                              use_single_precision,
                                              vector_width,
-                                             vec_lib);
+                                             vec_lib,
+                                             /*add_debug_information=*/false,
+                                             fast_math_flags);
     llvm_visitor.visit_program(*ast);
     return llvm_visitor.dump_module();
 }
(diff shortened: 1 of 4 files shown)


test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -1283,6 +1286,37 @@ SCENARIO("Vector library calls", "[visitor][llvm][vector_lib]") {
...
+                foo = (a * b) + c
+            }
+        )";
+
+        THEN("instructions are generated with the flags set") {
+            std::string module_string =
+                run_llvm_visitor(nmodl_text,
+                                 /*opt=*/true,
+                                 /*use_single_precision=*/false,
+                                 /*vector_width=*/1,
```
pramodk, line 1306: Just out of curiosity - why you used serial kernel?

test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -1283,6 +1286,37 @@ SCENARIO("Vector library calls", "[visitor][llvm][vector_lib]") {
...
+                                 /*vec_lib=*/"none",
+                                 /*fast_math_flags=*/{"nnan", "contract", "afn"});
+            std::smatch m;
+
+            // Check flags for produced 'fmul' and 'fadd' instructions.
+            std::regex fmul(R"(fmul nnan contract afn double %.*, %.*)");
+            std::regex fadd(R"(fadd nnan contract afn double %.*, %.*)");
```
pramodk, line 1313: 👍  - Ok, so without `reassoc`, this will be `fma` right?

pramodk (approved):
LGTM!

cc: @castigli to check FMA

test/unit/codegen/codegen_llvm_ir.cpp
georgemitenkov: We do not vectorize non-kernel functions, and here I was just testing that the flags are added (so purely semantic test)

test/unit/codegen/codegen_llvm_ir.cpp
georgemitenkov: For fma, what we really need is a `contract` (`reassoc` is more general and hence can be more inaccurate when combining FP instructions differently) - that one is more or less "safe" and is used for LLVM backends to create fmas.

georgemitenkov: I don't think there is fma in IR (i.e. it is not replaced by intrinsic). It is in SelectionDAG where fma nodes are created and subsequently converted to fma machine instructions (depending on the target)

georgemitenkov: Example: https://godbolt.org/z/7Gae9fT69 (removing `-emit-llvm -S` will show assembly with `vfmadd132pd`, while IR  has `fmul`s and `fadd`s still.

pramodk: Oh yes, I meant fmul and fadd is later combined into fma !

All looks good.

castigli:
I can confirm that I see the fma instructions being generated in the "compute-bound" kernel

PR #663, opened 2021-05-20
Fixed `alloca`s insertion point for LLVM backend

georgemitenkov:
With this PR `alloca` instructions are always inserted in the beginning of the function entry block. This is done to avoid them in the while or for loops, where allocations per iteration cause stack overflow (if the IR is not optimized). 

Example:
```c++
/// loop.mod
PROCEDURE loop(a, b) {
    WHILE (a > 0) {
        LOCAL x 
        x = b
    }
}
```
Before, this code would produce the following LLVM IR:
```llvm
define i32 @loop(double %a1, double %b2) {
  %a = alloca double, align 8
  store double %a1, double* %a, align 8
  %b = alloca double, align 8
  store double %b2, double* %b, align 8
  %ret_loop = alloca i32, align 4
  store i32 0, i32* %ret_loop, align 4
  br label %1

1:                                                ; preds = %4, %0
  %2 = load double, double* %a, align 8
  %3 = fcmp ogt double %2, 0.000000e+00
  br i1 %3, label %4, label %6

4:                                                ; preds = %1
  ; x is allocated inside the loop body - this is bad!
  %x = alloca double, align 8
  %5 = load double, double* %b, align 8
  store double %5, double* %x, align 8
  br label %1

6:                                                ; preds = %1
  %7 = load i32, i32* %ret_loop, align 4
  ret i32 %7
}
```
Now, the result is:
```llvm
define i32 @loop(double %a1, double %b2) {
  %a = alloca double, align 8
  %b = alloca double, align 8
  %ret_loop = alloca i32, align 4
  ; x is not allocated inside the loop body!
  %x = alloca double, align 8
  store double %a1, double* %a, align 8
  store double %b2, double* %b, align 8
  store i32 0, i32* %ret_loop, align 4
  br label %1

1:                                                ; preds = %4, %0
  %2 = load double, double* %a, align 8
  %3 = fcmp ogt double %2, 0.000000e+00
  br i1 %3, label %4, label %6

4:                                                ; preds = %1
  %5 = load double, double* %b, align 8
  store double %5, double* %x, align 8
  br label %1

6:                                                ; preds = %1
  %7 = load i32, i32* %ret_loop, align 4
  ret i32 %7
}
```

### DIFF
--- a/src/codegen/llvm/llvm_ir_builder.cpp
+++ b/src/codegen/llvm/llvm_ir_builder.cpp
@@ -144,7 +144,7 @@ void IRBuilder::allocate_function_arguments(llvm::Function* function,
     for (auto& arg: function->args()) {
         std::string arg_name = nmodl_arguments[i++].get()->get_node_name();
         llvm::Type* arg_type = arg.getType();
-        llvm::Value* alloca = builder.CreateAlloca(arg_type, /*ArraySize=*/nullptr, arg_name);
+        llvm::Value* alloca = create_alloca(arg_name, arg_type);
         arg.setName(arg_name);
         builder.CreateStore(&arg, alloca);
     }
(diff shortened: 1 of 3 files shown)


src/codegen/llvm/llvm_ir_builder.cpp
```diff
@@ -228,11 +228,35 @@ void IRBuilder::set_loop_metadata(llvm::BranchInst* branch) {
...
 /****************************************************************************************/
 
+llvm::Value* IRBuilder::create_alloca(const std::string& name, llvm::Type* type) {
+    // If insertion point for `alloca` instructions is not set, then set the first processed
+    // instruction to be the insertion point.
+    if (!alloca_ip) {
+        llvm::Value* alloca = builder.CreateAlloca(type, /*ArraySize=*/nullptr, name);
+        alloca_ip = llvm::cast<llvm::AllocaInst>(alloca);
+        return alloca;
+    }
```
pramodk, line 238: so we are assuming that there is always an `alloca` outside loop body and that will be our insertion point, right?
this is fine given that how our kernels look like (`node_id` and other variables are created at the beginning). May be we can add a comment here?

pramodk (approved):
LGTM! 👌

src/codegen/llvm/llvm_ir_builder.cpp
georgemitenkov: There is always one in the entry block since we create `alloca`s for every function argument.  I will add a comment saying that.

For our kernels, allocations will be also pulled out together (before they were separated by the vector for loop:
```llvm
define void @nrn_state_test(%test__instance_var__type* noalias nocapture readonly %mech1) #0 {
  %mech = alloca %test__instance_var__type*, align 8
  %id = alloca i32, align 4
  %node_id = alloca <4 x i32>, align 16
  %v = alloca <4 x double>, align 32
  ; now epilogue variables are also here!
  %epilogue_node_id = alloca i32, align 4
  %epilogue_v = alloca double, align 8
  store %test__instance_var__type* %mech1, %test__instance_var__type** %mech, align 8
  ; rest of the kernel
```

georgemitenkov: @pramodk Actually, if we have a void function with no arguments (not real, but still) - this would generate `alloca`s in-place. So I specified explicitly that we need to get the entry block and insert there.

pramodk: looks good. thanks!

PR #664, opened 2021-05-20
Avoid generating LLVM IR for Functions and Procedures if inlined

iomaganaris:
- Remove their blocks from the AST as well
- Print CMake config with all the `CMAKE_GENERATOR`s


test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -71,7 +78,7 @@ std::vector<std::shared_ptr<ast::Ast>> run_llvm_visitor_helper(
 
     SymtabVisitor().visit_program(*ast);
     SolveBlockVisitor().visit_program(*ast);
-    CodegenLLVMHelperVisitor(vector_width).visit_program(*ast);
+    CodegenLLVMHelperVisitor(vector_width, /*nmodl_inline*/ false).visit_program(*ast);
```
georgemitenkov, line 81:
```suggestion
    CodegenLLVMHelperVisitor(vector_width, /*nmodl_inline=*/false).visit_program(*ast);
```

test/unit/codegen/codegen_llvm_ir.cpp
```diff
@@ -1436,3 +1443,51 @@ SCENARIO("Dead code removal", "[visitor][llvm][opt]") {
...
+                a = 1
+                b = 2
+                test_add(a, b)
+                s = test_sub(a, b)
+            }
+        )";
+
+        THEN("when the code is inlined the procedure and function blocks are removed") {
+            std::string module_string =
+                run_llvm_visitor(nmodl_text, false, false, 1, "none", {}, true);
```
georgemitenkov, line 1483: nit: maybe can add `/*field=*/`-like comments?

iomaganaris: Good point, I forgot it :facepalm:

georgemitenkov (approved):
LGTM!

A question: would it make sense to remove function and procedure blocks even if not inlining? We transform everything to `CodegenFunction` anyway, so not visiting these.

test/unit/codegen/codegen_llvm_ir.cpp
iomaganaris: And I was wondering why `clang-format` complained about that. I forgot the `=` :facepalm:

pramodk:
> A question: would it make sense to remove function and procedure blocks even if not inlining? We transform everything to CodegenFunction anyway, so not visiting these.

 I wouldn't mind doing that if it's necessary. (keeping those functions/procedures their just helps in debugging where you can look at procedure/function and see how that is transformed/inlined)

georgemitenkov:
>I wouldn't mind doing that if it's necessary. (keeping those functions/procedures their just helps in debugging where you can look at procedure/function and see how that is transformed/inlined)

I see, thanks. It is not necessary, so let's keep everything then unless inlined!

Issue #667, opened 2021-05-21
Bugs in NMODL inliner (and consistency in AST)


georgemitenkov:
Consider the following NMODL:
```c++
PROCEDURE foo(a, b) {
    LOCAL m
    m = my_max(1, 3)
}

FUNCTION my_max(x, y) {
    IF (x > y) {
        my_max = x
    } else {
        my_max = y
    }
}
```
When inlining `my_max`, NMODL is
```
PROCEDURE foo(a, b) {
    LOCAL m, my_max_in_0
    {
        LOCAL x_in_0, y_in_0
        x_in_0 = 1
        y_in_0 = 3
        IF (x_in_0>y_in_0) {
            my_max_in_0 = x_in_0
        } ELSE {
            my_max_in_0 = y_in_0
        }
    }
    m = my_max_in_0
}
```

The problem is hidden in the AST structure: when inlining, last line is:
```json
"ExpressionStatement": [
                {
                  "BinaryExpression": [
                    {
                      "VarName": [
                        {
                          "Name": [
                            {
                              "String": [
                                {
                                  "name": "m"
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    },
                    {
                      "BinaryOperator": [
                        {
                          "name": "="
                        }
                      ]
                    },
                    {
                      "WrappedExpression": [
                        {
                          "Name": [
                            {
                              "String": [
                                {
                                  "name": "my_max_in_0"
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              ]
```

When parsing, last line is 
```json

"ExpressionStatement": [
                {
                  "BinaryExpression": [
                    {
                      "VarName": [
                        {
                          "Name": [
                            {
                              "String": [
                                {
                                  "name": "m"
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    },
                    {
                      "BinaryOperator": [
                        {
                          "name": "="
                        }
                      ]
                    },
                    {
                      "VarName": [
                        {
                          "Name": [
                            {
                              "String": [
                                {
                                  "name": "my_max_in_0"
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              ]
```

Not that in the first case `my_max_in_0` is `WrappedExpression`, and in the second - `VarName`. This inconsistency puts more pressure on the backends (handling more cases) and can introduce bugs.

Note: inliner does not seem to handle arrays as well :)
```c++
[...]

PR #669, opened 2021-05-24
Use VarName on the RHS of assignment expression

pramodk:
- NMODL parser uses VarName on the LHS of assignment expression
 - Inline visitor was using Name on the LHS of assignment expression

Related to #667

With this, original `hh.mod` can be run without any modifications:

```
DYLD_LIBRARY_PATH=/opt/intel/oneapi//compiler/2021.1.1/mac/compiler/lib \
   ./bin/nmodl ~/Downloads/CoreNeuron/coreneuron/mechanism/mech/modfile/hh.mod \
    passes --inline llvm --ir --vector-width 4 --veclib SVML --opt \
   benchmark --run --instance-size 100000 --repeat 4  \
   --libs /opt/intel/oneapi//compiler/2021.1.1/mac/compiler/lib/libsvml.dylib
```

### DIFF
--- a/src/visitors/inline_visitor.cpp
+++ b/src/visitors/inline_visitor.cpp
@@ -298,6 +298,8 @@ void InlineVisitor::visit_statement_block(StatementBlock& node) {
 /** Visit all wrapped expressions which can contain function calls.
  *  If a function call is replaced then the wrapped expression is
  *  also replaced with new variable node from the inlining result.
+ *  Note that we use `VarName` so that LHS of assignment expression
+ *  is `VarName`, similar to parser.
  */
 void InlineVisitor::visit_wrapped_expression(WrappedExpression& node) {
     node.visit_children(*this);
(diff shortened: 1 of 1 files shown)


src/visitors/inline_visitor.cpp
```diff
@@ -306,7 +308,8 @@ void InlineVisitor::visit_wrapped_expression(WrappedExpression& node) {
         auto expression = dynamic_cast<FunctionCall*>(e.get());
         if (replaced_fun_calls.find(expression) != replaced_fun_calls.end()) {
             auto var = replaced_fun_calls[expression];
-            node.set_expression(std::make_shared<Name>(new String(var)));
+            node.set_expression(
+                std::make_shared<VarName>(new Name(new String(var)), nullptr, nullptr));
```
georgemitenkov, line 312: Maybe
```suggestion
                std::make_shared<VarName>(new Name(new String(var)), /*at=*/nullptr, /*index=*/nullptr));
```
Not sure if this style is used in the rest of the codebase

georgemitenkov (approved):
LGTM!

iomaganaris (approved):
LGTM

PR #677, opened 2021-05-31
Fix OpenACC backend codegen for constructor and destructor

iomaganaris:


kotsaloscv (approved):
tested & works

src/codegen/codegen_c_visitor.cpp
```diff
@@ -3330,6 +3330,7 @@ void CodegenCVisitor::print_nrn_constructor() {
         const auto& block = info.constructor_node->get_statement_block();
         print_statement_block(*block.get(), false, false);
     }
+    print_kernel_data_present_annotation_block_end();
```
olupton, line 3333: GitHub doesn't let me comment where I want to comment...but: I am pretty sure the real problem is higher up this file, inside `print_global_function_common_code`, where we have:
```c++
if (type != BlockType::Destructor || type != BlockType::Constructor) {
  // We do not (currently) support DESTRUCTOR and CONSTRUCTOR blocks
  // running anything on the GPU.
  print_kernel_data_present_annotation_block_begin();
}
```
so the intention is clearly to not call `print_kernel_data_present_annotation_block_begin()` for constructors and destructors...but the condition is wrong. I think the correct fix is to replace `||` with `&&` there, not to call `print_kernel_data_present_annotation_block_end()` here.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -3341,6 +3342,7 @@ void CodegenCVisitor::print_nrn_destructor() {
         const auto& block = info.destructor_node->get_statement_block();
         print_statement_block(*block.get(), false, false);
     }
+    print_kernel_data_present_annotation_block_end();
```
olupton, line 3345: See above, I don't think we want this change.

olupton (changes requested):

src/codegen/codegen_c_visitor.cpp
iomaganaris: I think you're right Olli, yes :+1: I applied this change instead

olupton (approved):
LGTM, but the CI doesn't cover this at the moment so someone should check it locally.

kotsaloscv (approved):
I re-tested the amendement locally/BB5 (Olli's suggestion) and everything works fine.

alkino (approved):
Sorry for that :$

PR #681, opened 2021-06-01
Fix execution of OpenACC code with GPU disabled.

olupton:
These extra runtime branches on `NrnThread::compute_gpu` allow GPU/OpenACC-enabled translated modfiles to be executed on the host device too. This is required for the NEURON test suite to pass in a NEURON+CoreNEURON+NMODL+GPU build.


pramodk (approved):
LGTM

PR #682, opened 2021-06-02
Export GitLab CI logs.

olupton:


pramodk:
@olupton @matz-e @DriesVerachtert: Just of curiosity, have you come across any limits for number of gists on GitHub? I don't see anything obvious in the docs.

As gists are tied to account, bbpbuildbot in this case, I assume in worst case we will need to delete older gists and thats acceptable in our usage scenario.

pramodk (approved):
thank you @olupton! And @DriesVerachtert!  🚀 

(By the way - disabling neuron test in `llvm` branch was wrong call! I thought neuron tests are not necessary to run and hence disabled those. But in fact we are running neuron+coreneuron+nmodl tests. I will fix it shortly)

PR #685, opened 2021-06-03
[LLVM] Added saving to file utility

georgemitenkov:
Refactored a common functionality of saving IR to `.ll` file.

PR #686, opened 2021-06-03
[LLVM] Aliasing and `cpu` options for LLVM visitor and the benchmark

georgemitenkov:
* Added `--assume-may-alias` option to LLVM visitor that does not add `noalias` attribute for the instance struct pointer argument. This is useful for evaluating how aliasing affects IR generation and performance.

* Removed `--backend` option. Disabling features was just wrong.

* Added `--cpu` option for benchmarking.  We currently support:
```
"nehalem", "haswell", "broadwell", "skylake-avx512", "default"
```
@pramodk Feel free to add any other cpus

While this approach seems preferable to `backend` option, we still need to make sure that CPU combination and host machine have same architecture, etc. 

In theory, LLVM will break the vector widths to support these targets (Checking that)
- [ ] Check vector widths are adjusted
If this does not happen, we still can always provide the vector lengths that make sense.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -186,6 +186,9 @@ int main(int argc, const char* argv[]) {
     /// run llvm benchmark
     bool run_llvm_benchmark(false);
 
+    /// do not assume that instance struct fields do not alias
+    bool llvm_assume_alias(false);
+
     /// optimisation level for IR generation
     int llvm_opt_level_ir = 0;
 
@@ -201,8 +204,8 @@ int main(int argc, const char* argv[]) {
     /// the number of repeated experiments for the benchmarking
     int num_experiments = 100;
 
-    /// specify the backend for LLVM IR to target
-    std::string backend = "default";
+    /// specify the cpu for LLVM IR to target
+    std::string cpu = "default";
 #endif
 
     app.get_formatter()->column_width(40);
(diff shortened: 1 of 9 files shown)


src/main.cpp
```diff
@@ -351,9 +357,9 @@ int main(int argc, const char* argv[]) {
     benchmark_opt->add_option("--repeat",
                               num_experiments,
                               "Number of experiments for benchmarking ({})"_format(num_experiments))->ignore_case();
-    benchmark_opt->add_option("--backend",
-                       backend,
-                       "Target's backend ({})"_format(backend))->ignore_case()->check(CLI::IsMember({"avx2", "default", "sse2"}));
+    benchmark_opt->add_option("--cpu",
+                       cpu,
+                       "Target's backend ({})"_format(cpu))->ignore_case()->check(CLI::IsMember({"nehalem", "haswell", "broadwell", "skylake-avx512", "default"}));
```
pramodk, line 362: @georgemitenkov : what if we just remove the `check` part here? As this is benchmarking CLI option, we know what cpu name is valid. This way, it would be easy to try different things without need of building?

georgemitenkov: That's true. Fixed!

pramodk (approved):
LGTM

pramodk:
By the way @georgemitenkov : In #685 we have small change to pick-up correct clang-format. That needs to be copied here or that PR needs to be merged first.

pramodk (approved):
LGTM

pramodk:
>  Check vector widths are adjusted
> If this does not happen, we still can always provide the vector lengths that make sense.

@georgemitenkov : As this is needed for benchmarking, I will go ahead and merge it. We will check this as part of testing.

PR #690, opened 2021-06-07
Eigen openacc compatibility

kotsaloscv:
This PR solves the compatibility issues of Eigen with GPUs (I fixed both the OpenACC & CUDA backends).

The Eigen::PartialPivLU (LU decomposition to solve linear systems) is not compatible with GPUs (no __device__ tokens).

For matrices up to 4x4, the Eigen inverse() has template specializations decorated with __host__ & __device__ tokens. Therefore, we use the inverse method instead of the PartialPivLU (requires an invertible matrix) which supports both CPUs & GPUs.

For matrices 5x5 and above, Eigen does not provide GPU-enabled methods to solve small linear systems. For this reason, we use the Crout LU decomposition (Legacy code : coreneuron/sim/scopmath/crout_thread.cpp). For the CPU exectutions, we use the standard PartialPivLU from Eigen.

For the Crout LU-decomposition, I have included a unit test to compare its result with Eigen::PartialPivLU.


bbp-hpcteam (commented):
Just a quick note for failing clang-format: note that we use clang-format v11 for formatting check under CI. I have added an item for this Thursday meeting to discuss project specific clang-format version.

src/codegen/codegen_c_visitor.hpp
```diff
@@ -1656,6 +1656,13 @@ class CodegenCVisitor: public visitor::ConstAstVisitor {
         , float_type(float_type)
         , optimize_ionvar_copies(optimize_ionvar_copies) {}
 
+    static int count_length(const std::vector<SymbolType>& variables) {
+        int length = 0;
+        for (const auto& variable: variables) {
+            length += variable->get_length();
+        }
+        return length;
+    };
```
cattabiani, line 1665: this function is not really needed since it can be replaced with a one-liner:

```
std::accumulate(v.begin(), v.end(), 0, [](int l, const SymbolType& variable) {return a+= variable->get_length(); })
```

src/codegen/codegen_cuda_visitor.cpp
```diff
@@ -71,6 +74,10 @@ void CodegenCudaVisitor::print_atomic_op(const std::string& lhs,
 
 void CodegenCudaVisitor::print_backend_includes() {
     printer->add_line("#include <cuda.h>");
+
+    if (info.eigen_linear_solver_exist && count_length(info.state_vars) > 4) {
```
cattabiani, line 78: not blocking but a one-liner with accumulate here is better than a new static function since the code remains cleaner. As you prefer

test/unit/crout/crout.cpp
```diff
@@ -0,0 +1,114 @@
...
+                  Eigen::NumTraits<typename DerivedA::RealScalar>::dummy_precision(),
+              const typename DerivedA::RealScalar& atol =
+                  Eigen::NumTraits<typename DerivedA::RealScalar>::epsilon()) {
+    return ((a.derived() - b.derived()).array().abs() <= (atol + rtol * b.derived().array().abs()))
+        .all();
+}
+
+
+template <typename T>
+bool test_Crout_correctness(T rtol = 1e-6, T atol = 1e-6) {
```
cattabiani, line 39: these tolerances are quite big. why?

kotsaloscv: @cattabiani you mean that we can relax them even more (like 1e-3)?

cattabiani: Yes, I meant what Ohm said

kotsaloscv: @ohm314 & @cattabiani : I have modified Crout solver and I have resolved an accuracy issue that I had. I computed the relative error (A*x-b)/b for both eigen and crout, and crout's relative error was consistently two orders of magnitude higher than eigen one (before the modification). Therefore, for small tolerances my tests where failing. Now (after the modification), the relative error for both solvers has the same order of magnitude.

kotsaloscv:
For the build PGI pipeline we will need to bring back in CMake the following:
`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wc,--pending_instantiations=0")`
That's why it is failing for now

pramodk:
Superseded by #728

PR #695, opened 2021-06-18
Set the project version using git describe

### LINKED ISSUE #694: project version string is out of date
Probably we simply forgot to update the version string here:

https://github.com/BlueBrain/nmodl/blob/949a6ffb18f77d1dfd9bb317015b895b8820fef3/CMakeLists.txt#L11

but more generally it might be good to set the version automatically to the latest release tag.

ohm314:
We extend GitRevision.cmake a little to look up the last git tag. We
then call `project()` again in the main CMakeLists.txt to set the
version.

This fixes #694

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,10 +7,7 @@
 
 cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
 
-
-project(
-  NMODL
-  LANGUAGES CXX)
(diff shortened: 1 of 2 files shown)


pramodk (approved):
LGTM

ohm314:
Wow, that's what I call an express approval 😄

alkino:
One of the problem is to handle what happened if we shallow clone this repository.

PR #697, opened 2021-06-25
Fix the issue with ION ordering

### LINKED ISSUE #150: Issue with ION ordering in code generation
Consider below example from reduced_dentate model (#107), 

```
NEURON {
    SUFFIX ccanl
        USEION nca READ ncai, inca, enca WRITE enca, ncai VALENCE 2
        USEION lca READ lcai, ilca, elca WRITE elca, lcai VALENCE 2
        USEION tca READ tcai, itca, etca WRITE etca, tcai VALENCE 2
        RANGE caiinf, catau, cai, ncai, lcai,tcai, eca, elca, enca, etca
}

UNITS {
        (mV) = (millivolt)
    (molar) = (1/liter)
    (mM) = (milli/liter)
    (mA) = (milliamp)
    FARADAY = 96520 (coul)
    R = 8.3134  (joule/degC)
}

:INDEPENDENT {t FROM 0 TO 100 WITH 100 (ms)}

PARAMETER {
    depth = 200 (nm)    : assume volume = area*depth
    catau = 9 (ms)
    caiinf = 50.e-6 (mM)
    cao = 2 (mM)
}

ASSIGNED {
        celsius (degC)
    ica (mA/cm2)
    inca (mA/cm2)
    ilca (mA/cm2)
    itca (mA/cm2)
    cai (mM)

    enca (mV)
    elca (mV)
    etca (mV)
    eca (mV)
}

STATE {
    ncai (mM)
    lcai (mM)
    tcai (mM)
}
```

Here is comparison of data structures order from mod2c and nmodl : 

ccanl

Issues : 

* order of ion variables is wrong
* same ion variable can be used as read as well as write : in this case the variable shouldn't be declared as const

This issue was also seen while testing #148.

pramodk:
* NMODL was ordering ION variables based on symbol
  table i.e. the order in which variables appear
  in MOD file
* MOD2C orders ION variables based on the order
  of USEION statements in MOD file.
* If other blocks like PARAMETER appear first in
  MOD file with ION variables, the NMODL will end
  up with different order than MOD2C.
* In order to avoid this issue, use USEION statement
  from AST.

fixes #150

- [x] Add unit test for ION and other variables ordering

### DIFF
--- a/test/unit/visitor/inline.cpp
+++ b/test/unit/visitor/inline.cpp
@@ -27,7 +27,7 @@ using nmodl::parser::NmodlDriver;
 // Procedure/Function inlining tests
 //=============================================================================
 
-std::string run_codegen_helper_visitor(const std::string& text) {
+std::string run_inline_visitor(const std::string& text) {
     NmodlDriver driver;
     const auto& ast = driver.parse_string(text);
 
@@ -57,7 +57,7 @@ SCENARIO("Inlining of external procedure calls", "[visitor][inline]") {
 
         THEN("nothing gets inlinine") {
             std::string input = reindent_text(nmodl_text);
-            auto result = run_codegen_helper_visitor(input);
+            auto result = run_inline_visitor(input);
             REQUIRE(result == input);
         }
     }
@@ -91,7 +91,7 @@ SCENARIO("Inlining of function call as argument in external function", "[visitor
         THEN("External function doesn't get inlined") {
             std::string input = reindent_text(input_nmodl);
             auto expected_result = reindent_text(output_nmodl);
-            auto result = run_codegen_helper_visitor(input);
+            auto result = run_inline_visitor(input);
             REQUIRE(result == expected_result);
         }
     }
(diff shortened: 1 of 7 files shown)


test/unit/visitor/inline.cpp
```diff
@@ -27,7 +27,7 @@ using nmodl::parser::NmodlDriver;
 // Procedure/Function inlining tests
 //=============================================================================
 
-std::string run_inline_visitor(const std::string& text) {
+std::string run_codegen_helper_visitor(const std::string& text) {
```
iomaganaris, line 30: For this file the name of the function could be kept as `run_inline_visitor`, since we don't use the `CodegenHelperVisitor`?

iomaganaris (approved):
LGTM. Hopefully that's the last time visiting this :crossed_fingers:

test/unit/visitor/inline.cpp
pramodk: I was mistaken, I thought this was in codegen_helper test file.

PR #698, opened 2021-06-29
shasum checks for codecov.io bash script

alexsavulescu:
After codecov.io's security issue, it's worth adding extra checks for the bash uploader.


pramodk:
Referemce: https://about.codecov.io/security-update/

pramodk (approved):
same comment as nrn: https://github.com/neuronsimulator/nrn/pull/1353#discussion_r660951036

PR #700, opened 2021-07-13
Fix repr() and str() for AST python wrappers

ohm314:
The python bindings for AST classes were fixed to now properly print a
compact json representation of the node (and its children) when the repr
is called (implicitly or explictely). Additionally I added a `__str__`
to each AST node binding that will be used for string conversions (as in
print). This function returns the NMODL representation of the node (and
its children).

Added also two unit tests for this.


cattabiani:
cool! Is it related to the problems you noticed on friday?

alkino (approved):

Issue #702, opened 2021-07-16
Modifying the AST from python will create node copies that are lost


ohm314:
See the new test case here:

https://github.com/BlueBrain/nmodl/blob/c159821b2a73ad23416d1364727dd1ce279b46ba/test/unit/pybind/test_visitor.py#L94-L106

When we use a visitor to modify the AST (e.g. replace a node) from python the changes are lost. The AST nodes in the `vist_` functions seem to be copies of the "original" objects and changes on them are lost as we recurse out of the tree.

This seems to be a regression that was introduced when the visitors were changed to take node references rather than pointers. I confirmed that the issue is not present at version 0.2.

cattabiani:
once it is solved we should add a unit test to avoid repeating this error in the future

ohm314:
That unit test is what I refer to above! Thought would be good to start with that :)

Here's my work in progress: https://github.com/BlueBrain/nmodl/tree/fix/pyvst_lost_nodes
I've added a PtrVisitor class that takes pointers instead of references. I don't know, there might be a better way - one should maybe setup a tiny example with pybind11 and two classes and try to reproduce it there and find a better fix.

At any rate, my current code is not working yet. When you try to import `nmodl` you get:

```
generic_type: type "AstLookupVisitor" referenced unknown base type "nmodl::visitor::Visitor"
```

So there is a bit of a mess with the various visitors now and the way they inherit from each other.

alkino:
Fixed by #708

PR #708, opened 2021-07-20
Fixing pyvisitor template so that no node copies are done

ferdonline:
This will fix the Python visitor modifier which didn't modify the underlying AST, only a copy.

Tech detail:
We need to provide Pybind11 a reference wrapper object so it understands it's a reference and not a new object.

### DIFF
--- a/src/language/templates/pybind/pyvisitor.cpp
+++ b/src/language/templates/pybind/pyvisitor.cpp
@@ -103,25 +103,25 @@ namespace py = pybind11;
 
 {% for node in nodes %}
 void PyVisitor::visit_{{ node.class_name|snake_case }}(ast::{{ node.class_name }}& node) {
-    PYBIND11_OVERLOAD_PURE(void, Visitor, visit_{{ node.class_name|snake_case }}, node);
+    PYBIND11_OVERLOAD_PURE(void, Visitor, visit_{{ node.class_name|snake_case }}, std::ref(node));
 }
 {% endfor %}
 
 {% for node in nodes %}
 void PyAstVisitor::visit_{{ node.class_name|snake_case }}(ast::{{ node.class_name }}& node) {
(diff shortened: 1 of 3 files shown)


src/language/templates/pybind/pyvisitor.cpp
```diff
@@ -103,25 +103,25 @@ namespace py = pybind11;
 
 {% for node in nodes %}
 void PyVisitor::visit_{{ node.class_name|snake_case }}(ast::{{ node.class_name }}& node) {
-    PYBIND11_OVERLOAD_PURE(void, Visitor, visit_{{ node.class_name|snake_case }}, node);
+    PYBIND11_OVERLOAD_PURE(void, Visitor, visit_{{ node.class_name|snake_case }}, std::ref(node));
```
pramodk, line 106: Someone needs to know all the secrete / hidden documentation of pybind11 🙄 🎉

ferdonline: hidden or non-existing... this was after a lot of digging :/

pramodk: Ok...Worth an issue or PR in the upstream documentation then!

ohm314: oh wow.. i did not even know `std::ref` . amazing!

test/unit/pybind/test_visitor.py
```diff
@@ -8,6 +8,7 @@
 import nmodl
 from nmodl.dsl import ast, visitor
 import pytest
+import sys
```
ohm314, line 11: not needed

ohm314 (approved):
Fantastic! This fixes #702

src/language/templates/pybind/pyvisitor.cpp
cattabiani: +1 to Omar's comment

pramodk (approved):

PR #712, opened 2021-07-26
Add PRCELLSTATE mechanism to nmodl

### LINKED ISSUE #701: Track v and g variables when prcellstate is enabled
Similar to `mod2c`, `NMODL` should be able to track the variables of `v` and `g` when `prcellstate` is enabled (e.g. `-D NRN_PRCELLSTATE`).

Related `mod2c` code:
https://github.com/BlueBrain/mod2c/blob/92d62cc780ec09acde0524246cb34403bcf03fb3/src/mod2c_core/nocpout.c#L2226...L2235

As part of this ticket, we have:

* Like mod2c, we have to update _v_unused variable from v
* We can keep this variable update code in #ifdef so that this could be enabled only on conditional compilation. The reason for this is that this variable is only required when use prcellstate feature for debugging purpose. So in production simulation we want to avoid this extra memory update.

alkino:
PRCELLSTATE mechanism is a feature helpful to debug code.

Once nmodl is compiled you can use PRCELLSTATE by setting the variable "NRN_PRCELLSTATE" to 1.

Closes #701.

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -4329,15 +4329,15 @@ void CodegenCVisitor::print_data_structures() {
 }
 
 void CodegenCVisitor::print_v_unused() const {
-        printer->add_line("#if PRCELLSTATE");
-        printer->add_line("inst->v_unused[id] = v;");
-        printer->add_line("#endif");
+    printer->add_line("#if PRCELLSTATE");
+    printer->add_line("inst->v_unused[id] = v;");
+    printer->add_line("#endif");
 }
 
 void CodegenCVisitor::print_g_unused() const {
-        printer->add_line("#if PRCELLSTATE");
-        printer->add_line("inst->g_unused[id] = v;");
-        printer->add_line("#endif");
+    printer->add_line("#if PRCELLSTATE");
+    printer->add_line("inst->g_unused[id] = v;");
+    printer->add_line("#endif");
 }
 
 void CodegenCVisitor::print_compute_functions() {
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -2505,8 +2505,8 @@ void CodegenCVisitor::print_mechanism_global_var_structure() {
 
 
 void CodegenCVisitor::print_prcellstate_macros() const {
-    printer->add_line("#ifndef PRCELLSTATE");
-    printer->add_line("#define PRCELLSTATE 0");
+    printer->add_line("#ifndef NRN_PRCELLSTATE");
+    printer->add_line("#define NRN_PRCELLSTATE 0");
     printer->add_line("#endif");
 }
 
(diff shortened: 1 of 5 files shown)


src/codegen/fast_math.hpp
```diff
@@ -133,7 +133,7 @@ static inline double vexpm1(double initial_x) {
     double px = std::floor(LOG2E * x + 0.5);
     const int32_t n = px;
 
-    const uint64_t twopnm1 = (((uint64_t)(n - 1)) + 1023) << 52;
+    const uint64_t twopnm1 = (((uint64_t) (n - 1)) + 1023) << 52;
```
pramodk, line 136: unrelated / different clang-format?

pramodk (commented):
LGTM.

Did you test this using prcellstate feature of coreneuron?

Once CI passes, should be good to go.

src/codegen/fast_math.hpp
alkino: This is because https://github.com/BlueBrain/hpc-coding-conventions/blob/master/cpp/clang-format-7#L73

alkino: But I revert it

alkino:
As in mod2c use `NRN_PRCELLSTATE` instead of `PRCELLSTATE` for the macro name

src/codegen/codegen_c_visitor.cpp
```diff
@@ -4121,17 +4131,17 @@ void CodegenCVisitor::print_nrn_cur_conductance_kernel(const BreakpointBlock& no
...
+    std::string sum;
+    for (const auto& conductance: info.conductances) {
+        auto var = breakpoint_current(conductance.variable);
+        sum += get_variable_name(var);
+        if (&conductance != &info.conductances.back()) {
+            sum += "+";
         }
-        printer->add_line("double g = {};"_format(sum));
     }
+    printer->add_line("double g = {};"_format(sum));
```
pramodk, line 4143: Are we sure about this change? if conductances vector is empty, we are not supposed to emit this code double g = … ? (if there is a problem with prcellstate then you can check same condition there?)

alkino: The function `print_nrn_cur_conductance_kernel` is called only if `conductances` is not empty.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -4315,6 +4328,17 @@ void CodegenCVisitor::print_data_structures() {
...
 
+void CodegenCVisitor::print_v_unused() const {
+    printer->add_line("#if NRN_PRCELLSTATE");
+    printer->add_line("inst->v_unused[id] = v;");
+    printer->add_line("#endif");
+}
+
+void CodegenCVisitor::print_g_unused() const {
+    printer->add_line("#if NRN_PRCELLSTATE");
+    printer->add_line("inst->g_unused[id] = v;");
```
olupton, line 4339: Should the right hand side have `g` instead of `v`?

alkino: omg!

olupton (approved):
PR looks good to me.

It would be good to have an integration test of this, but I think that needs to be in CoreNEURON or NEURON. Perhaps it would be better to put this in NEURON because:
- in a single build we have 2 things to compare (NEURON and CoreNEURON), so we don't need to commit reference files to the repository (we could also compare direct/file transfer mode if needed)
- NEURON already has more infrastructure for this kind of test (`NeuronTestHelper`, [`compare_test_results.py`](https://github.com/neuronsimulator/nrn/blob/4c23a795484184e94f1134a408943eb205514366/test/scripts/compare_test_results.py), the script could be extended to compare prcellstate output files with a tolerance

alkino:
I rerun the PRCELLSTATE with coreneuron to test: everything fine

iomaganaris:
I also confirm that this works well and generates the same files as mod2c using `testcorenrn`

PR #715, opened 2021-08-13
Fix version handling when NMODL is submodule.

olupton:
Previously then when NMODL is built as a submodule of CoreNEURON the `GitRevision.cmake` helper from CoreNEURON was included, which does not set the expected variables. This change makes things more explicit.

Also print summary at the end of CMake execution for all build system generators, not just Unix Makefiles.


pramodk (approved):
LGTM

PR #716, opened 2021-08-13
Suppress spammy warnings in NVHPC 20.7+

olupton:
Apparently the diagnostic numbers changed in the rebranding from PGI to NVHPC: https://forums.developer.nvidia.com/t/many-all-diagnostic-numbers-increased-by-1-from-previous-values/146268/3

We might be able to fix a couple of these at source:
```
# "src/ast/all.hpp", warning #998-D: function "..." is hidden by "..." -- virtual function override intended?
# "src/codegen/codegen_cuda_visitor.cpp", NVC++-W-0277-Cannot inline function - data type mismatch
```
otherwise fixes will involve updating external packages and/or patching them.

### DIFF
--- a/cmake/CompilerHelper.cmake
+++ b/cmake/CompilerHelper.cmake
@@ -19,7 +19,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "PGI" OR CMAKE_CXX_COMPILER_ID MATCHES "NVHPC")
   # TODO : fix these warnings from template modification (#272)
   # ~~~
   if(${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 20.7)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --diag_suppress 1,82,111,115,177,186,611,997,1097,1625")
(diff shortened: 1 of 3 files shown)


cmake/CompilerHelper.cmake
```diff
@@ -18,5 +18,26 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "PGI" OR CMAKE_CXX_COMPILER_ID MATCHES "NVHPC")
   # messages specifically for AST. Disable these verbose warnings for now.
   # TODO : fix these warnings from template modification (#272)
   # ~~~
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --diag_suppress 1,82,111,115,177,186,611,997,1097,1625")
+  if(${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 20.7)
+    set(NMODL_COMPILER_WARNING_SUPPRESSIONS --diag_suppress=1,82,111,115,177,186,611,997,1097,1625)
+  else()
+    # https://forums.developer.nvidia.com/t/many-all-diagnostic-numbers-increased-by-1-from-previous-values/146268/3
```
pramodk, line 24: Sorry state of affairs! 🙄

pramodk (approved):
LGTM

PR #720, opened 2021-08-19
Avoid use of assert in the OpenACC kernel

pramodk:
* Use of assert with atomic capture result
  into wrong results
* See details in https://github.com/BlueBrain/mod2c/pull/68


olupton:
We reported a minimal reproducer of this problem (or of **a** similar problem, anyway) in https://forums.developer.nvidia.com/t/incorrect-gpu-results-with-pragma-acc-atomic-capture/187962.

PR #723, opened 2021-08-23
Fix warnings and improve ccache performance.

olupton:
This is a collection of minor fixes to compiler warnings (when compiling standalone).
It also changes how `flex` and `bison` are invoked so that relative paths are used.
This should improve the `ccache` hit rate on some files that the NVHPC compilers are extremely slow to compile.


olupton:
I suspect the CI failure (in CVF) comes from the recent "psolve direct" merges to NEURON and CoreNEURON. We should fix that, but the content of this PR could be reviewed already.

src/lexer/CMakeLists.txt
```diff
@@ -74,101 +74,124 @@ file(MAKE_DIRECTORY ${PROJECT_BINARY_DIR}/src/parser/nmodl ${PROJECT_BINARY_DIR}
...
 # Lexer & Parser commands
 # =============================================================================
-# command to generate nmodl parser
+# Command to generate nmodl parser Construct a relative path to avoid baking any absolute paths into
+# the output .cpp and .hpp files (via __FILE__ and so on), which cause ccache misses when the build
+# prefix changes (e.g. GitLab CI).
+set(NMODL_PARSER_SOURCE_DIR "${NMODL_PROJECT_SOURCE_DIR}/src/parser")
+set(NMODL_PARSER_BINARY_DIR "${NMODL_PROJECT_BINARY_DIR}/src/parser")
+file(RELATIVE_PATH NMODL_YY_FROM_PARSER_BINARY_DIR "${NMODL_PARSER_BINARY_DIR}"
+     "${NMODL_PARSER_SOURCE_DIR}/nmodl.yy")
```
pramodk, line 83: Ah ok! now understand!  👍

pramodk (approved):
LGTM!

iomaganaris (approved):
LGTM :+1:

src/lexer/CMakeLists.txt
olupton: Some extra info: CMake generates C++ compiler commands with absolute paths, but in that case `ccache` can re-write them on the fly to be relative paths ([base_dir](https://ccache.dev/manual/4.4.html#config_base_dir)). But this doesn't work for `flex` and `bison` because those are not launched by `ccache`.

alexsavulescu (approved):

PR #724, opened 2021-08-30
Adding GPU support for LLVM backend (+ some refactoring)

georgemitenkov:
This is a draft PR that aims to add GPU support for LLVM backend, focusing on CUDA targets and producing NVVM IR. What has been done:

- [x] New `Target` class has been added. The idea behind it is that previously vector width, precision of floating-point types, math libraries were passed separately. If we add GPU support, then we also would want to have GPU math library, vendor name, etc. Passing all these parameters to LLVM visitor does not seem like a nice solution, so I decided to create a  class  that encapsulates user-specified "ideal" target  - CPU/GPU.

For example:
```bash
#  default CPU
bin/nmodl mod/test.mod -o out llvm --ir

#  SIMD CPU
bin/nmodl mod/test.mod -o out llvm --ir cpu --name skylake --vector-width 2

# SIMD CPU with VLA support (for future SVE and not used atm)
bin/nmodl mod/test.mod -o out llvm --ir cpu --name aarch64 --vector-width 2 --vla

# GPU
bin/nmodl mod/test.mod -o out llvm --ir gpu --name cuda
```

This is also useful because API is nicer (`target->is_simd_cpu()` instead of `vector_width > 1`), classes have less fields, now there  is a better encapsulation of classes/members. Moreover, in the future we can add LLVM API for target triple, library information, etc. to query via `Target` class. 

- [x] GPU AST transformation and a new AST node
Now there is a `GPUId` node that represents a calculation for the thread id. Together with AST transformation, NMODL AST for GPU code generation looks like:

```c
    GPU_ID id
    INTEGER node_id
    DOUBLE v
    IF (id<mech->node_count) {
        v = mech->voltage[node_id]
        mech->m[id] = (mech->minf[id]-mech->m[id])/mech->mtau[id]
    }
}
```

This currently gives a runtime error during code generation because visitor for  `GPUId` is not added yet.

TODOs:

- [x] Add visitor for `GPUId`.
- [x] Add tests for AST transformation and NVMM intrinsics generation.
- [x] Kernel annotations
- [ ] Address spaces


georgemitenkov:
Also, the command line has changed now, see full spec below:

```bash
llvm
  LLVM code generation option
  Options:
    --ir REQUIRED                         Generate LLVM IR (false)
    --no-debug                            Disable debug information (false)
    --opt-level-ir INT:{0,1,2,3}          LLVM IR optimisation level (O0)
    --precision INT:{32,64}               Precision of floating-point types (64)
    --fast-math-flags TEXT:{afn,arcp,contract,ninf,nnan,nsz,reassoc,fast} ...
                                          Fast math flags for floating-point optimizations (none)

cpu
  LLVM CPU option
  Options:
    --name TEXT                           Specific CPU target to use
    --math-library TEXT:{Accelerate,libmvec,libsystem_m,MASSV,SLEEF,SVML,none}
                                          Math library for SIMD code generation (none)
    --vector-width INT                    Explicit vectorization width for IR generation (1)
    --vla BOOLEAN                         Generate IR for VLA target (false)

gpu
  LLVM GPU option
  Options:
    --name TEXT                           Specific CPU target to use
    --math-library TEXT:{libdevice}       Math library for SIMT code generation (none)

benchmark
  LLVM benchmark option
  Options:
    --run                                 Run LLVM benchmark (false)
    --opt-level-codegen INT:{0,1,2,3}     Machine code optimisation level (O0)
    --shared-libs TEXT:FILE ...           Shared libraries to link in JIT driver
    --mechanism-data-size INT             Size of mechanism data instance (10000)
    --num-experiments INT                 Number of experiments for benchmarking (100)
``

PR #726, opened 2021-09-02
Fix the issue with Eigen+OpenACC while executing on GPU

pramodk:
Update **_conspiracy theory here_** once @kotsaloscv  (and @olupton soon)
confirms that testcorenrn and channel-benchmark tests are working 😅 

Branches to use:

* neuron: master
* coreneuron: master
* nmodl: pramodk/eigen_openacc_compatibility

I quickly testcorenrn/hhkin and looks good.

@olupton : seems like there is something additional in channel-benchmark. Will look into details shortly...

Issue #311, opened 2020-04-27
Eigen compatibility with OpenACC


iomaganaris:
This issue was raised when running `nrnivmodl-core` with the following `ModelDB` model https://senselab.med.yale.edu/ModelDB/ShowModel?model=19176&file=%2fHCN2k%2fhcn2.mod#tabs-2 using `PGI 19.4` and `PGI 19.10` with the `OpenACC` backend generated from `NMODL`.
The generated `c++` file for `hcn2.mod` file contains a call to the following `Eigen solver` generated by the translation of the `DERIVATIVE` block:
```
X = Eigen::PartialPivLU<Eigen::Ref<Eigen::Matrix<double, 4, 4>>>(Jm).solve(F);
```
Compiling this file with `pgc++` there is the following issue (with `-Minfo=acc` added to the compilation flags):
```
PGCC-W-0155-External and Static variables are not supported in acc routine - _ZTVSt9exception (x86_64/core/mod2c/hcn2.cpp: 63)
std::exception::exception():
     22, include "multicore.hpp"
          32, include "membfunc.hpp"
               31, include "vector"
                    61, include "allocator.h"
                         46, include "c++allocator.h"
                              33, include "new_allocator.h"
                                   33, include "new"
                                        40, include "exception"
                                             63, Generating implicit acc routine seq
PGCC-W-0155-External and Static variables are not supported in acc routine - _ZTVSt9exception (x86_64/core/mod2c/hcn2.cpp: 63)
std::exception::exception() [subobject]:
      0, Generating implicit acc routine seq
PGCC-W-0155-External and Static variables are not supported in acc routine - _ZTVSt9exception (x86_64/core/mod2c/hcn2.cpp: 57)
std::bad_alloc::bad_alloc():
     22, include "multicore.hpp"
          32, include "membfunc.hpp"
               31, include "vector"
                    61, include "allocator.h"
                         46, include "c++allocator.h"
                              33, include "new_allocator.h"
                                   33, include "new"
                                        57, Generating implicit acc routine seq
```
This was due to throwing an exception in https://gitlab.com/libeigen/eigen/-/blob/master/Eigen/src/Core/util/Memory.h#L70.
[...]

Issue #135, opened 2019-04-06
Issue with newton solver when used with OpenACC backend


pramodk:
If I compile newton solver example with OpenACC backend we get:

```
Eigen::internal::scalar_sum_op<double, double>::operator ()(const double &, const double &) const:
     31, include "newton.hpp"
          14, include "LU"
               11, include "Core"
                    73, include "cassert"
                        120, include "Core"
                             219, include "BinaryFunctors.h"
                                   42, Generating implicit acc routine seq
                                       Generating acc routine seq
                                       Generating Tesla code
                             240, include "EigenBase.h"
                             243, include "CoreEvaluators.h"
                             261, include "Matrix.h"
                             265, include "CwiseUnaryOp.h"
                             269, include "Dot.h"
                             279, include "Transpose.h"
                             283, include "Redux.h"
PGCC-W-0155-Compiler failed to translate accelerator region (see -Minfo messages): Unsupported type 256. (/gpfs/bbp.cscs.ch/project/proj20/pramod_scratch/SC_BENCHMARKS/paper/coreneuron/build/coreneuron/cacumm.cpp: 1)
const double Eigen::internal::scalar_sum_op<double, double>::predux<__m128d>(const T1 &) const:
     31, include "newton.hpp"
          14, include "LU"
               11, include "Core"
                    73, include "cassert"
                        120, include "Core"
                             219, include "BinaryFunctors.h"
                                   48, Generating implicit acc routine seq
                                       Generating acc routine seq
                                       Generating Tesla code
                                      718, Accelerator restriction: unsupported operation: LDQ
                             240, include "EigenBase.h"
                             243, include "CoreEvaluators.h"
                             261, include "Matrix.h"
                             265, include "CwiseUnaryOp.h"
                             269, include "Dot.h"
                             279, include "Transpose.h"
                             283, include "Redux.h"
PGCC-F-0704-Compilation aborted due to previous errors. (/gpfs/bbp.cscs.ch/project/proj20/pramod_scratch/SC_BENCHMARKS/paper/coreneuron/build/coreneuron/cacumm.cpp)
```

PR #728, opened 2021-09-06
Compatibility issues between Eigen and GPUs (OpenACC/CUDA)

kotsaloscv:
This PR solves the compatibility issues between Eigen and GPUs (OpenACC/CUDA).

Two factors contribute to the above solution:

1. New Eigen branch (version 3.5 and above). Currently we are using a mirrored version of Eigen (https://github.com/BlueBrain/eigen/releases/tag/v3.5-alpha.1).
2. An API that makes possible to call any Eigen `__device__` function from within OpenACC regions. In more details, Eigen-3.5+ provides better GPU support; however, some functions cannot be called directly from within OpenACC regions. Therefore, we need to wrap them in a special API (decorate them with `__device__` & `acc routine` tokens), which allows us to eventually call them from OpenACC. Calling these functions from CUDA kernels presents no issue ...

This PR closes/replaces the following PRs:

1. #690 
2. #726 

**This PR works in combination with CoreNEURON's PR # 624 (minor changes to include the new Eigen API).**

This PR closes the following issues:

1. #311 
2. #135

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -7,6 +7,9 @@
 
 #include "codegen/codegen_acc_visitor.hpp"
 
+#include "ast/eigen_linear_solver_block.hpp"
+#include "ast/integer.hpp"
+
 
 using namespace fmt::literals;
 
@@ -73,6 +76,15 @@ void CodegenAccVisitor::print_backend_includes() {
         printer->add_line("#include <cuda_runtime_api.h>");
         printer->add_line("#include <openacc.h>");
     }
+
+    if (info.eigen_linear_solver_exist && std::accumulate(info.state_vars.begin(),
+                                                          info.state_vars.end(),
+                                                          0,
+                                                          [](int l, const SymbolType& variable) {
+                                                              return l += variable->get_length();
+                                                          }) > 4) {
+        printer->add_line("#include <partial_piv_lu/partial_piv_lu.h>");
+    }
 }
 
 
@@ -122,6 +134,23 @@ void CodegenAccVisitor::print_abort_routine() const {
 
 void CodegenAccVisitor::print_net_send_buffering_grow() {
     // can not grow buffer during gpu execution
+}
+
+void CodegenAccVisitor::print_eigen_linear_solver(const std::string& float_type,
+                                                  int N,
+                                                  const std::string& X,
+                                                  const std::string& Jm,
+                                                  const std::string& F) {
+    if (N <= 4) {
+        printer->add_line("{0} = {1}.inverse()*{2};"_format(X, Jm, F));
+    } else {
+        // Eigen-3.5+ provides better GPU support. However, some functions cannot be called directly
+        // from within an OpenACC region. Therefore, we need to wrap them in a special API (decorate
+        // them with
+        // __device__ & acc routine tokens), which allows us to eventually call them from OpenACC.
+        // Calling these functions from CUDA kernels presents no issue ...
+        printer->add_line("{0} = partialPivLu<{1}>({2}, {3});"_format(X, N, Jm, F));
+    }
 }
 
 /**
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -138,18 +138,13 @@ void CodegenAccVisitor::print_net_send_buffering_grow() {
 
 void CodegenAccVisitor::print_eigen_linear_solver(const std::string& float_type,
                                                   int N,
-                                                  const std::string& X,
+                                                  const std::string& Xm,
                                                   const std::string& Jm,
-                                                  const std::string& F) {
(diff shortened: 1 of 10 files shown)


src/codegen/codegen_acc_visitor.cpp
```diff
@@ -124,6 +136,18 @@ void CodegenAccVisitor::print_net_send_buffering_grow() {
     // can not grow buffer during gpu execution
 }
 
+void CodegenAccVisitor::print_eigen_linear_solver(const std::string& float_type,
+                                                  int N,
+                                                  const std::string& Xm,
```
iomaganaris, line 141:
```suggestion
                                                  const std::string& X,
```
Should this be like this to follow the declaration of the function in the header file or change the name of the variable in the declaration?

src/codegen/codegen_c_visitor.hpp
```diff
@@ -1898,6 +1899,11 @@ class CodegenCVisitor: public visitor::ConstAstVisitor {
     void visit_function_call(const ast::FunctionCall& node) override;
     void visit_eigen_newton_solver_block(const ast::EigenNewtonSolverBlock& node) override;
     void visit_eigen_linear_solver_block(const ast::EigenLinearSolverBlock& node) override;
+    virtual void print_eigen_linear_solver(const std::string& float_type,
+                                           int N,
+                                           const std::string& X,
```
iomaganaris, line 1904: Same here with the naming of the variable `X`?

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1779,13 +1785,16 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(const ast::EigenNewtonSolv
...
     std::string J = find_var_unique_name("J");
     std::string Jm = find_var_unique_name("Jm");
     std::string F = find_var_unique_name("F");
+    std::string Fm = find_var_unique_name("Fm");
 
     auto float_type = default_float_data_type();
     int N = node.get_n_state_vars()->get_value();
-    printer->add_line("Eigen::Matrix<{}, {}, 1> {};"_format(float_type, N, X));
+    printer->add_line("Eigen::Matrix<{}, {}, 1> {};"_format(float_type, N, Xm));
+    printer->add_line("{}* {} = {}.data();"_format(float_type, X, Xm));
```
iomaganaris, line 1797: I was wondering what is the purpose of defining the variable `X`, since I don't see it being used

iomaganaris (commented):
LGTM. Just a few questions

src/codegen/codegen_c_visitor.cpp
kotsaloscv: The variables X, F & J are used in the generated code defined in :
```
print_statement_block(*node.get_initialize_block(), false, false);
print_statement_block(*node.get_setup_x_block(), false, false);
```
So all of them are used.

iomaganaris: Aha, I see now. Thanks for the clarification :+1:

iomaganaris (approved):

src/solver/newton/newton.hpp
```diff
@@ -15,6 +15,10 @@
  * \brief Implementation of Newton method for solving system of non-linear equations
  */
 
+#if defined(_OPENACC) && !defined(DISABLE_OPENACC)
```
pramodk, line 18: Just a note: this will be also used in cuda backend. So this could be below?
```suggestion
#if (defined(__CUDACC__) || defined(_OPENACC)) && !defined(DISABLE_OPENACC)
```

src/solver/newton/newton.hpp
```diff
@@ -69,9 +73,13 @@ EIGEN_DEVICE_FUNC int newton_solver(Eigen::Matrix<double, N, 1>& X,
             // we have converged: return iteration count
             return iter;
         }
+#if defined(_OPENACC) && !defined(DISABLE_OPENACC)
```
pramodk, line 76: same

src/solver/newton/newton.hpp
```diff
@@ -142,10 +150,13 @@ EIGEN_DEVICE_FUNC int newton_numerical_diff_solver(Eigen::Matrix<double, N, 1>&
             // restore X
             X[i] += dX;
         }
-        // update X
-        // use in-place LU decomposition of J with partial pivoting
+#if defined(_OPENACC) && !defined(DISABLE_OPENACC)
+        X -= partialPivLu<N>(J, F);
+#else
```
pramodk, line 155: same

pramodk (approved):
Apart from minor comment, LGTM

src/solver/newton/newton.hpp
kotsaloscv: From within CUDA kernels, we can call immediately this eigen function and therefore no need to call the wrapper

pramodk: Oh ok. Perfect!

PR #730, opened 2021-09-06
Fix race condition issue with STEADYSTATE code generation on GPU backends

### LINKED ISSUE #729: Race condition in OpenACC/GPU code generation with STEADYSTATE
Consider [SH_na8st.mod](https://github.com/pramodk/reduced_dentate/blob/master/mechanisms/SH_na8st.mod) from reduced_dentate model which has `STEADYSTATE`:

```python
BREAKPOINT {
    SOLVE kin METHOD sparse
    g = gbar*o
    ina = g*(v - ena)*(1e-3)
}

INITIAL { SOLVE kin STEADYSTATE sparse }

KINETIC kin {
    rates(v)
    ~ c1 <-> c2 (a1, b1)
    ~ c2 <-> c3 (a2, b2)
    ~ c3 <-> o (a3, b3)
    ~ i1 <-> i2 (a1, b1)
    ~ i2 <-> i3 (a2, b2)
    ~ i3 <-> i4 (a3, b3)
    ~ i1 <-> c1 (ah, bh)
    ~ i2 <-> c2 (ah, bh)
    ~ i3 <-> c3 (ah, bh)
    ~ i4 <-> o  (ah, bh)
    CONSERVE c1 + c2 + c3 + i1 + i2 + i3 + i4 + o = 1
}
```

which results into following OpenACC code:

```cpp
    /** initialize channel */
    void nrn_init_na8st(NrnThread* nt, Memb_list* ml, int type) {
        #pragma acc data present(nt, ml, na8st_global) if(nt->compute_gpu)
        {
            int nodecount = ml->nodecount;
            int pnodecount = ml->_nodecount_padded;
            const int* __restrict__ node_index = ml->nodeindices;
            double* __restrict__ data = ml->data;
            const double* __restrict__ voltage = nt->_actual_v;
            Datum* __restrict__ indexes = ml->pdata;
            ThreadDatum* __restrict__ thread = ml->_thread;

            setup_instance(nt, ml);
            na8st_Instance* __restrict__ inst = (na8st_Instance*) ml->instance;
...
            #pragma acc update device (na8st_global)
            if (_nrn_skip_initmodel == 0) {
                int start = 0;
                int end = nodecount;
                #pragma acc parallel loop seq present(inst, node_index, data, voltage, indexes, thread) \
                    copyin(X[0:DIM], F[0:DIM], pivot[0:DIM], J[0:64]) if(nt->compute_gpu)
                for (int id = start; id < end; id++) {
                    int node_id = node_index[id];
....
                    Eigen::Matrix<double, 8, 1> Xm, Fm;
                    Eigen::Matrix<double, 8, 8> Jm;
                    double* X = Xm.data();
                    double* J = Jm.data();
                    double* F = Fm.data();
                    double dt_saved_value, old_c1, old_c2, old_c3, old_i1, old_i2, old_i3, old_i4;
                    dt_saved_value = nt->_dt;
                    nt->_dt = 1000000000.0;
                    
...

                    inst->c1[id] = X[0];
                    inst->c2[id] = X[1];
                    inst->c3[id] = X[2];
                    inst->i1[id] = X[3];
                    inst->i2[id] = X[4];
                    inst->i3[id] = X[5];
                    inst->i4[id] = X[6];
                    inst->o[id] = X[7];
                    nt->_dt = dt_saved_value;
```

Notice the update of `nt->_dt` in the for loop: if the mechanisms are executed in parallel (SIMD or GPU execution) then this will result into race condition.

See why `dt` is changed in:
* here: https://github.com/BlueBrain/nmodl/blob/1321c7b78b40331605caaa6a3d8b4bf6417214a8/src/visitors/steadystate_visitor.hpp#L29
* here: https://github.com/BlueBrain/nmodl/blob/1321c7b78b40331605caaa6a3d8b4bf6417214a8/src/visitors/steadystate_visitor.cpp#L58

pramodk:
* update of dt was done inside parallel/SIMD for loop
 * this leads to race condition and random failures on GPU
 * handle update of dt in code generation backend: pull dt
   update outside loop.
 * CUDA backend doesn't handle these updates yet. This should
   be handled in separate PR.
* Added new statement type `UpdateDt` because we can easily
   search and pull it outside loop to avoid the race condition.

fixes #729

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -297,5 +297,10 @@ void CodegenAccVisitor::print_net_send_buf_count_update_to_device() const {
     printer->add_line("#pragma acc update device(nsb->_cnt) if (nt->compute_gpu)");
 }
 
+void CodegenAccVisitor::print_dt_update_to_device() const {
+    printer->add_line("#pragma acc update device({}) if (nt->compute_gpu)"_format(
+        get_variable_name(naming::NTHREAD_DT_VARIABLE)));
+}
+
 }  // namespace codegen
 }  // namespace nmodl
--- a/src/codegen/codegen_acc_visitor.hpp
+++ b/src/codegen/codegen_acc_visitor.hpp
@@ -100,6 +100,9 @@ class CodegenAccVisitor: public CodegenCVisitor {
     // update NetSendBuffer_t count from host to device
     virtual void print_net_send_buf_count_update_to_device() const override;
 
+    // update dt from host to device
+    virtual void print_dt_update_to_device() const override;
+
     // synchronise/wait on stream specific to NrnThread
     virtual void print_device_stream_wait() const override;
 
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -318,6 +318,9 @@ void CodegenCVisitor::visit_verbatim(const Verbatim& node) {
     }
 }
 
+void CodegenCVisitor::visit_update_dt(const ast::UpdateDt& node) {
+    // dt change statement should be pulled outside already
+}
 
 /****************************************************************************************/
 /*                               Common helper routines                                 */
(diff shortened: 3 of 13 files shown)


iomaganaris (approved):
LGTM!

PR #733, opened 2021-09-15
Fix NMODL compilation when clang-format was found.

olupton:
Without this fix then building NMODL-as-a-submodule-of-CoreNEURON-as-a-submodule-of-NEURON would not
work if code formatting was enabled in NEURON following the recipe of https://github.com/neuronsimulator/nrn/pull/1460.

This was apparently because NMODL was formatting NMODL generated code using a clang-format configuration file from NEURON, which uses `Standard: Cpp03` (we should change that too). That breaks NMODL code compilation by adding a space before `_format` in `"..{}.."_format("bar")`, which is not valid.

Also change the handling of the --clang-format-opts option so that the `${NMODL_ClangFormat_OPTIONS}` option should actually work.
Previously the CMake code would generate something like:
```
  --clang-format-opts A B --clang-format-opts="--style=file"
```
where the `A B` part would then be ignored.


olupton:
GitLab CI needs https://github.com/BlueBrain/nmodl/pull/734 to pass, it's not related to this PR.

ohm314 (approved):
looks good to me!

PR #735, opened 2021-09-15
Add formatting checks to GitHub CI.

olupton:
Update GitHub Actions CI config to:
- run `clang-format` and `cmake-format` checks, so these can be dropped from the GitLab CI
- use `ninja` and `ccache` to improve performance
- not install so many packages
- test on Ubuntu 20.04 and macOS 11 too
- unfortunately **not** run the tests in parallel, because of https://github.com/BlueBrain/nmodl/issues/737

Also make the code that tries to derive the NMODL version from the git tag history fail more elegantly when the relevant information is not available (e.g. in a shallow git clone done in the CI).


olupton:
The top 3 workflows are with a "warm" ccache, and the bottom 3 are with a "cold" cache.
There is a big speedup for the "NMODL CI" workflow, and smaller improvements in the coverage/documentation.
Screenshot 2021-09-16 at 11 13 01

alexsavulescu (approved):

PR #738, opened 2021-09-16
Added CLI for transforming top LEVEL local variables to RANGE

### LINKED ISSUE #732: Don't change unconditionally top level LOCAL variables to ASSIGNED
At the moment top level `LOCAL` variables are turned into `ASSIGNED` using the `LocalToAssignedVisitor` in https://github.com/BlueBrain/nmodl/blob/5598d7cafe81512b22d627eda5f99a89c80ac217/src/main.cpp#L351
This has as a consequence the generated code (`float_variables_size()`) to have different number of float variables than the generated code by `mod2c` (`_psize` variable) and `NEURON nmodl` breaking the functionality of the code.
We should add flag to only do this transformation conditionally.
See related PR also: https://github.com/BlueBrain/nmodl/pull/347

iomaganaris:
- By default the option is disabled so that NMODL and nocmodl/mod2c
  generate code with the same instance variables
  
 Fixes #732

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -223,7 +223,7 @@ int main(int argc, const char* argv[]) {
          "Convert GLOBAL variables to RANGE ({})"_format(nmodl_global_to_range))->ignore_case();
     passes_opt->add_flag("--local-to-range",
          nmodl_local_to_range,
-         "Convert tol level LOCAL variables to RANGE ({})"_format(nmodl_local_to_range))->ignore_case();
+         "Convert top level LOCAL variables to RANGE ({})"_format(nmodl_local_to_range))->ignore_case();
     passes_opt->add_flag("--localize-verbatim",
         localize_verbatim,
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -218,6 +221,9 @@ int main(int argc, const char* argv[]) {
     passes_opt->add_flag("--global-to-range",
          nmodl_global_to_range,
          "Convert GLOBAL variables to RANGE ({})"_format(nmodl_global_to_range))->ignore_case();
+    passes_opt->add_flag("--local-to-range",
+         nmodl_local_to_range,
+         "Convert tol level LOCAL variables to RANGE ({})"_format(nmodl_local_to_range))->ignore_case();
```
olupton, line 226:
```suggestion
         "Convert top level LOCAL variables to RANGE ({})"_format(nmodl_local_to_range))->ignore_case();
```

olupton (approved):
LGTM, I only spotted a one-letter typo.

PR #741, opened 2021-09-20
Bug fix for net_init code generation on GPUs with OpenACC backend

pramodk:
* net_init was being executed on CPU but it was dereferencing
   gpu pointers on CPU via instance structure
 * Currently there is no support for getting device function pointers
 * As net_init is called from cpu side, a simple approach is to launch
   serial kernel for each invocation of net_init.
 * This is bit inefficient as large number of kernels could be launched.
   But, as this happens only during initialialization, we can use this
   solution for time being.
 * Note that mod2c "get away" by launching kernels on CPU because the
   net_init kernel is just handling spike events and those are handled
   on CPU anyway.

fixes #740

### Before merge

- [x] @olupton to confirm if the hippocampus model runs on the GPU with this PR
- [x] Update docs/clang-format

PR #745, opened 2021-09-29
live debugging for PRs only

alexsavulescu:


pramodk (commented):
LGTM, I will let Olli to confirm the changes for live-debug

.github/workflows/nmodl-ci.yml
```diff
@@ -100,7 +100,7 @@ jobs:
             ${{matrix.os}}-${{github.ref}}-
             ${{matrix.os}}-
 
-      - name: Build
+      - name: Build, Test and Install
```
olupton, line 103: What problem is this solving? When I was debugging #741 (I had replaced the live debugging condition with `if: failure()`) then it was working for failures in the "Test" step.

alexsavulescu: That's strange, because `failure()` only works for previous step and not steps. From my testing today, it didn't trigger.

olupton: Weird. I had e.g. [this](https://github.com/BlueBrain/nmodl/runs/3731763593) live debug session working with [this](https://github.com/BlueBrain/nmodl/blob/33b451e410d14a2c77e1e1a6aef1de79a98d4d71/.github/workflows/nmodl-ci.yml) configuration.

olupton: What do you want to do about this @alexsavulescu?

alexsavulescu: Let's do one step, no added benefit in having three.

olupton: OK. I just pushed a fix for the merge conflict.

Issue #737, opened 2021-09-16
CTest suite fails when run in parallel


olupton:
It seems that running the CTest suite in parallel can produce errors, for example [in this CI run with 3 parallel jobs](https://github.com/BlueBrain/nmodl/runs/3614630061?check_suite_focus=true#step:12:116):
```
51/85 Test #17: testparser/Scenario: Check parents in valid NMODL constructs ................................***Failed    0.71 sec

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
testparser is a Catch v2.0.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
Scenario: Check parents in valid NMODL constructs
-------------------------------------------------------------------------------
../test/unit/parser/parser.cpp:188
...............................................................................

../test/unit/parser/parser.cpp:195: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  NMODL Parser Error : can not open file : Unit.inc [Location : 2.25-34]

Cannot delete temporary file: No such file or directory
Cannot delete temporary file: No such file or directory
-------------------------------------------------------------------------------
Scenario: Check parents in valid NMODL constructs
-------------------------------------------------------------------------------
../test/unit/parser/parser.cpp:188
...............................................................................

../test/unit/parser/parser.cpp:188: FAILED:
due to unexpected exception with message:
  NMODL Parser Error : can not open file : Unit.inc [Location : 2.25-34]

===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 34 | 32 passed | 2 failed
```
I have also seen this locally. I suspect the problem is that multiple tests use a temporary file with the same name (`Unit.inc`).

1uc:
Currently working.

PR #746, opened 2021-09-29
Use hpc/gitlab-pipelines helpers.

olupton:
This should reduce duplication and make NMODL more similar to, for example, CoreNEURON.
The clang-format and cmake-format checks have already been migrated to the GitHub Actions CI in https://github.com/BlueBrain/nmodl/pull/735, so we are not really losing them.
Test parallelism is disabled because of https://github.com/BlueBrain/nmodl/issues/737.


.gitlab-ci.yml
```diff
@@ -1,34 +1,11 @@
 include:
   - project: hpc/gitlab-pipelines
-    file: github-project-pipelines.gitlab-ci.yml
+    file:
+      - spack-build-components.gitlab-ci.yml
+      - github-project-pipelines.gitlab-ci.yml
```
matz-e, line 5: This can be a list? Damn…

matz-e (approved):

PR #747, opened 2021-10-08
Ubsan match

ohm314:


ohm314:
sorry - went into the wrong repo...

PR #752, opened 2021-10-13
codecov new uploader

alexsavulescu:
* bash script will be deprecated
* download new binary uploader and perform integrity checks


alexsavulescu:
> LGTM (by the way, I haven't looked at the all details, I assume these are standard instructions from codecov.io)

Indeed, all working without any change. For the old shasum checks I had to fiddle around :/

PR #758, opened 2021-10-22
Return var name with index

### LINKED ISSUE #511: ' parsed as operator
for this [PR](https://github.com/BlueBrain/nmodl/pull/495) I need to get the variable of a derivative. For example:

```
a' -> a
F'[0] -> F[0]
```

Right now, the ' is embedded in the definition of the name node. Name vs PrimeName. In addition, this means that we can have only 1st derivatives of a variable (in case we want to expand the language). 

It would be nice to have it as an operator instead so that we can convert from derivative to variable simply erasing that node.

WeinaJi:
* Add function in visitor_utils to return node name with index for IndexedName node
* Add a visitor to visit IndexedName node and get the dependencies of DiffEqExpression

Fixes #511


iomaganaris (approved):
LGTM. Maybe @cattabiani wants to have a look as well?

test/unit/visitor/node_index.cpp
```diff
@@ -0,0 +1,80 @@
...
+                auto result_dependencies = get_indexedname_dependencies(*ast).second;
+                REQUIRE(result_name == var);
+                REQUIRE(result_dependencies.first == expect.first);
+                REQUIRE(result_dependencies.second == expect.second);
+            }
+            THEN("Get dependencies") {
+                auto ast = to_ast(nmodl_text_b);
+                std::unordered_set<std::string> vars{"m", "h"};
+                std::string var("m");
+                auto expect = std::make_pair(var, vars);
```
cattabiani, line 72: this pair is not really useful but I do not mind

test/unit/visitor/node_index.cpp
```diff
@@ -0,0 +1,80 @@
...
+                auto result_name = get_indexedname_dependencies(*ast).first;
+                auto result_dependencies = get_indexedname_dependencies(*ast).second;
+                REQUIRE(result_dependencies.first == expect.first);
+                REQUIRE(result_dependencies.second == expect.second);
+            }
+        }
+    }
+}
```
cattabiani, line 80: may we also test this case?

`a' -> a`

src/visitors/visitor_utils.cpp
```diff
@@ -260,4 +260,19 @@ std::pair<std::string, std::unordered_set<std::string>> statement_dependencies(
     return {key, out};
 }
 
+std::string get_indexed_name(const ast::IndexedName& node) {
+    return node.get_node_name() + "[" + to_nmodl(node.get_length()) + "]";
```
cattabiani, line 264: I think you can use fmt instead to avoid extending the string

src/visitors/visitor_utils.cpp
```diff
@@ -260,4 +260,19 @@ std::pair<std::string, std::unordered_set<std::string>> statement_dependencies(
     return {key, out};
 }
 
+std::string get_indexed_name(const ast::IndexedName& node) {
```
cattabiani, line 263: shouldn't this be a method of the class `IndexedName`?

src/visitors/visitor_utils.cpp
```diff
@@ -260,4 +260,19 @@ std::pair<std::string, std::unordered_set<std::string>> statement_dependencies(
...
+    return node.get_node_name() + "[" + to_nmodl(node.get_length()) + "]";
+}
+
+std::string get_full_var_name(const ast::VarName& node) {
```
cattabiani, line 267: also this could be a method of the class `VarName`

cattabiani (changes requested):
Very nice! I have just a few questions and a small addition to ask:
- why do you need `IndexedNameVisitor`? Only for the unit test? I forgot now: why the dependencies need to be updated and how was it working before?

cattabiani:
>LGTM. Maybe @cattabiani wants to have a look as well?
yes, thx @iomaganaris I missed this pr

src/visitors/visitor_utils.cpp
iomaganaris: I think that this is just a pain to do because of the need of `to_nmodl` function. `get_length` returns an `Expression` that cannot be easily converted to `string` since it might have various different forms. I'm not sure if it's worth it to complicate the current code that much

src/visitors/visitor_utils.cpp
iomaganaris: For reasons applied at the above function I think it's fine to keep it as it is

test/unit/visitor/node_index.cpp
iomaganaris: This is already tested with
```
std::string nmodl_text_b = R"(
            BREAKPOINT  {
                SOLVE states STEADYSTATE sparse
            }
            DERIVATIVE states {
                m' = m + h
            }
        )";
```
I believe

src/visitors/indexedname_visitor.hpp
```diff
@@ -0,0 +1,68 @@
...
+    /// Get dependencies for the DiffEqExpression node
+    void visit_diff_eq_expression(ast::DiffEqExpression& node) override;
+
+    void visit_program(ast::Program& node) override;
+
+    /// get the attribute indexed_name
+    std::string get_indexed_name();
+
+    /// get the attribute dependencies
+    std::pair<std::string, std::unordered_set<std::string>> get_dependencies();
```
pramodk, line 62: nitpick: could you elaborate comment here ? I don't easily guess what is `attribute` referring here.

pramodk (approved):
LGTM

src/visitors/indexedname_visitor.hpp
WeinaJi: Hi @pramodk , the attribute means the private member `std::pair<std::string, std::unordered_set<std::string>> dependencies` of this visitor class.  The value of `dependencies` is assigned by the member function  above
``` 
    /// Get dependencies for the DiffEqExpression node
    void visit_diff_eq_expression(ast::DiffEqExpression& node) override;
```

WeinaJi:
@iomaganaris , thanks for following this PR.  Let's try to close it this year !

PR #766, opened 2021-11-02
Replace macro name by value

alkino:
macro is created by DEFINE statements, and always got a direct value
DEFINE NAME_PTR INTEGER_PTR is the only accepted way

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -69,13 +69,8 @@ void CodegenCVisitor::visit_integer(const Integer& node) {
     if (!codegen) {
         return;
     }
-    const auto& macro = node.get_macro();
     const auto& value = node.get_value();
-    if (macro) {
-        macro->accept(*this);
-    } else {
-        printer->add_text(std::to_string(value));
-    }
+    printer->add_text(std::to_string(value));
 }
 
 
--- a/test/integration/mod/cabpump.mod
+++ b/test/integration/mod/cabpump.mod
@@ -53,6 +53,8 @@ DERIVATIVE state {
 	if (drive_channel <= 0.) { drive_channel = 0.  }   : cannot pump inward 
         ca' = drive_channel/18 + (cainf -ca)/taur*11
 	cai = ca
+
+    if (FOO == 0) { }
 }
 
 : to test code generation for TABLE statement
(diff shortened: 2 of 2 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -69,13 +69,8 @@ void CodegenCVisitor::visit_integer(const Integer& node) {
...
     }
-    const auto& macro = node.get_macro();
     const auto& value = node.get_value();
-    if (macro) {
-        macro->accept(*this);
-    } else {
-        printer->add_text(std::to_string(value));
-    }
+    printer->add_text(std::to_string(value));
 }
```
pramodk, line 74: Was there any bug here? 🤔 

Note that the name was printed **_intentionally_**. When we have code like:

```python
DEFINE init 1
DEFINE rise 2
DEFINE fall 3
DEFINE off 4

NET_RECEIVE(w) {
  if (flag == init) {
    WATCH (v > -55) rise
  }else if (flag == rise) {
    g = gna
    e = ena
    WATCH (v > 10) fall
  }else if (flag == fall) {
    g = gk
    e = ek
    WATCH (v < -70) off
  }else if (flag == off) {
    g = gpas
    e = erev
    WATCH (v > -55) rise
  }
}
```
Having these conceptual names like `init`, `fall`, `rise` (of voltage) is useful in the generated code rather than 1/2/3/4. Because we still look at the code during debugging.

alkino: The problem is that the macro was not defined inside the c++ file.
I will look on this side now.
This PR is only here to "discuss", don't worry ^^

pramodk: I see that you mentioned offending model + mod file. I ran test with WATCH statement
and see that what I mentioned above (as expectation) is not true i.e. DEFINE is being replaced with values in same cases. Before merging this PR, I would just verify how in certain cases we are currently replacing DEFINE with values without this PR.

alkino: We looked with @iomaganaris , and we found that for `BynaryExpression` special case, the integer is converted as parsing to a `Double` so a direct value and no more `macro`.
In the case found in `nmodldb` this is not an expression but a direct value, so we keep the meaning of `integer` that can be `direct value` or `named value`.

iomaganaris:
Could you please add various typical usages values defined using the `DEFINE` keyword in one of the mod files inside https://github.com/BlueBrain/nmodl/tree/master/test/integration/mod to make sure that the generated code is compiling correctly?

iomaganaris (approved):
LGTM

PR #768, opened 2021-11-03
Replace placeholder arg_v with the correct name in function containin…

alkino:
…g table stmt

Fixing:
```
PROCEDURE settables(b) {
    TABLE minf FROM 1 TO 2 WITH 1
    minf = 0
}
```

Checking that the following construct is not valid because `PROCEDURE` with `TABLE` statement should have exactly one parameter.
```
PROCEDURE settables(a, b) {
    TABLE <...>
    <...>
```

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -1606,12 +1606,17 @@ void CodegenCVisitor::print_table_replacement_function(const ast::Block& node) {
     print_function_declaration(node, name);
     printer->start_block();
     {
-        printer->add_line("if ( {} == 0) {}"_format(use_table_var, "{"));
-        printer->add_line("    {}({}, arg_v);"_format(function_name, internal_method_arguments()));
+        const auto& params = node.get_parameters();
+        printer->add_line("if ( {} == 0) {{"_format(use_table_var));
+        printer->add_line("    {}({}, {});"_format(function_name,
+                                                   internal_method_arguments(),
+                                                   params[0].get()->get_node_name()));
         printer->add_line("     return 0;");
         printer->add_line("}");
 
-        printer->add_line("double xi = {} * (arg_v - {});"_format(mfac_name, tmin_name));
+        printer->add_line("double xi = {} * ({} - {});"_format(mfac_name,
+                                                               params[0].get()->get_node_name(),
+                                                               tmin_name));
         printer->add_line("if (isnan(xi)) {");
         for (const auto& var: table_variables) {
             auto name = get_variable_name(var->get_node_name());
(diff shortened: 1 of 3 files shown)


src/visitors/semantic_analysis_visitor.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "ast/ast.hpp"
+#include "visitors/ast_visitor.hpp"
+
+namespace nmodl {
+namespace visitor {
+
+class SemanticAnalysisVisitor: public ConstAstVisitor {
+  private:
```
pramodk, line 10: this new visitor LGTM. May be can go in a separate PR.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1606,12 +1606,21 @@ void CodegenCVisitor::print_table_replacement_function(const ast::Block& node) {
     print_function_declaration(node, name);
     printer->start_block();
     {
-        printer->add_line("if ( {} == 0) {}"_format(use_table_var, "{"));
-        printer->add_line("    {}({}, arg_v);"_format(function_name, internal_method_arguments()));
+        const auto& params = node.get_parameters();
+        if (params.size() != 1) {
+            throw std::runtime_error(
+                "FUNCTION or PROCEDURE containing a TABLE stmt must have exactly one argument");  // Check it in parser?
+        }
```
pramodk, line 1613: nitpick - we can remove this from here as your semantic visitor will do this now. (Or could be just assert here)

.github/workflows/clang_cmake_format_check.yaml
```diff
@@ -0,0 +1,42 @@
+name: clang-cmake-format-check
+
+concurrency:
+  group: ${{ github.workflow }}#${{ github.ref }}
+  cancel-in-progress: true
+
```
pramodk, line 6: Q - Is this relevant change in this PR? Or included accidentally.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1606,12 +1606,21 @@ void CodegenCVisitor::print_table_replacement_function(const ast::Block& node) {
...
+        printer->add_line("    {}({}, {});"_format(function_name,
+                                                   internal_method_arguments(),
+                                                   params[0].get()->get_node_name()));
         printer->add_line("     return 0;");
         printer->add_line("}");
 
-        printer->add_line("double xi = {} * (arg_v - {});"_format(mfac_name, tmin_name));
+        printer->add_line("double xi = {} * ({} - {});"_format(mfac_name,
+                                                               params[0].get()->get_node_name(),
+                                                               tmin_name));
```
pramodk, line 1623: the fix LGTM!

pramodk (approved):
LGTM

iomaganaris (approved):

PR #769, opened 2021-11-04
Add a Semantic Analysis Visitor for check semantic

alkino:


src/main.cpp
```diff
@@ -316,6 +317,8 @@ int main(int argc, const char* argv[]) {
         /// just visit the ast
         AstVisitor().visit_program(*ast);
 
+        SemanticAnalysisVisitor().visit_program(*ast);
```
pramodk, line 320:
```suggestion
        logger->info("Running semantic analysis visitor");
        SemanticAnalysisVisitor().visit_program(*ast);
```

pramodk (commented):
LGTM. Would be nice to add little doxygen before merging.

src/visitors/semantic_analysis_visitor.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "ast/function_block.hpp"
+#include "ast/procedure_block.hpp"
+#include "ast/table_statement.hpp"
+
+namespace nmodl {
+namespace visitor {
+
+void SemanticAnalysisVisitor::visit_procedure_block(const ast::ProcedureBlock& node) {
+    in_procedure_function = true;
+    one_arg_in_procedure_function = node.get_parameters().size() == 1;
```
iomaganaris, line 11: Since inside this function there could be multiple checks for `ProcedureBlocks` I was wondering what would be the best way to keep each one of them kind of separated so we understand which code is performing which check. I was thinking about putting the code in separate functions but this might not be very feasible due to case where where we have checks for member variables like `in_procedure_function` or calls to `visit_children` needed to be done after or before other checks that happen inside this function. Maybe at least writing the code in code blocks and providing some more documentation in the form of comment close to the code would be a good solution? What do you think?

iomaganaris (approved):
LGTM
Just a small question/suggestion for improvement of the code structure

src/visitors/semantic_analysis_visitor.cpp
pramodk: Yeah..I was thinking of suggestions like above. e.g. instead of doing visit children, you can just check if there is a TABLE statement exist (using `collect_nodes()`). But then I thought we can leave such changes/improvements later, when we will implement other semantic checks. It would be easy to see what would be better way to re-organise.

iomaganaris: Sure. Anyway this change is good so we can improve as we add checks in this visitor

PR #772, opened 2021-11-08
Develop semantic analysis

alkino:
- Use log system instead of only throw
- Add a new check: destructor blocks only in point process type mod file

### DIFF
--- a/src/visitors/semantic_analysis_visitor.cpp
+++ b/src/visitors/semantic_analysis_visitor.cpp
@@ -30,9 +30,6 @@ void SemanticAnalysisVisitor::visit_table_statement(const ast::TableStatement&) 
     /// <-- This code is for test 1
     if (in_procedure_function && !one_arg_in_procedure_function) {
         logger->critical("SemanticAnalysisVisitor :: The procedure or function containing the TABLE statement should contains exactly one argument.");
-        if (!continue_on_critical) {
-            throw std::runtime_error("SemanticAnalysisVisitor :: The procedure or function containing the TABLE statement should contains exactly one argument.");
-        }
     }
     /// -->
 }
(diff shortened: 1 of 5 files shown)


alkino:
@iomaganaris should we throw or only send message?

iomaganaris:
> @iomaganaris should we throw or only send message?

Discussed offline. I suggested mimicking the behavior of `CodegenCompatibilityVisitor` (just printing warnings and returning 1 from the main function)

iomaganaris (commented):
LGTM in general.
Small nitpicks and suggestions

test/unit/visitor/semantic_analysis.cpp
```diff
@@ -59,8 +59,49 @@ SCENARIO("TABLE stmt", "[visitor][semantic_analysis]") {
...
+
+SCENARIO("Destructor block", "[visitor][semantic_analysis]") {
+    GIVEN("A point-process mod file, with a destructor") {
+        std::string nmodl_text = R"(
+            NEURON {
+                POINT_PROCESS test
+            }
+
+            DESTRUCTOR {
+            }
```
pramodk, line 76: adapt this test to have `DESTRUCTOR` before`NEURON` block.

pramodk (changes requested):

src/visitors/semantic_analysis_visitor.cpp
```diff
@@ -1,31 +1,64 @@
...
+}
+
+void SemanticAnalysisVisitor::visit_destructor_block(const ast::DestructorBlock& node) {
+    /// <-- This code is for check 2
+    if (!is_point_process) {
+        logger->warn(
+            "SemanticAnalysisVisitor :: This mod file is not point process but contains a "
+            "destructor.");
+        check_fail = true;
     }
```
pramodk, line 60: If we have mod file containing `DESTRUCTOR` block before `NEURON` block:

```
DESTRUCTOR {

}

NEURON {
    THREADSAFE
    :POINT_PROCESS test
    SUFFIX test
}
```

above implementation wouldn't work. I would suggest to set `is_point_process` in `::check()` method (e.g. by looking up for `Suffix` node type in the whole AST)

pramodk (approved):
LGTM

PR #773, opened 2021-11-12
Reorganise code generation for Python bindings, improve build speed

olupton:
These changes aim to make NMODL less slow to build when Python bindings are enabled (and, to a lesser extent, when they're not). One standalone change is to avoid including `spdlog` header in `utils/string_utils.hpp` just to get `fmt`, which apparently makes the build O(20%) faster 🤷

The other changes are all around code generation for Python bindings:
- e9202435626e7784cd6161189e757a51660f90d9 `all.hpp` now includes other headers instead of duplicating their content in one massive file (no particular speed difference)
- 240ac8dd5e8c2c12a30ee20d4fd929ab32505e38 uses type names explicitly in template arguments, instead of passing instances of those types around. This makes the different `py::class_` declarations independent from one another.
- c0470be08e999653c103c72bcf405e74853bdaf4 exploits this independence by splitting the generated code across more .cpp files (2 instead of 1 seems enough to make a meaningful difference)

Together, this makes NMODL compile in around 2m40s instead of 6m20s in my test setup.

### DIFF
--- a/src/language/code_generator.py
+++ b/src/language/code_generator.py
@@ -1,5 +1,5 @@
 # ***********************************************************************
-# Copyright (C) 2018-2019 Blue Brain Project
+# Copyright (C) 2018-2021 Blue Brain Project
 #
 # This file is part of NMODL distributed under the terms of the GNU
 # Lesser General Public License. See top-level LICENSE file for details.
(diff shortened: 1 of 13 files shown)


ohm314:
uh, very cool! :)

src/language/templates/pybind/pynode.cpp
```diff
@@ -0,0 +1,89 @@
+/*************************************************************************
+ * Copyright (C) 2018-2021 Blue Brain Project
+ *
+ * This file is part of NMODL distributed under the terms of the GNU
+ * Lesser General Public License. See top-level LICENSE file for details.
+ *************************************************************************/
+
+///
+/// THIS FILE IS GENERATED AT BUILD TIME AND SHALL NOT BE EDITED.
```
ohm314, line 9: I don't think this should be committed - or am I missing something?

src/language/code_generator.py
```diff
@@ -194,6 +192,23 @@ def workload(self):
                         )
                         tasks.append(task)
                         yield task
+                elif filepath == pynode_cpp_tpl:
+                    chunk_length = math.ceil(len(self.nodes) / num_pybind_files)
```
ohm314, line 196: Can you not just use the `//` operator (`+ 1`)?

src/language/code_generator.py
```diff
@@ -194,6 +192,23 @@ def workload(self):
...
+                    for n in range(num_pybind_files):
```
ohm314, line 197: `n` --> `chunk_k`

just nitpicking

ohm314 (commented):
Looks good to me. I had some minor things only.
But I would prefer someone else also take a look at it.

src/language/templates/pybind/pynode.cpp
olupton: Seems to be the precedent that we commit this text in the template files, e.g. https://github.com/BlueBrain/nmodl/blob/55e4a6f10c7843bdc1e8a2cb9d1193d2a3590a56/src/language/templates/pybind/pyast.cpp#L8-L10 too 🤷

src/language/code_generator.py
olupton: As a direct substitution, that gives a different answer if `num_pybind_files` divides `len(self.nodes)` exactly? To me it's also less expressive, but I can change it if you insist :)

src/language/code_generator.py
olupton: Done in a28a3d2 .

src/language/code_generator.py
ohm314: nope, I don't insist. It's fine.

tristan0x (approved):
Looks very good to me!
Very nice contribution @olupton

PR #774, opened 2021-11-15
azureCI: macOS-10.14 is deprecated to macOS-10.15

alkino:
See: https://devblogs.microsoft.com/devops/hosted-pipelines-image-deprecation/

### DIFF
--- a/.github/workflows/nmodl-ci.yml
+++ b/.github/workflows/nmodl-ci.yml
@@ -26,7 +26,7 @@ jobs:
     name: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-18.04, macos-10.14, ubuntu-20.04, macos-11 ]
+        os: [ ubuntu-18.04, macos-10.15, ubuntu-20.04, macos-11 ]
       fail-fast: true
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/nmodl-ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     name: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-18.04, macos-10.15, ubuntu-20.04, macos-11 ]
+        os: [ ubuntu-18.04, macos-10.14, ubuntu-20.04, macos-11 ]
```
olupton, line 29: Is this documented anywhere? It doesn't seem to be listed here: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

alkino:
Ahah I tried like that ^^ without hope ^^

SHould we totally delete support of macos 10.14?

.github/workflows/nmodl-ci.yml
pramodk: May be worth verifying : **macos-10.14 image is using macos-10.15 #2347** : https://github.com/actions/virtual-environments/issues/2347

If there is no option to use 10.14, we can let it go!

olupton (approved):

PR #775, opened 2021-11-16
Forbid PARAMETER if it is written and not a range

alkino:
Example of bad design:

```
NEURON {
  POINT_PROCESS two_exp_syn1
  RANGE onset
}

PARAMETER {
  onset = 0 (ms)
  T
}

INITIAL {
  T = -log(1/onset)
}
```

Here `T` is a `ast::ParamAssign` and is `GLOBAL` as not defined `RANGE` in `NEURON` block.
So `T` becomes a `GLOBAL`, but we try to write it in `INITIAL`.

As user can modify `PARAMETER` it leads to possible non-thread safe code.

Before this patch, when compiling the generated `cpp` file, it leads to several :
`x86_64/corenrn/mod2c/two_exp.cpp:96:28: error: ‘T’ was not declared in this scope`

After this patch, the `cpp` file is not generated and **nmodl** give an error:
`[NMODL] [error] :: Code Incompatibility :: "T" variable found at [UNKNOWN] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations`

We don't fix it, we disable it by design.

### DIFF
--- a/src/codegen/codegen_compatibility_visitor.cpp
+++ b/src/codegen/codegen_compatibility_visitor.cpp
@@ -18,6 +18,8 @@ using namespace fmt::literals;
 namespace nmodl {
 namespace codegen {
 
+using symtab::syminfo::NmodlType;
+
 const std::map<ast::AstNodeType, CodegenCompatibilityVisitor::FunctionPointer>
     CodegenCompatibilityVisitor::unhandled_ast_types_func(
         {{AstNodeType::MATCH_BLOCK,
@@ -37,6 +39,7 @@ const std::map<ast::AstNodeType, CodegenCompatibilityVisitor::FunctionPointer>
          {AstNodeType::SOLVE_BLOCK,
           &CodegenCompatibilityVisitor::return_error_if_solve_method_is_unhandled},
          {AstNodeType::GLOBAL_VAR, &CodegenCompatibilityVisitor::return_error_global_var},
+         {AstNodeType::PARAM_ASSIGN, &CodegenCompatibilityVisitor::return_error_param_var},
          {AstNodeType::POINTER_VAR, &CodegenCompatibilityVisitor::return_error_pointer},
          {AstNodeType::BBCORE_POINTER_VAR,
           &CodegenCompatibilityVisitor::return_error_if_no_bbcore_read_write}});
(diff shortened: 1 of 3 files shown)


alkino:
So, the check is current ` has_property(PARAM_ASSIGN) && ! has_property(RANGE) && write_count > 0`.

Should we check for all type of variable? `read_ion` / `write_ion` and others that I still ignore it exists.

src/codegen/codegen_compatibility_visitor.cpp
```diff
@@ -73,6 +76,20 @@ std::string CodegenCompatibilityVisitor::return_error_global_var(
...
+std::string CodegenCompatibilityVisitor::return_error_param_var(
+    ast::Ast& node,
+    const std::shared_ptr<ast::Ast>& ast_node) {
+    auto param_assign = std::dynamic_pointer_cast<ast::ParamAssign>(ast_node);
+    std::stringstream error_message_global_var;
+    auto symbol = node.get_symbol_table()->lookup(param_assign->get_node_name());
+    if (!symbol->has_any_property(NmodlType::range_var) && symbol->get_write_count() > 0) {
+        error_message_global_var
+            << "\"{}\" variable found at [{}] should be defined as a RANGE variable instead of GLOBAL to enable backend transformations\n"_format(
+                   symbol->get_name(), symbol->get_token().position());
```
pramodk, line 88: `nitpick`: I am curious why description says `found at [UNKNOWN]` i.e. position information for the `T` is not available.

pramodk (approved):
@alkino : LGTM and adding simple unit tests would be great!

src/codegen/codegen_compatibility_visitor.cpp
alkino: I started looking at why, but it's quite hard. It comes from the lexer.

pramodk: Ok, it could be ignored here.

src/symtab/symbol.hpp
```diff
@@ -348,6 +348,11 @@ class Symbol {
     bool is_variable() const noexcept;
 
     std::string to_string() const;
+
+    bool is_writable() const noexcept {
+        return has_any_property(syminfo::NmodlType::range_var) ||
+               has_any_property(syminfo::NmodlType::write_ion_var);
+    }
```
pramodk, line 355: @alkino :  
* `ASSIGNED` variables (`assigned_definition`) are also behave as a range by default. 
* Same thing for state variables
* Technically, IIRC, read ion variable can be also can be written by user (I think there exist some mod files
You can see the comments in the code here: https://github.com/BlueBrain/nmodl/blob/abdbecf364ea65aa2347cd3c6363691f7c6ea022/src/codegen/codegen_helper_visitor.cpp#L373

iomaganaris (approved):
LGTM

olupton:
Just to note that one of the mechanisms included in NEURON falls foul of this: https://github.com/neuronsimulator/nrn/blob/514d0b62b478e1245f0ea18d020e955362f223f1/test/rxd/ecs/pump.mod

nrnhines:
> one of the mechanisms included in NEURON falls foul of this

That's clearly a mistake. The values should be specified in the PARAMETER block instead of the INITIAL block, then they will be modifiable from the interpreter. I'll mention it to Adam.

nrnhines:
> one of the mechanisms included in NEURON falls foul of this

neuronsimulator/nrn#1607 fixes those.

PR #778, opened 2021-11-18
Cast index of array to int

### LINKED ISSUE #776: ASSIGNED variable are double and so cannot be use as array index
Sometime people wrote mod files:

```
NEURON {
  RANGE PRE
}

ASSIGNED {
  PRE[50]
  numpulses = 0
}

NET_RECEIVE() {
  PRE[numpulses] = numpulses
  numpulses = numpulses + 1
}
```

what should we do with it?

Try to find when declare them `int` instead of `double`, or forbid this kind of mod files?

With **mod2c** we got a c-casting `(int) numpulses` when mandatory.

alkino:
Sometimes variables are double, so cast them to int, to have valid c.

Fix #776

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -133,7 +133,7 @@ void CodegenCVisitor::visit_var_name(const VarName& node) {
     }
     if (index) {
         printer->add_text("[");
-        printer->add_text("((int)");
+        printer->add_text("static_cast<int>(");
         index->accept(*this);
         printer->add_text(")");
         printer->add_text("]");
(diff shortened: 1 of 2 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -133,7 +133,9 @@ void CodegenCVisitor::visit_var_name(const VarName& node) {
     }
     if (index) {
         printer->add_text("[");
+        printer->add_text("((int)");
```
olupton, line 136:
```suggestion
        printer->add_text("static_cast<int>(");
```
[ES.49: If you must use a cast, use a named cast](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts-named).

src/codegen/codegen_c_visitor.cpp
```diff
@@ -145,7 +147,9 @@ void CodegenCVisitor::visit_indexed_name(const IndexedName& node) {
     }
     node.get_name()->accept(*this);
     printer->add_text("[");
+    printer->add_text("((int)");
```
olupton, line 150:
```suggestion
    printer->add_text("static_cast<int>(");
```

olupton (approved):

PR #779, opened 2021-11-24
Enable local variable rename visitor by default

### LINKED ISSUE #777: Should avoid renaming if variables are shadowed
In this example:
```
NEURON {
    RANGE minf
}

ASSIGNED {
    minf
}

DERIVATE states {
    LOCAL minf
    minf = 1 / exp(2)
}
```
is currently converted to
```c++
void nrn_state_kA(...) {
    double inst->minf[id];
    inst->minf[id] = 1.0 / exp(2.0)
}
```

We should avoid renaming if we have local shadowing

Today, it works with `passes --local-rename` option.

It happens on the model `226010` of **nmodldb**

**mod2c** scope variable like that
```c++
void nrn_state_kA() {
  {
    double _lminf;
    _lminf = 1.0 / exp(2.0)
  }
}
```
And does not rename them

alkino:
It has been tested that it works too for variable inside VERBATIM block

Fix #777

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -117,7 +117,7 @@ int main(int argc, const char* argv[]) {
     bool localize_verbatim(false);
 
     /// true if local variables to be renamed
-    bool local_rename(false);
+    bool local_rename(true);
 
     /// true if inline even if verbatim block exist
     bool verbatim_inline(false);


pramodk (approved):
LGTM (if all CI passes).

PR #780, opened 2021-12-01
Use nrn_pragma_acc(...) macro and add nrn_pragma_omp(...) alternatives.

olupton:
Instead of directly printing
```c++
#pragma acc ...
```
we now print pairs of macros
```c++
nrn_pragma_acc(...)
nrn_pragma_omp(...)
```
so the translated sources can be compiled both with OpenACC and with OpenMP target offload.

The current OpenMP implementation does not include any `depend` or `nowait` clauses, as an early attempt to do this caused crashes with NVHPC 21.9. This will be revisited later.


pramodk (approved):
LGTM

pramodk (approved):

PR #781, opened 2021-12-06
Clang warnings

tristan0x:


olupton:
If you also enable relevant warnings in https://github.com/BlueBrain/nmodl/blob/4a67187535eb6d79d7ef74095dacd6434948a43d/.github/workflows/nmodl-ci.yml#L73 then regressions should be flagged in the diffs of PRs (see https://github.com/BlueBrain/nmodl/pull/751).

tristan0x:
> If you also enable relevant warnings in
> 
> https://github.com/BlueBrain/nmodl/blob/4a67187535eb6d79d7ef74095dacd6434948a43d/.github/workflows/nmodl-ci.yml#L73
> 
> then regressions should be flagged in the diffs of PRs (see #751).

Are you saying I should register the new flags in `nmodl/.github/workflows/nmodl-ci.yml`?
I have clang specific flags to add but the diff portion you sent is for both GCC and Clang, right?

olupton:
I don't mind exactly how it's done, maybe you can just pass your new `NMODL_PEDANTIC` option there, as long as it doesn't add `-Werror`.

You're right that the CMake command that I linked to will be run for both Linux/GCC and macOS/Clang.

1uc:
Too old.

PR #783, opened 2021-12-13
Integrate changes from NERSC GPU hackathon.

olupton:
See https://github.com/BlueBrain/CoreNeuron/pull/713 for context.

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -147,7 +147,7 @@ void CodegenAccVisitor::print_eigen_linear_solver(const std::string& float_type,
     if (N <= 4) {
         printer->add_line("{0} = {1}.inverse()*{2};"_format(Xm, Jm, Fm));
     } else {
-        printer->add_line("{0} = partialPivLu<{1}>({2}, {3});"_format(Xm, N, Jm, Fm));
+        printer->add_line("{0} = partialPivLu{1}({2}, {3});"_format(Xm, N, Jm, Fm));
     }
 }
 
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -296,10 +296,12 @@ void CodegenAccVisitor::print_newtonspace_transfer_to_device() const {
 
 
 void CodegenAccVisitor::print_instance_variable_transfer_to_device() const {
-    printer->start_block("if(nt->compute_gpu)");
-    printer->add_line("Memb_list* dml = cnrn_target_deviceptr(ml);");
-    printer->add_line("cnrn_target_memcpy_to_device(&(dml->instance), &(ml->instance));");
-    printer->end_block(1);
+    if (!info.artificial_cell) {
+        printer->start_block("if(nt->compute_gpu)");
+        printer->add_line("Memb_list* dml = cnrn_target_deviceptr(ml);");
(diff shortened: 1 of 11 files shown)


iomaganaris (approved):
LGTM

alkino (approved):

src/codegen/codegen_acc_visitor.cpp
```diff
@@ -265,53 +275,59 @@ std::string CodegenAccVisitor::get_variable_device_pointer(const std::string& va
...
-    printer->add_line("    auto device_vec = static_cast<double*>(acc_copyin(vec, vec_size));");
-    printer->add_line("    auto device_ns = static_cast<NewtonSpace*>(acc_deviceptr(*ns));");
-    printer->add_line("    auto device_thread = static_cast<ThreadDatum*>(acc_deviceptr(thread));");
+    printer->start_block("if(nt->compute_gpu)");
+    printer->add_line("double* device_vec = cnrn_target_copyin(vec, vec_size / sizeof(double));");
+    printer->add_line("void* device_ns = cnrn_target_deviceptr(*ns);");
+    printer->add_line("ThreadDatum* device_thread = cnrn_target_deviceptr(thread);");
     printer->add_line(
-        "    acc_memcpy_to_device(&(device_thread[{}]._pvoid), &device_ns, sizeof(void*));"_format(
+        "cnrn_target_memcpy_to_device(&(device_thread[{}]._pvoid), &device_ns);"_format(
```
pramodk, line 289: just curious - was auto deduced type was different here?

src/codegen/codegen_acc_visitor.cpp
```diff
@@ -265,53 +275,59 @@ std::string CodegenAccVisitor::get_variable_device_pointer(const std::string& va
...
 void CodegenAccVisitor::print_newtonspace_transfer_to_device() const {
     int list_num = info.derivimplicit_list_num;
-    printer->add_line("if (nt->compute_gpu) {");
-    printer->add_line("    auto device_vec = static_cast<double*>(acc_copyin(vec, vec_size));");
-    printer->add_line("    auto device_ns = static_cast<NewtonSpace*>(acc_deviceptr(*ns));");
-    printer->add_line("    auto device_thread = static_cast<ThreadDatum*>(acc_deviceptr(thread));");
+    printer->start_block("if(nt->compute_gpu)");
+    printer->add_line("double* device_vec = cnrn_target_copyin(vec, vec_size / sizeof(double));");
+    printer->add_line("void* device_ns = cnrn_target_deviceptr(*ns);");
+    printer->add_line("ThreadDatum* device_thread = cnrn_target_deviceptr(thread);");
```
pramodk, line 287: just curious - was auto deduced type was different here?

pramodk (approved):
LGTM 👌

src/codegen/codegen_acc_visitor.cpp
olupton: I think `auto` or `auto*` would also be fine.
I believe I just left the type in so it was still spelled out once on the line.

src/codegen/codegen_acc_visitor.cpp
olupton: I think for this one it was important to have `device_ns` be `void*` rather than `NewtonSpace*`.

This is because now we have
```c++
template <typename T>
void cnrn_target_memcpy_to_device(T*, T*) { ... }
```
and if you pass `void**` and `NewtonSpace**` it will fail to deduce `T`. We have to pass two pointers of the same type (up to const qualification).

PR #784, opened 2021-12-14
Fix NVHPC + OpenMP ~ OpenACC compilation

olupton:
`#pragma acc routine seq` seems to play better with templates than `#pragma omp declare target` does. Here is a hideous workaround...


olupton:
This is being tested with https://github.com/BlueBrain/CoreNeuron/pull/717.

iomaganaris (approved):
Oh god :man_facepalming:

pramodk (approved):
LGTM

PR #788, opened 2021-12-21
Changes for async execution of OpenACC and OpenMP

iomaganaris:
Follows changes from https://github.com/BlueBrain/CoreNeuron/pull/725

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -319,7 +319,7 @@ void CodegenAccVisitor::print_device_atomic_capture_annotation() const {
 
 void CodegenAccVisitor::print_device_stream_wait() const {
     printer->start_block("if(nt->compute_gpu)");
-    printer->add_line("nrn_pragma_acc(wait(nt->streams[nt->stream_id]))");
+    printer->add_line("nrn_pragma_acc(wait async(nt->streams[nt->stream_id]))");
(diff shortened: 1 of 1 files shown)


src/codegen/codegen_acc_visitor.cpp
```diff
@@ -265,53 +275,57 @@ std::string CodegenAccVisitor::get_variable_device_pointer(const std::string& va
...
-    printer->add_line("#pragma acc atomic capture");
+    printer->add_line("nrn_pragma_acc(atomic capture)");
+    printer->add_line("nrn_pragma_omp(atomic capture)");
 }
 
 
 void CodegenAccVisitor::print_device_stream_wait() const {
-    printer->add_line("#pragma acc wait(nt->stream_id)");
+    printer->start_block("if(nt->compute_gpu)");
+    printer->add_line("nrn_pragma_acc(wait async(nt->streams[nt->stream_id]))");
```
pramodk, line 320: @iomaganaris / @olupton: does`acc wait async(stream_id)` change any execution semantics here compared to previous `acc wait (stream_id)` ? 

> from what I read in the nvforum - “wait async” is used to create dependencies between different async queues. So if your parallel regions had different queue numbers, you add “wait(Q1) async(Q2)” where Q2 should wait for Q1 to finish before running.

1uc:
Timed out.

PR #791, opened 2021-12-23
Bug fix: Parameter updated in VERBATIM block shouldn't be const

pramodk:
* if PARAMETER variable of RANGE type is updated in VERBATIM
   block then it's write_count could be 0 even though it's
   updated in VERBATIM block
   - such variable can not be declared as `const` instance variable
   - one such example is https://github.com/nrnhines/tqperf/blob/master/mod/invlfire.mod
 * Codegen helper visitor now keep track of all symbols that
   are used in all verbatim blocks and the codegen visitor
   check this list before deciding if variable can be const or not.
   - note that variable could be read-only in verbatim block but
     currently we don't have robut C code analysis capability
     and it's not worth (yet)
 * this issue was encountered in https://github.com/BlueBrain/CoreNeuron/pull/701

### DIFF
--- a/test/unit/codegen/codegen_c_visitor.cpp
+++ b/test/unit/codegen/codegen_c_visitor.cpp
@@ -9,6 +9,7 @@
 
 #include "ast/program.hpp"
 #include "codegen/codegen_c_visitor.hpp"
+#include "codegen/codegen_helper_visitor.hpp"
 #include "parser/nmodl_driver.hpp"
 #include "test/unit/utils/test_utils.hpp"
 #include "visitors/symtab_visitor.hpp"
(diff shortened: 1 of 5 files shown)


test/unit/codegen/codegen_c_visitor.cpp
```diff
@@ -222,3 +222,48 @@ SCENARIO("Check instance variable definition order", "[codegen][var_order]") {
...
+            }
+        )";
+
+        THEN("Variable used in VERBATIM shouldn't be marked as const") {
+            std::stringstream ss;
+
+            /// parse mod file & print mechanism structure
+            const auto& ast = NmodlDriver().parse_string(nmodl_text);
+            auto cvisitor = create_c_visitor(ast, nmodl_text, ss);
+            cvisitor->print_mechanism_range_var_structure();
```
olupton, line 255: I think this is fine for now. Maybe if this file gets too repetitive then we could try something like
```c++
auto const reindented_generated_code = invoke_with_c_visitor([](auto& cvisitor) {
  cvisitor->print_mechanism_range_var_structure();
});
```
which would also handle the lifetime issues.

olupton (approved):

PR #798, opened 2022-01-12
Add a python binding for the parent pointer

ohm314:
The getter and setter methods are bound to a pybind11 property.


iomaganaris (approved):
LGTM :+1: 
Only suggestion is whether there can be a test added that uses this

PR #799, opened 2022-01-18
Support for BEFORE/AFTER blocks in code generation

pramodk:
* separate functions are generated for each before/after
       block just like other global functions init/state/current
* functions are then registed with coreneuron using `hoc_reg` api
* this will help to close BlueBrain/CoreNeuron/pull/748 and  neuronsimulator/nrn#1581
* there is no unit test but the code generation changes will be validated by integration test neuronsimulator/testcorenrn#8

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -2124,7 +2124,7 @@ std::string CodegenCVisitor::conc_write_statement(const std::string& ion_name,
  * case we first update current mechanism's shadow vector and then add statement
  * to queue that will be used in reduction queue.
  */
-std::string CodegenCVisitor::process_shadow_update_statement(ShadowUseStatement& statement,
+std::string CodegenCVisitor::process_shadow_update_statement(const ShadowUseStatement& statement,
                                                              BlockType type) {
     // when there is no operator or rhs then that statement doesn't need shadow update
     if (statement.op.empty() && statement.rhs.empty()) {
@@ -2926,13 +2926,8 @@ void CodegenCVisitor::print_mechanism_register() {
 
     /// register all before/after blocks
     for (size_t i = 0; i < info.before_after_blocks.size(); i++) {
-        // \todo: mod2c/neuron implementation does some ion variable
-        // manipulation which is not clear and needs to be discussed/understood
-        // with Michael Hines:
-        // https://github.com/BlueBrain/mod2c/blob/4898ef45064804f9c7815765721d2b23b67e40b3/src/mod2c_core/nocpout.c#L1478
-
-        const auto& block = info.before_after_blocks[i];
         // register type and associated function name for the block
+        const auto& block = info.before_after_blocks[i];
         size_t register_type = get_register_type_for_ba_block(block);
         std::string function_name("nrn_before_after_{}"_format(i));
         printer->add_line("hoc_reg_ba(mech_type, {}, {});"_format(function_name, register_type));
@@ -3589,6 +3584,12 @@ void CodegenCVisitor::print_before_after_block(const ast::Block* node, size_t bl
     printer->add_line("double v = voltage[node_id];");
     print_v_unused();
 
+    // read ion statements
+    const auto& read_statements = ion_read_statements(BlockType::Equation);
+    for (auto& statement: read_statements) {
+        printer->add_line(statement);
+    }
+
     /// print main body
     std::shared_ptr<ast::BABlock> ba_block;
     if (node->is_before_block()) {
@@ -3599,6 +3600,13 @@ void CodegenCVisitor::print_before_after_block(const ast::Block* node, size_t bl
     printer->add_indent();
     ba_block->get_statement_block()->visit_children(*this);
     printer->add_newline();
+
+    // write ion statements
+    const auto& write_statements = ion_write_statements(BlockType::Equation);
+    for (auto& statement: write_statements) {
+        auto text = process_shadow_update_statement(statement, BlockType::Equation);
+        printer->add_line(text);
+    }
 
     /// loop end including data annotation block
     print_channel_iteration_block_end();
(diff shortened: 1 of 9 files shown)


nrnhines:
> there is no unit test but

My plan for neuronsimulator/nrn#1581 is to add a test that covers all the BEFORE/AFTER combinations for INITIAL, BREAKPOINT, SOLVE (and STEP)

nrnhines:
Note that the only BEFORE/AFTER combinations supported in classical NEURON are
```
nrnoc/fadvance.cpp:	nrn_ba(nth, AFTER_SOLVE);
nrnoc/fadvance.cpp:        nrn_ba(nrn_threads + i, BEFORE_INITIAL);
nrnoc/fadvance.cpp:        nrn_ba(nrn_threads + i, AFTER_INITIAL);
nrnoc/treeset.cpp:	nrn_ba(_nt, BEFORE_BREAKPOINT);
```
and
```
nrncvode/netcvode.cpp:	nrn_ba(nt, BEFORE_STEP);
```
I'm modifying the t9.mod file in nrntest/fast and that is a work in progress. But test.py
```
from neuron import h
s = h.Section()
s.insert("t9")
print("calling finitialize")
h.finitialize()
print("calling fadvance")
h.fadvance()
```
is printing
```
$ python test.py
calling finitialize
BEFORE INITIAL 1 20  0 0 t=0
INITIAL 1 21 0 0 t=0
AFTER INITIAL 1 22 0 0 t=0
BEFORE BREAKPOINT 1 24 1 0 t=0
BEFORE STEP t=0
calling fadvance
BEFORE BREAKPOINT 1 24 2 0 t=0.0125
PROCEDURE p (from SOLVE statement in BREAKPOINT) t=0.025
BREAKPOINT this model computes no current t=0.025
AFTER SOLVE 1 25 2 1 t=0.025
BEFORE STEP t=0.025
```
A proper test will need two mod files: with and without computation of current.

pramodk:
@nrnhines: do you plan to create test soon? Otherwise we can merge this PR as it passes existing modified Gfluct3.mod test in testcorenrn repository. And then we will have substantive test, we can check if any changes required in this nmodl.

nrnhines:
@pramodk 
> do you plan to create test soon

A few days. I'll merge now so it becomes easier to investigate the HNN model

nrnhines:
@pramodk 
A test of BEFORE and AFTER has been added to neuronsimulator/nrn/#1581

pramodk:
excellent! that helped me to catch some (easier) issues, I pushed 6d1af176320825405eb0353646c6bf22546ddbf4 to fix those. Test added in neuronsimulator/nrn/pull/1581 also passes using NMODL with this PR.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -2737,6 +2737,45 @@ void CodegenCVisitor::print_global_variables_for_hoc() {
...
 }
 
+/**
+ * Return registration type for a given BEFORE/AFTER block
+ * /param block A BEFORE/AFTER block being registered
+ *
+ * Depending on a block type i.e. BEFORE or AFTER and also type
+ * of it's associated block i.e. BREAKPOINT, INITIAL, SOLVE and
+ * STEP, the registration type (as an integer) is calculated.
+ * These values are then interpreted by CoreNEURON internally.
```
olupton, line 2747: I guess this is OK for now, but in general it would be nicer if we generated more expressive code, e.g.
```c++
BlockType::BEFORE | BlockType::SOLVE
```
instead of `12`, and made sure that that compiled to the right thing on the [Core]NEURON side.

olupton (approved):
Code changes look reasonable to me

src/codegen/codegen_c_visitor.cpp
iomaganaris: I would create an issue about that and merge this PR if that's okay with you

iomaganaris (approved):
LGTM

src/codegen/codegen_c_visitor.cpp
iomaganaris: Done in: https://github.com/BlueBrain/nmodl/issues/802

pramodk: > I guess this is OK for now, but in general it would be nicer if we generated more expressive code,

good point!  Just to be sure, you meant hard-coded values of 10/20/1,2.... could be replaced with the enum usage like below:
```cpp
+/*
  * Depending on a block type i.e. BEFORE or AFTER and also type
  * of it's associated block i.e. BREAKPOINT, INITIAL, SOLVE and
  * STEP, the registration type (as an integer) is calculated.
  * These values are then interpreted by CoreNEURON internally.
  */
+enum BABlockTypeValues {
+    BREAKPOINT = 1,
+    SOLVE = 2,
+    INITIAL = 3,
+    STEP = 4,
+    BEFORE = 10,
+    AFTER = 20,
+};
+
+/**
+ * Return registration type for a given BEFORE/AFTER block
+ * /param block A BEFORE/AFTER block being registered
+ */
 static size_t get_register_type_for_ba_block(const ast::Block* block) {
     size_t register_type = 0;
     BAType ba_type;
     /// before block have value 10 and after block 20
     if (block->is_before_block()) {
-        register_type = 10;
+        register_type = BABlockTypeValues::BEFORE;
         ba_type =
             dynamic_cast<const ast::BeforeBlock*>(block)->get_bablock()->get_type()->get_value();
     } else {
-        register_type = 20;
+        register_type = BABlockTypeValues::AFTER;
...
```
that's all?

olupton: No; I meant that instead of NMODL generating code with magic numbers, it (c|sh)ould generate code with names. Then the mapping from names to numbers can happen in one place ([Core]NEURON) instead of being replicated in two places.

pramodk: Ah ok. I didn’t think much about it as changing API with mod2c in the middle is headache.

PR #800, opened 2022-01-18
Fixing \r as return char inside comments

alkino:
If a single line comment in a mod file is end by only \r (not \r\n or \n),
currently it breaks because until the next \r\n or \n everything is inside
the comment


iomaganaris (approved):
Line breaks are nasty things :stuck_out_tongue: 
LGTM :+1:

PR #804, opened 2022-01-27
Use non linear newton solver for sparse method

### LINKED ISSUE #801: Invalid sparse method
In this mod file: https://github.com/BlueBrain/nmodldb/blob/downloader/models/db/modeldb/239421/mod/cdp5.mod

We try to solve with the `sparse` method with the sympy solver but:
```
[NMODL] [warning] :: SympySolverVisitor :: solve_lin_system python exception: nonlinear term encountered: -BTC*b1*ca*dsqvol*dt/(diam**2*vrat)
```
And so, next the NeuronSolver but, `sparse` method is not yet supported with it in nmodl.

```
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
```

alkino:
It is possible to give non linear system to sparse solver.
The sparse method was mistaken using linear solver, now use deriv implicit.

Fix #801

### DIFF
--- a/src/visitors/sympy_solver_visitor.cpp
+++ b/src/visitors/sympy_solver_visitor.cpp
@@ -555,9 +555,8 @@ void SympySolverVisitor::visit_derivative_block(ast::DerivativeBlock& node) {
             }
         }
 
-        if (solve_method == codegen::naming::SPARSE_METHOD) {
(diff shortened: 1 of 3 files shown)


cattabiani (approved):
LGTM!

PR #806, opened 2022-02-02
call get_node_name only on nodes that get it

alkino:
When a function is called sometimes we need to have nt first, if the user did not add it, we do it for him.
The current process was broken because expression such as `a+b` does not have `get_node_name` function.

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -1412,12 +1412,15 @@ void CodegenCVisitor::print_function_call(const FunctionCall& node) {
     auto arguments = node.get_arguments();
     printer->add_text("{}("_format(function_name));
 
-    auto possible_nt_variable_types = arguments.front()->get_node_type() == AstNodeType::NAME &&
-                                      arguments.front()->get_node_type() == AstNodeType::STRING &&
-                                      arguments.front()->get_node_type() ==
-                                          AstNodeType::CONSTANT_VAR &&
-                                      arguments.front()->get_node_type() == AstNodeType::VAR_NAME &&
-                                      arguments.front()->get_node_type() == AstNodeType::LOCAL_VAR;
+    bool possible_nt_variable_types;
+    if (!defined_method(name)) {
+        possible_nt_variable_types = arguments.front()->get_node_type() == AstNodeType::NAME &&
+                                     arguments.front()->get_node_type() == AstNodeType::STRING &&
+                                     arguments.front()->get_node_type() ==
+                                         AstNodeType::CONSTANT_VAR &&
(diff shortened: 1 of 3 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -1418,7 +1418,15 @@ void CodegenCVisitor::print_function_call(const FunctionCall& node) {
             printer->add_text(", ");
         }
     } else if (nmodl::details::needs_neuron_thread_first_arg(function_name) &&
-               arguments.front()->get_node_name() != "nt") {
+               (arguments.front()->get_node_type() != AstNodeType::NAME &&
+                    arguments.front()->get_node_type() != AstNodeType::STRING &&
+                    arguments.front()->get_node_type() != AstNodeType::CONSTANT_VAR &&
+                    arguments.front()->get_node_type() != AstNodeType::VAR_NAME &&
+                    arguments.front()->get_node_type() != AstNodeType::LOCAL_VAR ||
+                arguments.front()->get_node_name() != "nt")) {
```
iomaganaris, line 1426:
```suggestion
               ((arguments.front()->get_node_type() == AstNodeType::NAME ||
                    arguments.front()->get_node_type() == AstNodeType::STRING ||
                    arguments.front()->get_node_type() == AstNodeType::CONSTANT_VAR ||
                    arguments.front()->get_node_type() == AstNodeType::VAR_NAME ||
                    arguments.front()->get_node_type() == AstNodeType::LOCAL_VAR) &&
                arguments.front()->get_node_name() != "nt")) {
```
I think this is a bit more understandable like this?
Also maybe you could set all the checks for the `AstNodeType`s in a variable to make it more clear?
```
auto possible_nt_variable_types = arguments.front()->get_node_type() == AstNodeType::NAME ||
                    arguments.front()->get_node_type() == AstNodeType::STRING ||
                    arguments.front()->get_node_type() == AstNodeType::CONSTANT_VAR ||
                    arguments.front()->get_node_type() == AstNodeType::VAR_NAME ||
                    arguments.front()->get_node_type() == AstNodeType::LOCAL_VAR;
```

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1412,13 +1412,25 @@ void CodegenCVisitor::print_function_call(const FunctionCall& node) {
     auto arguments = node.get_arguments();
     printer->add_text("{}("_format(function_name));
 
+    bool possible_nt_variable_types;
```
iomaganaris, line 1415:
```suggestion
    bool possible_nt_variable_types{false};
```
Maybe better to avoid the issue reported by the GitHub Actions bot?

iomaganaris (approved):
LGTM
There might be an issue with the extra commit about the `POINTER` error but I think it should be fine when merged to `master`

PR #807, opened 2022-02-02
Launch CI to show that it fails for doc live-debug-docs

alkino:
The problem is that doctest try to compile documentation of `pkg_resources` package. We should block it.


alkino:
This patch introduce new `yield_lines` and doc seems not buildable by our tool: https://github.com/jaraco/jaraco.text/commit/e2cd8bbbf369957cd3489747ffd31f9205e653ea

PR #812, opened 2022-03-08
Fixup various CI plans.

olupton:
- Drop `:imported-members:`, which seems to cause crashes. It's currently unclear to me if this is compromising the generated documentation, and if so how much, but it seems worthwhile to unblock the CI. https://github.com/BlueBrain/nmodl/issues/814 for follow up.
- Follow changes to `hpc/gitlab-pipelines` helpers.
- Install `ninja` in some more CI pipelines.
- Relax version constraints.

Closes https://github.com/BlueBrain/nmodl/pull/807.

### DIFF
--- a/.github/workflows/nmodl-doc.yml
+++ b/.github/workflows/nmodl-doc.yml
@@ -16,7 +16,7 @@ on:
 
 env:
   BUILD_TYPE: Release
-  DEFAULT_PY_VERSION: 3.9
+  DEFAULT_PY_VERSION: 3.8
   DESIRED_CMAKE_VERSION: 3.15.0
 
 jobs:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -52,6 +52,13 @@ def __getitem__(self, item):
 def get_sphinx_command():
     """Lazy load of Sphinx distutils command class
     """
+    # If nbconvert is installed to .eggs on the fly when running setup.py then
+    # templates from it will not be found. This is a workaround.
+    if 'JUPYTER_PATH' not in os.environ:
+        import nbconvert
+        os.environ['JUPYTER_PATH'] = os.path.realpath(os.path.join(os.path.dirname(nbconvert.__file__), '..', 'share', 'jupyter'))
+        print("Setting JUPYTER_PATH={}".format(os.environ['JUPYTER_PATH']))
```
pramodk, line 60: `note`: Are these "temporary" workarounds for newer version or permanent fixes? if it's temporary and issue exist somewhere, link would be nice.

pramodk (approved):
LGTM

setup.py
olupton: My impression is that they changed the logic here around 6.0.0 and that it is a bit fragile, particularly when it is not used as a "system" install (e.g. Spack, with packages in their own prefixes, and with the Python stuff that also installs dependencies on the fly to separate prefixes under `.eggs/`). I came across https://github.com/spack/spack/pull/21827, for example. I did not find any particularly good reference though...

pramodk: ok, thanks for clarification.

PR #815, opened 2022-03-10
[LLVM] Allocate InstanceStruct on the GPU using cudaMallocManaged

iomaganaris:
- Enable CUDA backend with `NMODL_ENABLE_LLVM_CUDA` cmake option
- Updated minimum required `cmake` version to `3.17` to accommodate linking with CUDA libraries with ease


pramodk (approved):
LGTM

PR #817, opened 2022-03-12
[LLVM][GPU] Separated CPU and GPU CLI options

georgemitenkov:
Now, CLI has two options: `cpu` and `gpu` that allow
users to target different platforms. For example,

```
bin/nmodl mod/test.mod -o out llvm --ir

bin/nmodl mod/test.mod -o out llvm --ir cpu --name skylake --vector-width 2

bin/nmodl mod/test.mod -o out llvm --ir gpu --name cuda
```

Moreover, `assume_no_alias` option was dropped and
made default (it didn't affect the computation in
our experiments).

The new CLI looks like:
```
llvm
  LLVM code generation option
  Options:
    --ir REQUIRED                         Generate LLVM IR (false)
    --no-debug                            Disable debug information (false)
    --opt-level-ir INT:{0,1,2,3}          LLVM IR optimisation level (O0)
    --single-precision                    Use single precision floating-point types (false)
    --fmf TEXT:{afn,arcp,contract,ninf,nnan,nsz,reassoc,fast} ...
                                          Fast math flags for floating-point optimizations (none)

cpu
  LLVM CPU option
  Options:
    --name TEXT                           Name of CPU platform to use
    --math-library TEXT:{Accelerate,libmvec,libsystem_m,MASSV,SLEEF,SVML,none}
                                          Math library for SIMD code generation (none)
    --vector-width INT                    Explicit vectorization width for IR generation (1)

gpu
  LLVM GPU option
  Options:
    --name TEXT                           Name of GPU platform to use
    --math-library TEXT:{libdevice}       Math library for GPU code generation (none)

benchmark
  LLVM benchmark option
  Options:
    --run                                 Run LLVM benchmark (false)
    --opt-level-codegen INT:{0,1,2,3}     Machine code optimisation level (O0)
    --libs TEXT:FILE ...                  Shared libraries to link IR against
    --instance-size INT                   Instance struct size (10000)
    --repeat INT                          Number of experiments for benchmarking (100)
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -176,26 +176,29 @@ int main(int argc, const char* argv[]) {
     /// use single precision floating-point types
     bool llvm_float_type(false);
 
-    /// llvm vector width
-    int llvm_vec_width = 1;
+    /// optimisation level for IR generation
+    int llvm_opt_level_ir = 0;
 
-    /// vector library name
-    std::string vector_library("none");
+    /// math library name
+    std::string llvm_math_library("none");
 
     /// disable debug information generation for the IR
-    bool disable_debug_information(false);
+    bool llvm_no_debug(false);
 
     /// fast math flags for LLVM backend
     std::vector<std::string> llvm_fast_math_flags;
 
+    /// traget CPU platform name
+    std::string llvm_cpu_name = "default";
+
+    /// traget GPU platform name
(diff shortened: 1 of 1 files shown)


iomaganaris (approved):
LGTM

PR #824, opened 2022-03-16
Fixed suffix_random_string and added explanations

iomaganaris:


src/visitors/visitor_utils.cpp
```diff
@@ -32,17 +32,27 @@ std::string suffix_random_string(const std::set<std::string>& vars,
...
-    // has original_string as prefix and, if yes,
-    // try to use a different string in the form "original_string"_"random_string"
-
-    const auto it = vars.lower_bound(new_string);
-    while ((it != vars.end()) && new_string == it->substr(0, new_string.size())) {
-        random_string = singleton_random_string_class.reset_random_string(original_string, use_num);
-        new_string = original_string;
-        new_string += "_" + random_string;
+    // If the "original_string" is not in the set of the variables to check then
+    // return the "original_string" without suffix
```
cattabiani, line 35: (suggestions on how to improve the comments)

what is inside the set of variables to check? Why do we need to check only this set?

src/visitors/visitor_utils.cpp
```diff
@@ -32,17 +32,27 @@ std::string suffix_random_string(const std::set<std::string>& vars,
...
+    if (vars.find(original_string) == vars.end()) {
+        return new_string;
+    }
+    // Check if there is a variable defined in the mod file and, if yes, try to use
+    // a different string in the form "original_string"_"random_string"
+    // If there is already a "random_string" assigned to the "originl_string" return it
+    if (singleton_random_string_class.random_string_exists(original_string)) {
+        random_suffix += "_" + singleton_random_string_class.get_random_string(original_string);
```
cattabiani: I would assign instead of +=. It is less error-prone if, by any strange reason, random_suffix is not empty

src/visitors/visitor_utils.cpp
```diff
@@ -32,17 +32,27 @@ std::string suffix_random_string(const std::set<std::string>& vars,
...
+        new_string = original_string + random_suffix;
+    } else {
+        // Check if the "random_string" already exists in the set of variables and if it does try
+        // to find another random string to add as suffix
+        while (vars.find(new_string) != vars.end()) {
+            random_suffix +=
```
cattabiani: same here

src/visitors/visitor_utils.cpp
```diff
@@ -32,17 +32,27 @@ std::string suffix_random_string(const std::set<std::string>& vars,
...
-        new_string += "_" + random_string;
+    // If the "original_string" is not in the set of the variables to check then
+    // return the "original_string" without suffix
+    if (vars.find(original_string) == vars.end()) {
+        return new_string;
+    }
+    // Check if there is a variable defined in the mod file and, if yes, try to use
+    // a different string in the form "original_string"_"random_string"
+    // If there is already a "random_string" assigned to the "originl_string" return it
+    if (singleton_random_string_class.random_string_exists(original_string)) {
```
cattabiani, line 44: this is the case where we have already generated a random suffix right? In this way we have always the same suffix given a certain prefix (just to check that this is what you want to do, not an error atm)

iomaganaris: Yes exactly

src/visitors/visitor_utils.cpp
iomaganaris: Already taken care of it :wink:

cattabiani (approved):
LGTM!

PR #826, opened 2022-03-16
Get rid of one unnecessary shared_ptr

ohm314:
The field ast in GlobalToRangeVisitor doesn't need to be a shared pointer,
so we store it as a reference that is initialized at construction time.


pramodk (approved):
LGTM

PR #829, opened 2022-03-21
[LLVM][GPU] Added CUDADriver to execute benchmark on GPU

iomaganaris:
- Added CUDADriver to compile LLVM IR string generated from CodegenLLVMVisitor to PTX string and then exexcute it using CUDA API
- Tested the CUDADriver with external example LLVM IR code and works fine
- CUDA used to build NMODL should be the same version as the CUDA driver version installed on the machine the benchmark is executed otherwise there is `CUDA_ERROR_INVALID_PTX` error
- Added linkage to `libdevice` library
- To execute the benchmark with the `libdevice` we need to run the following:
```
./bin/nmodl ../test.mod --output "llvm_cuda" --verbose debug llvm --no-debug --ir --opt-level-ir 3 gpu --target-arch "sm_80" --name "nvptx64" --math-library libdevice benchmark --run --libs "${CUDA_ROOT}/nvvm/libdevice/libdevice.10.bc" --opt-level-codegen 3 --instance-size 10000000 --repeat 2 --grid-dim-x 4096 --block-dim-x 256
```
WIP:
- [x] At the moment the generated LLVM IR from NMODL doesn't generate right PTX code due to the address spaces not being set properly Edit: Generated LLVM IR has instructions that convert pointers in generic address space to global pointers so generated LLVM IR code is now executable to GPU using this PR
- [x] Kernel name in nvvm anotation should change to have the real name instead of "kernel"
- [x] Handle compilation options
- [x] Transform LLVM IR to bitcode to pass to `nvvmAddModuleToProgram`. Passing string is deprecated
- [x] Find a way to set `Triplet` and `DataLayout` for GPUDriver in non-hardcoded way (?)
- [x] Check why code cannot be compiled on GPU with kernel attributes
This is the output of NVVM when it tries to compile the kernel with the default attributes:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Compilation Log:
 nmodl_kernel: parse Unknown attribute kind (62) (Producer: 'LLVM13.0.0' Reader: 'LLVM 7.0.1')
 ```
 I am using CUDA 11.4.2.
- [ ] Check why code cannot be compiled on GPU with debug flags. This is the output with debug flags:
```
[NMODL] [info] :: CUDA JIT ERROR LOG: ptxas application ptx input, line 264; fatal   : Parsing error near '-': syntax error
ptxas fatal   : Ptx assembly aborted due to errors
 ```
 The related PTX code:
 ```
 ...
	.section	.debug_pubnames
	{
.b32 LpubNames_end0-LpubNames_start0
LpubNames_start0:
.b8 2
...
LpubNames_end0:
	}
	.section	.debug_pubtypes
	{
.b32 LpubTypes_end0-LpubTypes_start0
LpubTypes_start0:
.b8 2
.b8 0
.b32 .debug_info
.b32 182
.b32 0
LpubTypes_end0:
	}
	.section	.debug_loc	{	}
 ```
 You can reproduce the issue if you remoce the `--no-debug` option in the command above
- [x] Added GitLab CI test that executes benchmark in Cascade Lake CPU and V100 GPU node


georgemitenkov:
@iomaganaris Regarding some points from your TODO list:

1) Transform LLVM IR to bitcode to pass to nvvmAddModuleToProgram. Passing string is deprecated
You can use `WriteBitcodeToFile` function that takes a `llv,::Module` and the file path. 

 2) Find a way to set Triplet and DataLayout for GPUDriver in non-hardcoded way (?)
Usually you can create target machine (`llvm::TargetMachine`) and get all this information automatically, like it is currently implemented for CPUs (see `set_triple_and_data_layout` in `jit_driver.cpp`). For GPUs it is not really possible, as we need to provide some device information. However, what we can do something like:
[https://github.com/llvm/llvm-project/blob/main/polly/lib/CodeGen/PPCGCodeGeneration.cpp#L2301](https://github.com/llvm/llvm-project/blob/main/polly/lib/CodeGen/PPCGCodeGeneration.cpp#L2301)

test/benchmark/CMakeLists.txt
```diff
@@ -11,6 +12,9 @@ set(LLVM_BENCHMARK_SOURCE_FILES
 include_directories(${LLVM_INCLUDE_DIRS})
 add_library(llvm_benchmark STATIC ${LLVM_BENCHMARK_SOURCE_FILES})
 add_dependencies(llvm_benchmark lexer util visitor)
+if(NMODL_ENABLE_LLVM_CUDA)
+  target_link_libraries(llvm_benchmark PRIVATE CUDA::cudart CUDA::nvrtc ${CUDA_NVVM_LIBRARY_DIR}/libnvvm.so)
```
pramodk, line 16: should we avoid this hardcoding `libnvvm.so`? (this `.so` may not work on windows)

src/main.cpp
```diff
@@ -695,40 +732,64 @@ int main(int argc, const char* argv[]) {
...
+                                                           modfile,
+                                                           output_dir,
+                                                           shared_lib_paths,
+                                                           num_experiments,
+                                                           instance_size,
+                                                           platform,
+                                                           llvm_opt_level_ir,
+                                                           llvm_opt_level_codegen);
+                        benchmark.run(ast);
+                    }
```
pramodk, line 791: `nitpick`: can above duplication be removed as `platform` can be used to decide if passed `gpu_execution_parameters` can be ignored?

test/benchmark/cuda_driver.cpp
```diff
@@ -0,0 +1,166 @@
...
+        cuGetErrorName(err, &ret);
+        throw std::runtime_error("CUDA error: " + std::string(ret));
+    }
+}
+
+void CUDADriver::checkNVVMErrors(nvvmResult err) {
+    if (err != NVVM_SUCCESS) {
+        size_t program_log_size;
+        nvvmGetProgramLogSize(prog, &program_log_size);
+        auto program_log = (char*) malloc(program_log_size);
```
pramodk, line 38: Curious- is this `malloc` way is a standard approach? what happens to allocated memory when this exception is catched somewhere?

pramodk (commented):
Looking good. I added minor comments/ suggestions. 
I assume George could have better feedback on the core benchmarking part.

CMakeLists.txt
```diff
@@ -165,6 +165,10 @@ if(NMODL_ENABLE_LLVM)
   if(NMODL_ENABLE_LLVM_CUDA)
     enable_language(CUDA)
     find_package(CUDAToolkit)
+    set(CUDA_NVVM_INCLUDE_DIR ${CUDAToolkit_LIBRARY_ROOT}/nvvm/include)
+    set(CUDA_NVVM_LIBRARY_DIR ${CUDAToolkit_LIBRARY_ROOT}/nvvm/lib64)
+    set(NMODL_CUDA_INCLUDE_DIRECTORIES ${CUDAToolkit_INCLUDE_DIRS} ${CUDA_NVVM_INCLUDE_DIR})
```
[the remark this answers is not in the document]
iomaganaris: I've updated some parts of the file and used this in the end :+1:

test/benchmark/CMakeLists.txt
iomaganaris: I changed this to `${CUDA_NVVM_LIBRARIES}` so it should provide the correct library in Windows. I didn't think about them in the beginning :see_no_evil:

src/main.cpp
```diff
@@ -197,6 +197,24 @@ int main(int argc, const char* argv[]) {
...
+    int llvm_cuda_grid_dim_z = 1;
+
+    /// X dimension of block in threads for GPU execution
+    int llvm_cuda_block_dim_x = 1;
+
+    /// Y dimension of block in threads for GPU execution
+    int llvm_cuda_block_dim_y = 1;
+
+    /// Z dimension of block in threads for GPU execution
+    int llvm_cuda_block_dim_z = 1;
```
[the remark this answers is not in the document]
iomaganaris: Yes, actually we should probably hide them from the CLI

src/main.cpp
```diff
@@ -389,6 +408,24 @@ int main(int argc, const char* argv[]) {
...
+                                "Grid dimension Z ({})"_format(llvm_cuda_grid_dim_z))->ignore_case();
+    benchmark_opt->add_option("--blockDimX",
+                                llvm_cuda_block_dim_x,
+                                "Block dimension X ({})"_format(llvm_cuda_block_dim_x))->ignore_case();
+    benchmark_opt->add_option("--blockDimY",
+                                llvm_cuda_block_dim_y,
+                                "Block dimension Y ({})"_format(llvm_cuda_block_dim_y))->ignore_case();
+    benchmark_opt->add_option("--blockDimZ",
+                                llvm_cuda_block_dim_z,
+                                "Block dimension Z ({})"_format(llvm_cuda_block_dim_z))->ignore_case();
```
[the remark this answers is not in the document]
iomaganaris: Yes I think it makes sense to remove them

test/benchmark/gpu_parameters.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+struct GPUExecutionParameters {
+    int gridDimX;
+    int gridDimY;
+    int gridDimZ;
+    int blockDimX;
+    int blockDimY;
+    int blockDimZ;
+    int sharedMemBytes;
+};
```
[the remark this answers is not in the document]
iomaganaris: I just declared them for completeness but I guess we'll never use them so we can remove them

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -465,6 +465,10 @@ void CodegenLLVMVisitor::write_to_variable(const ast::VarName& node, llvm::Value
 }
 
 void CodegenLLVMVisitor::wrap_kernel_functions() {
+    // Wrapper doesn't work on GPU
```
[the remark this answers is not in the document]
georgemitenkov: So I think that we have to annotate the wrapper function as well, something like:

```c++
if (platform.is_gpu()) {
   // `wrapper_func` is the current instance of `llvm::Function` we create
  annotate_kernel_with_nvvm(wrapper_func);
} else {
  ir_builder.set_kernel_attributes();
}
```

src/main.cpp
georgemitenkov: I guess ideally  we would want to have 3 classes: Platform, GPUPlatform <: Platform, CPUPlatform <: Platform and all associated parameters can be passed there. Then in the code we can have a check if the platform is a GPU, say, cast it, and query grid/block dimensions.

I guess it is some work though, so having an extra optional argument to `Platform` constructor will work, as Pramod said :)

test/benchmark/llvm_benchmark.cpp
```diff
@@ -36,14 +41,14 @@ void LLVMBenchmark::generate_llvm(const std::shared_ptr<ast::Program>& node) {
...
-void LLVMBenchmark::run_benchmark(const std::shared_ptr<ast::Program>& node) {
+void LLVMBenchmark::run_benchmark_on_cpu(const std::shared_ptr<ast::Program>& node) {
     // Set the codegen data helper and find the kernels.
     auto codegen_data = codegen::CodegenDataHelper(node, llvm_visitor.get_instance_struct_ptr());
     std::vector<std::string> kernel_names;
     llvm_visitor.find_kernel_names(kernel_names);
 
     // Get feature's string and turn them off depending on the cpu.
-    std::string cpu_name = cpu == "default" ? llvm::sys::getHostCPUName().str() : cpu;
+    const auto cpu_name = platform.get_name();
```
georgemitenkov, line 51: Note that `get_name()` can return "default" if user has't specified the target CPU. So I guess we should have:
```
const auto cpu_name  platform.get_name() == "default" ? llvm::sys::getHostCPUName().str() : platform.get_name();
```

test/benchmark/cuda_driver.cpp
iomaganaris: Improved this :+1: I think after C++17 this can become even better with the use of `str.data()`

test/benchmark/cuda_driver.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+std::string print_bitcode_to_string(const llvm::Module& module) {
+    std::string bitcode_string;
+    llvm::raw_string_ostream os(bitcode_string);
+    WriteBitcodeToFile(module, os);
+    os.flush();
+    return bitcode_string;
+}
+
+void CUDADriver::init(const std::string& gpu, BenchmarkInfo* benchmark_info) {
```
georgemitenkov, line 95: I don't think `gpu` is used anywhere?

test/benchmark/cuda_driver.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+  public:
+    explicit CUDADriver(std::unique_ptr<llvm::Module> m)
+        : module(std::move(m)) {}
+
+    /// Initializes the CUDA GPU JIT driver.
+    void init(const std::string& gpu, BenchmarkInfo* benchmark_info = nullptr);
+
+    /// Lookups the entry-point without arguments in the CUDA module and executes it.
+    void execute_without_arguments(const std::string& entry_point,
```
georgemitenkov, line 91: `nitpick`: There is a bit of duplication between `execute_without_arguments` and `execute_with_arguments`. The only difference is the line

```c++
// void* kernel_parameters[] = {};
void* kernel_parameters[] = {&arg};
```

Maybe construct `void*` in `run_with_arguments()` and `run_without_arguments()` bodies, and then have a single `execute()` function that  takes `kernel_parameters` as an argument?

src/codegen/llvm/codegen_llvm_visitor.cpp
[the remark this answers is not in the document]
georgemitenkov: Also, wrapper has a return statement (I think for CPUs motivation was to return error codes), which doesn't make sense for CUDA kernels. Therefore we must have these
```c++
// Create a 0 return value and a return instruction.
 ir_builder.create_i32_constant(0);
 ir_builder.create_return(ir_builder.pop_last_value());
```
conditioned on the platform we use?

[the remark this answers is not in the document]
iomaganaris: Is there any reason to return something also in the CPU backend?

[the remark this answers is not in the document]
iomaganaris: I added a fix based on the above in https://github.com/BlueBrain/nmodl/pull/829/commits/a66d1554863e39b10b04e32b5f46f91f9df31c62. Let me know what you think about that

georgemitenkov:
@iomaganaris Regarding the errors: I think both of them are related to bitcode versions that we use.
```
terminate called after throwing an instance of 'std::runtime_error'
what():  Compilation Log:
nmodl_kernel: parse Unknown attribute kind (62) (Producer: 'LLVM13.0.0' Reader: 'LLVM 7.0.1')
```
Here, I suppose that we read bitcode with older LLVM version (7)? Looking at attributes' kinds 62 is `nofree` function attribute (added by optimization passes) which based on this [RFC](https://lists.llvm.org/pipermail/llvm-dev/2018-July/124555.html) was added after 2018, hence it shouldn't exist in LLVM version < 7?
```
[NMODL] [info] :: Device Compute Capability: 8.6
terminate called after throwing an instance of 'std::runtime_error'
what():  Compilation Log:
nmodl_kernel: parse Invalid record
```
This one I am not sure, but the error is definitely related to parsing wrong (different version/older) bitcode.


I don't have access to CUDA so cannot check the errors myself :( Were you able to track which method in `CUDADriver`/`GPUBenchmark` throws it?

iomaganaris:
@georgemitenkov 

Regarding the first error I think that the problem is that NVVM is based on LLVM 7.0.1 according to NVIDIA's documentation so it's not possible to parse these attributes https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html.
The error regarding the debug attributes come from [the same line](https://github.com/BlueBrain/nmodl/pull/829/files#diff-0927862cf67ef551eb7e95d4545ac9eafc8bedc0c49ebb140c576400cb1d67b1R141) where we compile the LLVM IR to PTX. I'm guessing that there is some similar issue there with the generated LLVM IR with debug attributes not being able to be parsed by LLVM 7.0.1 based NVVM.

iomaganaris:
With the latest changes in this PR and George's changes in https://github.com/BlueBrain/nmodl/pull/835 I managed to execute via the benchmark the test mod file with `exp` that has been added as a unit test in the mentioned PR.
The only thing still missing is support for debug flags. I need to investigate this a bit to see what LLVM does

iomaganaris:
Seems like addition of debug symbols for the GPU is not straightforward. I found [this mail in the llvm mailing list](https://lists.llvm.org/pipermail/llvm-dev/2017-November/118841.html) that summarizes the reasons why debug information in the GPU code are not working out of the box.
By changing by hand the following line in the generated PTX code:
```
.b32 LpubTypes_end0-LpubTypes_start0
```
to calculate explicitly the sector sizes I managed to get the PTX compiled to a `cubin` however I didn't try to load it and execute it to see if the rest of the debug info are correct.
I've also tested running the CUDA JIT with `nsys --profile` and `ncu` and they both provide information for the kernels executed.
Instead of adding more stuff in this PR about the debug information on GPU code and making it even bigger I am wondering whether we should review and merge it and handle the addition of debug info in another PR.
cc: @georgemitenkov @pramodk

georgemitenkov:
@iomaganaris I think reviewing & merging and dealing with debug information in separate PR sounds reasonable. Maybe we can then have a warning that debug information is not supported by nvptx backend?

iomaganaris:
I added a warning about this and automatically disabling the addition of debug symbols when there is a GPU platform selected. I have also created #844 to track this issue

pramodk (approved):
some minor comments/suggestions buts LGTM!

@iomaganaris : sorry for missing this from review. Lets get this in asap!

.gitlab-ci.yml
```diff
@@ -68,3 +67,45 @@ test:gcc:
...
+    SPACK_PACKAGE: nmodl
+    SPACK_PACKAGE_SPEC: ~legacy-unit~python+llvm+llvm_cuda
+    SPACK_INSTALL_EXTRA_FLAGS: -v
+    SPACK_PACKAGE_COMPILER: gcc
+
+test_benchmark:cpu:
+  extends:
+    - .benchmark_config
+  script:
+    - module load unstable git
```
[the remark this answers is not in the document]
iomaganaris: Indeed it should be needed. I just followed the spack setup guide for bb5

test/benchmark/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 # =============================================================================
 set(LLVM_BENCHMARK_SOURCE_FILES
     ${CMAKE_CURRENT_SOURCE_DIR}/jit_driver.cpp ${CMAKE_CURRENT_SOURCE_DIR}/jit_driver.hpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/cuda_driver.cpp ${CMAKE_CURRENT_SOURCE_DIR}/cuda_driver.hpp
```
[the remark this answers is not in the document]
iomaganaris: Good point 👍

src/pybind/pynmodl.cpp
```diff
@@ -171,10 +182,14 @@ class JitDriver {
     benchmark::BenchmarkResults run(std::shared_ptr<nmodl::ast::Program> node,
                                     std::string& modname,
                                     int num_experiments,
-                                    int instance_size) {
+                                    int instance_size,
```
ohm314, line 185: shoudln't this be conditional on running with GPU enabled?
I would add the grid size into the CodegenConfig maybe? or maybe into an additional constructor for JitDriver that is in a `#ifdef NMODL_LLVM_CUDA_BACKEND`.

src/pybind/pynmodl.cpp
```diff
@@ -171,10 +182,14 @@ class JitDriver {
...
+                                    int cuda_grid_dim_x,
+                                    int cuda_block_dim_x) {
         cg_driver.prepare_mod(node);
         nmodl::codegen::CodegenLLVMVisitor visitor(modname, cfg.output_dir, platform, 0);
         visitor.visit_program(*node);
+        const GPUExecutionParameters gpu_execution_parameters{cuda_grid_dim_x,
```
ohm314, line 191: this should only be needed if your target is a GPU, no?

ohm314 (commented):
Only looked at the changes in pyjit... I think there might be some small improvements useful to avoid having to fix things later

src/pybind/pynmodl.cpp
iomaganaris: I think that the grid dimensions conceptionally shouldn't be part of `CodegenConfig` since they don't influence in any way the code generation.
What we could do is create another declaration of `run` function with these parameters only `#ifdef NMODL_LLVM_CUDA_BACKEND` however this would duplicate some parts of the code so maybe it's better to just keep those parameters as optional and just in the CPU case they are not taken into consideration

ohm314: OK, that's fine with me. It'll need maybe still some cleanup at a later point.

test/benchmark/cuda_driver.cpp
```diff
@@ -0,0 +1,197 @@
...
+    case 72:
+        return CU_TARGET_COMPUTE_72;
+    case 75:
+        return CU_TARGET_COMPUTE_75;
+    case 80:
+        return CU_TARGET_COMPUTE_80;
+    case 86:
+        return CU_TARGET_COMPUTE_86;
+    default:
+        throw std::runtime_error("Unsupported compute architecture");
```
pramodk, line 102: I probably don't know full workflow here but one comment come to my mind: Newer GPUs can run binaries compiled for older generations. If detected compute capability is "bigger", do we need to throw an exception or we can get away by targeting older architecture?

test/benchmark/cuda_driver.cpp
```diff
@@ -0,0 +1,197 @@
...
+    // set up size of compilation log buffer
+    jitOptions[0] = CU_JIT_INFO_LOG_BUFFER_SIZE_BYTES;
+    size_t jitLogBufferSize = 1024 * 1024;
+    jitOptVals[0] = (void*) jitLogBufferSize;
+
+    // set up pointer to the compilation log buffer
+    jitOptions[1] = CU_JIT_INFO_LOG_BUFFER;
+    char* jitLogBuffer = new char[jitLogBufferSize];
+    jitOptVals[1] = jitLogBuffer;
+
```
pramodk, line 165: Objects explicitly allocated like `jitLogBuffer` and `jitOptions` are internally freed by llvm?

pramodk (approved):
I quickly skimmed through the changes, nothing major I can point out. Apart from clarification comments, this is good from my side.

test/benchmark/cuda_driver.cpp
iomaganaris: There is no need to worry about older/newer generations here since for the CUDA JIT we actually read the exact architecture of the GPU we're going to use and pass it to it. This function just makes the conversion from the device target architecture of the exact GPU we are using to `CUjit_target` to be used by the CUDA JIT. I should rename the name of the function though to make it more clear

pramodk: sounds good!

test/benchmark/cuda_driver.cpp
```diff
@@ -0,0 +1,197 @@
...
+    if (!std::string(jitLogBuffer).empty()) {
+        logger->info("CUDA JIT INFO LOG: {}"_format(std::string(jitLogBuffer)));
+    }
+    if (!std::string(jitErrorLogBuffer).empty()) {
+        logger->info("CUDA JIT ERROR LOG: {}"_format(std::string(jitErrorLogBuffer)));
+    }
+    free(jitOptions);
+    free(jitOptVals);
+    free(jitLogBuffer);
+    free(jitErrorLogBuffer);
```
pramodk, line 196: should be `delete` here?

pramodk (changes requested):

test/benchmark/cuda_driver.cpp
iomaganaris: Fixed :+1:

pramodk (approved):

PR #830, opened 2022-03-21
Accept array variable in diffeq lexer

alkino:
The file below was failing because `m' =  (inf[0] - m)/tau[0]` cannot be tokenized by the lexer due to the parameter.
https://github.com/BlueBrain/nmodldb/blob/downloader/models/db/modeldb/266806/mod/Cav2_3.mod

### DIFF
--- a/src/lexer/diffeq.ll
+++ b/src/lexer/diffeq.ll
@@ -27,6 +27,7 @@
 
 D   [0-9]
 E   [Ee][-+]?{D}+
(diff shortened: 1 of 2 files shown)


pramodk (commented):
Looks good but it would be great if you can add some description in the PR summarising what was the issue and/or mod file that was causing the error. (By looking at the title, it’s not clear)

pramodk (approved):
LGTM 👌

PR #835, opened 2022-03-30
[LLVM][GPU][+refactoring] Replacement of math intrinsics with library calls

georgemitenkov:
This PR adds a LLVM pass that replaces math intrinsics
with calls to math library. In particular:

*  Functionality of replacement with SIMD functions is factored
out into a separate file and LLVM version dependencies are
dropped (we use LLVM 13 already anyway).

* A pass to replace intrinsics with libdevice calls when targeting
CUDA platforms has been added. So far only `exp` is supported
(single and double precision)

* Added a test to check the replacement

Note: factoring replacement functionality into a separate file
allows us to completely drop the dependency on target information
inside `LLVMCodegenVisitor`😊

### DIFF
--- a/src/codegen/llvm/replace_with_lib_functions.cpp
+++ b/src/codegen/llvm/replace_with_lib_functions.cpp
@@ -13,9 +13,9 @@
 #include "llvm/Analysis/OptimizationRemarkEmitter.h"
 #include "llvm/Analysis/ScalarEvolution.h"
 #include "llvm/CodeGen/ReplaceWithVeclib.h"
+#include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/InstIterator.h"
 #include "llvm/IR/IntrinsicsNVPTX.h"
-#include "llvm/IR/IRBuilder.h"
 #include "llvm/IR/LegacyPassManager.h"
 
 namespace llvm {
@@ -29,7 +29,6 @@ bool ReplaceMathFunctions::runOnModule(Module& module) {
     // If the platform supports SIMD, replace math intrinsics with library
     // functions.
     if (platform->is_cpu_with_simd()) {
-
         // First, get the target library information and add vectorizable functions for the
         // specified vector library.
         Triple triple(sys::getDefaultTargetTriple());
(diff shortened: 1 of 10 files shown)


src/codegen/llvm/replace_with_lib_functions.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+        // Add vectorizable functions to the target library info.
+        switch (library->second) {
+        case VecLib::LIBMVEC_X86:
+            if (!triple.isX86() || !triple.isArch64Bit())
+                break;
+        default:
+            tli.addVectorizableFunctionsFromVecLib(library->second);
+            break;
+        }
```
iomaganaris, line 121: Just a personal opinion, not sure what should be the proper way to do it or what could be the benefit of the `switch` but I think it would  be more understandable to write this like:
```suggestion
       if (library->second != VecLib::LIBMVEC_X86 || (triple.isX86() && triple.isArch64Bit())) {
           tli.addVectorizableFunctionsFromVecLib(library->second);
       }
```
Feel free to keep this as you prefer

src/codegen/llvm/replace_with_lib_functions.cpp
```diff
@@ -0,0 +1,214 @@
...
+    bool is_nvvm_intrinsic = id == Intrinsic::nvvm_read_ptx_sreg_ntid_x ||
+            id == Intrinsic::nvvm_read_ptx_sreg_nctaid_x ||
+            id == Intrinsic::nvvm_read_ptx_sreg_ctaid_x ||
+            id == Intrinsic::nvvm_read_ptx_sreg_tid_x;
+    if (id == Intrinsic::not_intrinsic || is_nvvm_intrinsic)
+        return false;
+
+    // Map of supported replacements. For now it is only exp.
+    static const std::map<std::string, std::string> libdevice_name = {
+            {"llvm.exp.f32", "__nv_expf"},
```
iomaganaris, line 172: I think it would be good to add also `pow` and maybe look for any other math function commonly used in the mod files and also add it to the `x86` and `aarch64` maps. I can look at this also the next days

iomaganaris (approved):
LGTM
Just a few small suggestions

PR #838, opened 2022-04-05
[LLVM] Code formatting changes

iomaganaris:
- Cherry-pick update of hpc-coding-conventions from `master`
- Run clang-format with clang-format-13 (part of llvm 13 we are using)

Please let me know if there is any issue with these changes and the clang-format version you are using


georgemitenkov:
@iomaganaris Thanks for doing that! I think previously clang-forma-8 was used, so having clang-format-13 makes everything way easier!

georgemitenkov (approved):

PR #840, opened 2022-04-07
Add an option to "clean" mod files

alkino:


alkino:
This idea is abandonned

PR #843, opened 2022-04-08
Add a binary to format mod files

alkino:


ohm314:
I'm wondering if we don't want to add this rather into the nmodl main binary as a `--format` option?

pramodk:
I have same question as Omar: could we refactor existing `main()` of the nmodl binary and compose such options there?

pramodk:
@alkino : I think you mentioned that this was just as proof-of-concept i.e. we don't prefer separate binary for this purpose. So safe to close?

PR #845, opened 2022-04-19
Avoid using raw pointers in InlineVisitor

### LINKED ISSUE #809: Code generation issue with inlining
Installing `NEURON` with `NMODL` on Apple Macbook Air with M1 processor generated the following issue when compiling the [na3n.mod file from the reduced_dentate repo](https://github.com/neuronsimulator/reduced_dentate/blob/master/mechanisms/na3n.mod):
```
arm64/corenrn/mod2c/na3n.cpp:544:32: error: use of undeclared identifier 'bets_in_0'
                exptrap_in_4 = bets_in_0;
```
Seems like there is some issue with the code generated in 
```c++
exptrap_in_4 = bets_in_0;
```
which should be:
```c++
if (x_in_4 >= 700.0) {
    exptrap_in_4 = exp(700.0);
} else {
    exptrap_in_4 = exp(x_in_4);
}
```
after the inline pass.
Unfortunately I couldn't reproduce the issue when I installed `NEURON` with `CoreNEURON` and `NMODL` using the following commands:
```bash
git clone --recursive https://github.com/neuronsimulator/nrn.git
cd nrn
mkdir build_nmodl
python3 -v venv venv
. venv/bin/activate
pip3 install -r nrn_requirements.txt
pip3 install -U pip setuptools scikit-build Jinja2 PyYAML pytest 'sympy>=1.3,<1.9' 'cmake-format==0.6.13'  # NMODL requirements
brew install flex bison
export PATH="/opt/homebrew/opt/bison/bin:/opt/homebrew/opt/flex/bin:$PATH"
export PYTHONPATH="$(pwd)/build_nmodl/install/lib/python:$PYTHONPATH"
export SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
cmake .. -DCMAKE_INSTALL_PREFIX=./install -DNRN_ENABLE_CORENEURON=ON -DCORENRN_ENABLE_NMODL=ON -DNRN_ENABLE_RX3D=OFF -DNRN_ENABLE_INTERVIEWS=OFF -DNRN_ENABLE_TESTS=ON
cmake --build . --target install --parallel 1
ctest -R reduced_dentate
```

iomaganaris:
Usage of raw pointers in InlineVisitor seems like it generated bugs when running on the Apple M1 processor. Instead of using raw pointers handle cloning statements in need of replacing using `unique_ptr`.

Fixes #809 

Note: Unfortunately I couldn't validate this in any way apart from running `NMODL` multiple times since valgrind is not working on mac. Before this change rerunning `NMODL` with certain mod file would randomly crash with issues similar to the ones on mentioned on the above issue.

### DIFF
--- a/src/visitors/inline_visitor.cpp
+++ b/src/visitors/inline_visitor.cpp
@@ -159,7 +159,7 @@ bool InlineVisitor::inline_function_call(ast::Block& callee,
     }
 
     /// get a copy of function/procedure body
-    auto inlined_block = callee.get_statement_block()->clone();
+    auto inlined_block = std::make_unique<ast::StatementBlock>(*(callee.get_statement_block()->clone()));
 
     /// function definition has function name as return value. we have to rename
     /// it with new variable name
@@ -177,7 +177,7 @@ bool InlineVisitor::inline_function_call(ast::Block& callee,
         add_return_variable(*inlined_block, new_varname);
     }
 
-    auto statement = new ast::ExpressionStatement(inlined_block);
+    auto statement = new ast::ExpressionStatement(std::move(inlined_block));
 
     if (to_replace) {
         replaced_statements[caller_statement] = statement;
--- a/src/visitors/inline_visitor.cpp
+++ b/src/visitors/inline_visitor.cpp
@@ -159,7 +159,8 @@ bool InlineVisitor::inline_function_call(ast::Block& callee,
     }
 
     /// get a copy of function/procedure body
-    auto inlined_block = std::make_unique<ast::StatementBlock>(*(callee.get_statement_block()->clone()));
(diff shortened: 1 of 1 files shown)


src/visitors/inline_visitor.cpp
```diff
@@ -159,7 +159,8 @@ bool InlineVisitor::inline_function_call(ast::Block& callee,
     }
 
     /// get a copy of function/procedure body
-    auto inlined_block = callee.get_statement_block()->clone();
+    auto inlined_block = std::make_unique<ast::StatementBlock>(
```
alkino, line 162: Here you should call the ctor and not make_unique.

alkino (approved):

PR #847, opened 2022-04-27
Updating copyright year 2022

bbpgithubaudit:


pramodk (commented):
additional change also need in: https://github.com/BlueBrain/nmodl/blob/cde5dbf02fe93bb1b0b8cc9b5a6ebe0747ab8720/LICENSE#L4

and, all source codes have lines like: https://github.com/BlueBrain/nmodl/blob/cde5dbf02fe93bb1b0b8cc9b5a6ebe0747ab8720/src/main.cpp#L2

pramodk:
LGTM

@alkino: did you just grep and replace the pattern?  If you had to do something extra (in case there are inconsistent patterns), would be helpful if you could post it here, and then I can apply the same to the `llvm` branch.

alkino:
```
git ls-files | grep -Ee "\.(h|yaml|h|l|cpp|py|hpp)\$" | xargs sed -Ei "s/([0-9]{4})-[0-9]{4}/\1-2022/"
```

At the end I did the last one by hand with:

```
git grep Copyright | grep -v 2022
```

pramodk (approved):
thanks! 

LGTM!

PR #849, opened 2022-04-27
Change _format() by fmt::format

### LINKED ISSUE #848: replace _format with fmt::format
`fmt` has deprecated its UDL-based `_format` API. This is starting to through warnings in our build: https://github.com/BlueBrain/nmodl/runs/6081886419?check_suite_focus=true#step:12:135

We should switch to the `fmt::format` API, which also provides more compile-time checks.

alkino:
_format() is now deprecated with fmt

Fix #848 

TODO: rebase llvm branch

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -60,8 +60,8 @@ using namespace visitor;
 using nmodl::parser::NmodlDriver;
 
 int main(int argc, const char* argv[]) {
-    CLI::App app{
-        "NMODL : Source-to-Source Code Generation Framework [{}]"_format(Version::to_string())};
+    CLI::App app{fmt::format("NMODL : Source-to-Source Code Generation Framework [{}]",
+                             Version::to_string())};
 
     /// list of mod files to process
     std::vector<std::string> mod_files;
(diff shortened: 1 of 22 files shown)


src/main.cpp
```diff
@@ -179,80 +179,92 @@ int main(int argc, const char* argv[]) {
     app.add_option("--scratch", scratch_dir, "Directory for intermediate code output")
         ->capture_default_str()
         ->ignore_case();
-    app.add_option("--units", units_dir, "Directory of units lib file")->capture_default_str()->ignore_case();
```
ohm314, line 182: this is an unrelated change, is that because you changed clang-format?

ohm314 (commented):
Code builds fine and changes look reasoanble. But you should remove all mentions of `using namespace fmt::literals;` since they should not be needed anymore.

src/main.cpp
alkino: No idea why. But it is mandatory

iomaganaris (approved):
LGTM 👍

PR #853, opened 2022-04-29
[LLVM][GPU] Atomic updates support

georgemitenkov:
This PR adds support for atomic updates on GPUs. In particular:

1. Helper visitor generates atomic statements with `+=` and `-=` once again.
2. LLVM visitor now knows how to lower atomic statements for CPUs (trivially) and GPUs.
3. A corresponding test was added.


src/codegen/llvm/codegen_llvm_helper_visitor.cpp
```diff
@@ -295,23 +295,9 @@ static std::shared_ptr<ast::CodegenAtomicStatement> create_atomic_statement(
...
-
-    // make sure only + or - operator is used
-    if (op_str != "-" && op_str != "+") {
-        throw std::runtime_error("Unsupported binary operator for atomic statement");
-    }
-
-    auto* rhs = create_expression("{}[{}] {} {} "_format(var, var_index, op, rhs_str));
-    return std::make_shared<ast::CodegenAtomicStatement>(lhs,
-                                                         ast::BinaryOperator{ast::BOP_ASSIGN},
-                                                         rhs);
```
pramodk, line 314: nice!

src/codegen/llvm/codegen_llvm_visitor.cpp
```diff
@@ -468,31 +468,73 @@ void CodegenLLVMVisitor::visit_boolean(const ast::Boolean& node) {
...
+            throw std::runtime_error("Error: atomic updates for non-instance variable\n");
+
+        const auto& node = std::dynamic_pointer_cast<ast::CodegenInstanceVar>(identifier);
+        const auto& instance_name = node->get_instance_var()->get_node_name();
+        const auto& member_node = node->get_member_var();
+        const auto& member_name = member_node->get_node_name();
+
+        if (!instance_var_helper.is_an_instance_variable(member_name))
+            throw std::runtime_error("Error: " + member_name +
+                                     " is not a member of the instance variable\n");
```
pramodk, line 506: to clarify: is this somehow overlaps with line 496+497?

pramodk (approved):
All looks reasonable to me!

src/codegen/llvm/codegen_llvm_visitor.cpp
iomaganaris: After merge with `llvm` branch those lines are `539`, `540`

iomaganaris: I think that
```
if (!identifier->is_codegen_instance_var())
            throw std::runtime_error("Error: atomic updates for non-instance variable\n");
```
checks whether `identifier` is type of `InstanceStruct` and 
```
if (!instance_var_helper.is_an_instance_variable(member_name))
            throw std::runtime_error("Error: " + member_name +
                                     " is not a member of the instance variable\n");
```
if `member_name` is a variable defined in the mod file in such way that there will be multiple instance of it so it will be a pointer variable inside the `InstanceStruct`, for example a `RANGE` variable. The second check is because it could only be such variables that need atomic insturctions

iomaganaris (approved):
LGTM :+1: 
Also had a look at the generated LLVM/PTX files and look reasonable to me.
Tests of functionality on CPU (atomics w/ vec width 1) and GPU are passing as well

src/codegen/llvm/codegen_llvm_visitor.cpp
georgemitenkov: What @iomaganaris said is right, I think. First check is for `InstanceStruct`, second is that we are generating code for pointer variable (member of instance). I think that it will always be like this (based on helper visitor) - so it;s nore of a sanity check :)

PR #855, opened 2022-04-30
Fix for instance struct type creation

georgemitenkov:
This is a small fix to make sure that the named struct
type (instance struct) is created only once.


pramodk (approved):
thanks for fixing this!

(It was confusing for me but I  didn't know the backend part enough, should have asked!)

iomaganaris (approved):
LGTM

PR #860, opened 2022-05-04
Make "sparse" solver check if equations are linear.

ctrl-z-9000-times:
If the system is linear, then newtons method always converges
in exactly one iteration. When using the sparse solver on
linear systems omit the newtons iteration and solve directly.

This should make the resulting code run marginally faster by
skipping the check for convergence. Currently the check for
convergence is implemented as "error = sqrt(|F|^2)".


nmodl/ode.py
```diff
@@ -291,7 +293,19 @@ def solve_non_lin_system(eq_strings, vars, constants, function_calls):
...
     code = _interweave_eqs(vecFcode, vecJcode)
 
-    return code
+    return code, linear
+
+
+def _is_linear(eqs, state_vars, sympy_vars):
+    for expr in eqs:
+        for x in state_vars:
+            for y in state_vars:
```
alkino, line 302: Could we here use: https://docs.python.org/3/library/itertools.html#itertools.combinations instead to increase speed?

ctrl-z-9000-times: Yes, I can do that.
Then it would only check one of the derivatives: "dxdy", "dydx" but not both.

ctrl-z-9000-times: done

test/unit/visitor/sympy_solver.cpp
```diff
@@ -2062,8 +2046,8 @@ SCENARIO("Solve NONLINEAR block using SympySolver Visitor", "[visitor][solver][s
                 }{
                     nmodl_eigen_x[0] = x
                 }{
-                    nmodl_eigen_f[0] = 5.0-nmodl_eigen_x[0]
-                    nmodl_eigen_j[0] = -1.0
+                    nmodl_eigen_f[0] = 5.0-pow(nmodl_eigen_x[0], 3)
+                    nmodl_eigen_j[0] = -3.0 * pow(nmodl_eigen_x[0], 2)
```
pramodk, line 2050: I will wait for comments from @cattabiani but just from code generation perspective I want to point out that use of `pow` is expensive and typically we want to rewrite/simplify `x^2` to `x*x`. IIRC, we use somewhere sympy to do this simplification.

cattabiani: true

cattabiani: I agree. There are more `pows` in the code too.

nmodl/ode.py
```diff
@@ -291,7 +294,18 @@ def solve_non_lin_system(eq_strings, vars, constants, function_calls):
     # interweave
     code = _interweave_eqs(vecFcode, vecJcode)
 
-    return code
+    return code, linear
+
+
+def _is_linear(eqs, state_vars, sympy_vars):
+    for expr in eqs:
+        for (x, y) in itertools.combinations_with_replacement(state_vars, 2):
```
alkino, line 302: why with replacement, do we want to try dxdx?

test/unit/visitor/sympy_solver.cpp
cattabiani: ok, sorry, I missread the comment. IIRC pow was used because we are not sure, in general, that the exponent is a positive integer

cattabiani: a few points about replacing pows:
- it brings some marginal improvements (I believe) if it remains robust and sound for all the mod files. It could be more error-prone from the user perspective. I let @alkino verify this
- pow has always being used, since I stated working on nmodl. I did not know about this Liam adversity to the function. I do not think it is really related to this pr
- the way to change it (that I see atm) would be to partially remove pow from the list of available functions in ode.py or a replacement afterward. Both ways look very cumbersome, hacky and maybe error-prone for not a lot of improvement

ctrl-z-9000-times: The reason "pow" appears in this PR is because I changed the testcase from solving a linear equation "x=5" to solving a non-linear equation "x^3=5".

I changed that test because it is a test for the nonlinear solver, and I wanted it to continue testing the same code paths.

nmodl/ode.py
ctrl-z-9000-times: Yes, dxdx is intentional.
Otherwise it would not detect that x^2 is nonlinear.

alkino: Thanks for clarification!

test/unit/visitor/sympy_solver.cpp
pramodk: Adding info from @cattabiani via Slack:

> the relevant ticket and mod file that needs to be be checked with this PR is: https://github.com/BlueBrain/nmodl/issues/801

ctrl-z-9000-times: Ok, I ran this PR on the file cdp5.mod and it works.
It successfully runs and the resulting C++ file correctly uses newtons method.

Here is the file in question:
https://senselab.med.yale.edu/ModelDB/showmodel?model=239421&file=/Purkinjecell_2017/mod_files/cdp5.mod#tabs-2

PR #864, opened 2022-05-11
[LLVM][SIMD] Atomic updates support

georgemitenkov:
This PR introduces atomic write support for SIMD code generation.

Current  update kernels sometimes require atomic updates, usually in the form of
```
index = indices[id]
data[index] += update[id]
```
Since it is not possible to generate atomics with SIMD, we insert a scalar loop
that updates every element of the vector one by one.

In particular, this PR:
- Adds code generation functionality for atomic writes on SIMD platforms
- Adds both execution and IR tests
- Refactors common GPU/SIMD atomics code


georgemitenkov:
Note: pipeline will fail without the patch that fixes instance struct generation (see another PR)

iomaganaris (approved):
LGTM :+1: 
I just added a test to make sure that also the correct calculations are done in the `POINT_PROCESS` mod files

PR #865, opened 2022-05-11
[LLVM][FIX] Float generation fix in LLVM helper visitor

georgemitenkov:
This PR fixes how single-precision floats are handles in LLVM
helper visitor. Before, `DOUBLE` was always created which was
not consistent with LLVM visitor (that lowered `DOUBLE` to 32-
or 64-bit types) and instance struct initialisation in testing (which
assumed that for `DOUBLE` it always holds that
`sizeof(double) = 8`).

Now, there is a clear separation between `FLOAT` and `DOUBLE`
values in helper visitor.

Note: this bug was not discovered before since execution tests were
only considering doubles. So, one of the tests was changed to use
floats.

### DIFF
--- a/test/unit/codegen/codegen_llvm_instance_struct.cpp
+++ b/test/unit/codegen/codegen_llvm_instance_struct.cpp
@@ -131,51 +131,51 @@ SCENARIO("Instance Struct creation", "[visitor][llvm][instance_struct]") {
             size_t node_count_index = 19;
             // Check if the various instance struct fields are properly initialized
             REQUIRE(compare(instance_data.members[minf_index],
-                            generate_dummy_data<double>(minf_index, num_elements)));
+                            generate_dummy_data<float>(minf_index, num_elements)));
             REQUIRE(compare(instance_data.members[ena_index],
-                            generate_dummy_data<double>(ena_index, num_elements)));
+                            generate_dummy_data<float>(ena_index, num_elements)));
             REQUIRE(compare(instance_data.members[ion_ena_index],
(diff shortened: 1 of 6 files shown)


test/unit/codegen/codegen_llvm_instance_struct.cpp
```diff
@@ -131,51 +131,51 @@ SCENARIO("Instance Struct creation", "[visitor][llvm][instance_struct]") {
             size_t node_count_index = 19;
             // Check if the various instance struct fields are properly initialized
             REQUIRE(compare(instance_data.members[minf_index],
-                            generate_dummy_data<double>(minf_index, num_elements)));
+                            generate_dummy_data<float>(minf_index, num_elements)));
```
iomaganaris, line 134: I would maybe keep this test using `double` since this is the default use case and we don't have another test about this. Not sure why it was set to `use_single_precision` from the beginning. Probably some negligence from my side :innocent:

iomaganaris (commented):
LGTM :+1:
Just a small suggestion regarding one of the tests

test/unit/codegen/codegen_llvm_instance_struct.cpp
georgemitenkov: Sure! I will change an execution test instead :)

iomaganaris: Thank you! :slightly_smiling_face:

iomaganaris (approved):

PR #867, opened 2022-05-13
[LLVM] Fixes compilation with LLVM codegen disabled

iomaganaris:
- Makes sure that `llvm` branch compiles with `LLVM` codegen and `Python bindings` disabled or enabled
- Selectively enables options of Python Codegen and Python JIT Driver
- Fixes the order of linkage of static variables for the generation of `nmodl` binary

### DIFF
--- a/src/pybind/pynmodl.cpp
+++ b/src/pybind/pynmodl.cpp
@@ -12,12 +12,15 @@
 
 #include "ast/program.hpp"
 #include "codegen/codegen_driver.hpp"
-#include "codegen/llvm/codegen_llvm_visitor.hpp"
 #include "config/config.h"
 #include "parser/nmodl_driver.hpp"
 #include "pybind/pybind_utils.hpp"
-#include "test/benchmark/llvm_benchmark.hpp"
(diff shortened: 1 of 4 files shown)


src/pybind/pynmodl.cpp
```diff
@@ -301,6 +311,9 @@ PYBIND11_MODULE(_nmodl, m_nmodl) {
              "instance_size"_a,
              "cuda_grid_dim_x"_a = 1,
              "cuda_block_dim_x"_a = 1);
+#else
+        ;
```
pramodk, line 315: `pure nitpick`: only thing cached my attention in this PR is this! looked like pattern from legacy nrn code! 🙃

pramodk (approved):
LGTM

src/pybind/pynmodl.cpp
iomaganaris: I know it's definitely not nice :see_no_evil:
Do you have any suggestion about how this could be handled better?

pramodk: don't think so !

PR #870, opened 2022-05-17
fixup! CI errors with sympy 1.9 and 1.10

### LINKED ISSUE #749: sympy 1.9 incompatibility with test: Solve ODEs with cnexp or euler method using SympySolverVisitor
[sympy 1.9 was released 2 days ago](https://pypi.org/project/sympy/1.9/), which has caused [CI failures](https://github.com/BlueBrain/nmodl/runs/3856512273):
```
../test/unit/visitor/sympy_solver.cpp:566: FAILED:
  REQUIRE( result[0] == "z' = a/z+b/z/z" )
with expansion:
  "z = (0.5*pow(a, 2)*pow(z, 2)-a*b*z+pow(b, 2)*log(a*z+b))/pow(a, 3)"
  ==
  "z' = a/z+b/z/z"
```

https://github.com/BlueBrain/nmodl/pull/748 stops the new sympy version from being used in the CI, but the underlying issue should be fixed so that we can remove these constraints.

cattabiani:
- bin is a function in python: renamed variable in test to not get this error
- sympy now can solve `z'=a/z+b/z/z`. It is getting hard to find simple equations
that it cannot solve.

Warning: this is a downgrade of the testing capabilities. I could not find
an ode with simple functions (no sin, exp, log etc. Only simple multiplications
and additions) that cannot be solved by sympy. Best case the test hangs waiting
for sympy to solve a very long equation. For this reason, with this PR, the
fact that the code should return the equation untouched if it cannot be processed
is untested

fix #749

makes the change in sympy in this PR obsolete:
https://github.com/BlueBrain/spack/pull/1591

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ def _config_exe(exe_name):
 
 install_requirements = [
     "PyYAML>=3.13",
-    "sympy>=1.3,<1.9",
+    "sympy>=1.3",
 ]
 
 
--- a/test/unit/visitor/sympy_solver.cpp
+++ b/test/unit/visitor/sympy_solver.cpp
@@ -563,13 +563,12 @@ SCENARIO("Solve ODEs with cnexp or euler method using SympySolverVisitor",
         THEN("Integrate equations analytically where possible, otherwise leave untouched") {
             auto result = run_sympy_solver_visitor(nmodl_text);
             REQUIRE(result.size() == 4);
-            REQUIRE(result[0] == "z' = a/z+b/z/z");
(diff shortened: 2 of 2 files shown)


olupton:
Can you modify the version constraint here:
https://github.com/BlueBrain/nmodl/blob/4e4dc027ed067b0543a51b1441be60b064b7a798/setup.py#L103
please? Then the CI should test a newer version.

olupton (approved):
LGTM 🚀

Once this is merged can you please make a PR to BlueBrain/spack so that `nmodl` will use newer `py-sympy` versions?

PR #874, opened 2022-05-25
[LLVM] Adding python script for benchmarking

iomaganaris:
- Updated mod files and cpp to latest changes
- Load externa shared library with kernels compiled with other compilers and run it with nmod JIT


test/benchmark/benchmark_script.py
```diff
@@ -0,0 +1,528 @@
...
+        self.libdevice_lib = libdevice_lib
+        self.nmodl_exe = nmodl_exe
+
+
+class BenchmarkConfig:
+    mod_files = ""
+    architectures = ""
+    compilers = ""
+    math_libraries = ["SVML", "SLEEF"]
+    fast_math = [False, True]
```
ohm314, line 49: I see how you use this below but i find it a bit weird - maybe I'll propose a better way in a separate commit

ohm314 (commented):
Made a few suggestions - I'll try to push some improvements on the benchmark script though.

test/benchmark/benchmark_script.py
iomaganaris: If you have a better suggestion feel free to add it :+1: We can also discuss about that if you want

ohm314 (approved):

PR #879, opened 2022-06-02
Update GitHub Actions CI, minor CMake fixes.

olupton:
- Only pin major versions, update some of those.
  - See https://github.com/JamesIves/github-pages-deploy-action/discussions/592
- Drop sympy version constraints.
- Coverage from a Debug, not Release, build.
- Prefer `ubuntu-20.04` to `ubuntu-18.04`, add `ubuntu-22.04` too.
- Drop `add_external_project` helper that was only used to bootstrap `hpc-coding-conventions`.
- Do not add a CLI11 submodule if the relevant targets already exist, this fixes some combinations of NMODL-as-a-submodule build.
- Fix `NMODL_3RDPARTY_DIR` to avoid the absolute path being prepended twice.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,37 +16,37 @@ on:
       - release/**
 
 env:
-  BUILD_TYPE: Release
-  CTEST_PARALLEL_LEVEL: 1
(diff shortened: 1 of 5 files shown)


pramodk (approved):
LGTM

PR #883, opened 2022-06-15
Check unit empty

### LINKED ISSUE #8: Issue warning for empty units during semantic analysis
Empty unit name declaration is valid according to the original implementation in NEURON:

i.e.

```console
UNITS 
{
                   () = (millivolt)
}
```

is accepted now. Note the empty bracket on the left where there should be a name. Similarly, function declarations like the below are now also accepted:

```console
FUNCTION ssCB(kdf(), kds()) (mM) 
{
}
```

Above behaviour was implemented in #3 in order to pass the fail existing ModelDB files. 

When we will have semantic analysis pass for error checking, we can add a check to see if there is an empty unit definition like above and then error / throw exception. This way, parsing will be successful but the semantic analysis will be a "proper" place to do such checking.

alkino:
Fix #8 (endly)


alkino:
I got only one problem `(_)` is not considered as an empty name, because the name is `_`.

Here `_` means space.

src/visitors/semantic_analysis_visitor.cpp
```diff
@@ -66,5 +68,14 @@ void SemanticAnalysisVisitor::visit_destructor_block(const ast::DestructorBlock&
     /// -->
 }
 
+void SemanticAnalysisVisitor::visit_unit(const ast::Unit& node) {
+    /// <-- This code is for check 3
+    if (node.get_name()->get_value().empty()) {
+        logger->warn("SemanticAnalysisVisitor:: An unit cannot be created without name.");
```
iomaganaris, line 74: Since the mod file translation fails with this I'm wondering whether we should print this message as `error` instead of `warning`.

pramodk:
@alkino : If this is a good change, I wonder if similar change should be done on neuron side. This way, we won't have situation where neuron will accept the mod file but coreneuron will fail to compile 🤔

src/visitors/semantic_analysis_visitor.cpp
alkino: done

pramodk:
> I guess we should modify the parser to strip spaces around name, right?

I believe so! I don't this is where "spaces" as a name make sense.

pramodk:
> @alkino : If this is a good change, I wonder if similar change should be done on neuron side

What do you think @alkino ? Can you check if NEURON accepts mod file containing empty unit name?

PR #884, opened 2022-06-16
Check that token is not nullptr

### LINKED ISSUE #53: Symtabvisitor try to dereference nullptr token
If we add new global block in the AST without token and then run symtab visitor then we get segfault from:

```c++
void SymtabVisitor::add_model_symbol_with_property(ast::Node* node, NmodlType property) {
    auto token = node->get_token();   => nullptr !
    auto name = node->get_node_name();
    auto symbol = std::make_shared<Symbol>(name, node, *token);
```

alkino:
Fix #53

### DIFF
--- a/src/visitors/symtab_visitor_helper.hpp
+++ b/src/visitors/symtab_visitor_helper.hpp
@@ -149,6 +149,9 @@ void SymtabVisitor::setup_symbol(ast::Node* node, NmodlType property) {
 
 void SymtabVisitor::add_model_symbol_with_property(ast::Node* node, NmodlType property) {
     auto token = node->get_token();
+    if (!token) { // token can be nullptr
+        return;
+    }
     auto name = node->get_node_name();
     auto symbol = std::make_shared<Symbol>(name, node, *token);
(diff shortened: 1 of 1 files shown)


src/visitors/symtab_visitor_helper.hpp
```diff
@@ -149,6 +149,9 @@ void SymtabVisitor::setup_symbol(ast::Node* node, NmodlType property) {
 
 void SymtabVisitor::add_model_symbol_with_property(ast::Node* node, NmodlType property) {
     auto token = node->get_token();
+    if (!token) { // token can be nullptr
+        return;
+    }
```
pramodk, line 154: @alkino : I believe in this case we need to add `node` to symbol table but with "default" `ModToken` object (i.e. token which will not have location information associated with it).

Otherwise, here we are just skipping adding symbol in the symbol table which means we will not find that symbol and that might be problematic somewhere else.

pramodk (changes requested):

pramodk (approved):

PR #889, opened 2022-07-01
Require C++17.

olupton:


src/pybind/pynmodl.cpp
```diff
@@ -138,12 +135,13 @@ void init_visitor_module(py::module& m);
...
 
-
 PYBIND11_MODULE(_nmodl, m_nmodl) {
     m_nmodl.doc() = "NMODL : Source-to-Source Code Generation Framework";
     m_nmodl.attr("__version__") = nmodl::Version::NMODL_VERSION;
 
-    py::class_<nmodl::PyNmodlDriver> nmodl_driver(m_nmodl, "NmodlDriver", nmodl::docstring::driver);
+    py::class_<nmodl::parser::NmodlDriver>(m_nmodl, "nmodl::parser::NmodlDriver");
+    py::class_<nmodl::PyNmodlDriver, nmodl::parser::NmodlDriver> nmodl_driver(
+        m_nmodl, "NmodlDriver", nmodl::docstring::driver);
```
olupton, line 144: Is there a simple "why"?

azure-pipelines.yml
```diff
@@ -321,7 +321,7 @@ stages:
       condition: succeeded()
       displayName: 'Install Python from python.org'
     - script: |
-        export MACOSX_DEPLOYMENT_TARGET=10.9
+        export MACOSX_DEPLOYMENT_TARGET=10.14
```
olupton, line 324: Was this needed @alexsavulescu? Or just part of a general update? (note I removed this change from the NEURON PR, but I can put it back in...)

pramodk: Ref: https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/

> To get C++11 and C++14 support, MACOSX_DEPLOYMENT_TARGET needs to be set to (at least) "10.9". By default, cibuildwheel already does this, building 64-bit-only wheels for macOS 10.9 and later.

> To use C++17 library features and link against the C++ runtime library, set MACOSX_DEPLOYMENT_TARGET to "10.13" or "10.14" (or higher) - macOS 10.13 offers partial C++17 support (e.g., the filesystem header is in experimental, offering #include <experimental/filesystem> instead of #include <filesystem>); macOS 10.14 has full C++17 support.

src/pybind/pynmodl.cpp
alexsavulescu: https://stackoverflow.com/questions/62854830/error-by-wrapping-c-abstract-class-with-pybind11

olupton: Thanks!

PR #896, opened 2022-07-12
Fix GPU code reductions

iomaganaris:
Only print atomic pragmas for updates of `rhs` and `d` vectors on GPU if mechanism is `POINT_PROCESS`.
Otherwise this is not needed since for other types of mechanisms the indexes of these vectors are not clashing with each other

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -57,7 +57,7 @@ void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType ty
 
 
 void CodegenAccVisitor::print_atomic_reduction_pragma() {
-    if (!info.artificial_cell) {
+    if (info.point_process) {
         printer->add_line("nrn_pragma_acc(atomic update)");
         printer->add_line("nrn_pragma_omp(atomic update)");
     }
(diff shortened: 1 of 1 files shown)


src/codegen/codegen_acc_visitor.cpp
```diff
@@ -57,7 +57,7 @@ void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType ty
 
 
 void CodegenAccVisitor::print_atomic_reduction_pragma() {
-    if (!info.artificial_cell) {
+    if (info.point_process) {
```
pramodk, line 60: Just a `pure nitpick` Ioannis: should we remove this `if` condition from here and put into where this function is called? i.e. `print_atomic_reduction_pragma` should just print the pragma and shouldn't have the extra logic? This way code/usage would be a bit clear?

pramodk (approved):
LGTM.

Just quickly checked in mod2c: 

https://github.com/BlueBrain/mod2c/blob/8565d3c178a195a489fae0623d6338c2e92cd1e5/src/mod2c_core/noccout.c#L153
https://github.com/BlueBrain/mod2c/blob/8565d3c178a195a489fae0623d6338c2e92cd1e5/src/mod2c_core/noccout.c#L981

which has:

```
if (point_process) {
			rhs_d_pnt_race("+=", "-=");
		}
```

PR #898, opened 2022-07-18
Install llvm 13.0.1 in MacOS builds in Azure

iomaganaris:
`brew` installs by default `LLVM 14` which has the following bug with CMake https://github.com/llvm/llvm-project/issues/53950
For `llvm` branch we just need `LLVM 13.0.0/13.0.1`


ohm314 (approved):
lgtm

PR #899, opened 2022-07-20
Use new formatting utility

tristan0x:
Bump hpc-coding-conventions submodule providing a new way to
format the code. It is not required to run CMake anymore.
Just use `cmake/hpc-coding-conventions/bin/format` utility.


pramodk (approved):
Nice! LGTM!

Issue #216, opened 2019-06-12
lvalue required as left operand of assignment


iomaganaris:
@nrnhines reported the following compilation error while trying to translate [`cabpump.mod`](https://github.com/neuronsimulator/nrn/blob/master/share/demo/release/cabpump.mod):
```
cabpump.cpp:326:52: error: lvalue required as left operand of assignment
                 (inst->ca+id*10) = cadifpmp_global.ca0;
```
The original file and a patched one that resolves some of the issues of this `mod file` can be found in [cabpump.zip](https://github.com/BlueBrain/nmodl/files/3281214/cabpump.zip).

PR #901, opened 2022-07-27
Fix use of array with floating variables

alkino:
When using an array we want the value and not the pointer on the value.

Generated code move from `inst->var + size * index` to `inst->var[size * index]` with `double* inst->var`


alkino:
Should close #216

pramodk (commented):
@alkino : would be nice to update one of the MOD file [here](https://github.com/BlueBrain/nmodl/tree/master/test/integration/mod) with array variable?

alkino:
There is still some other problems with array, making it difficult to compile.

iomaganaris (commented):
I believe this is a welcome addition. The branch needs rebasing and some of the unit tests need fixing after https://github.com/BlueBrain/nmodl/pull/924 though

alkino:
Not the right way to do

PR #904, opened 2022-08-05
Reorganise "global" variables to not be global

olupton:
See https://github.com/BlueBrain/CoreNeuron/pull/795 and https://github.com/BlueBrain/mod2c/pull/78.

TODO:
- [x] See if Eigen/PartialPivLU workaround can be simplified further.
- [x] Revert to upstream fmt repository if https://github.com/fmtlib/fmt/pull/3043 is merged in time.
- [x] Open an Eigen PR with my workarounds for `nvc++ -cuda` --> https://gitlab.com/libeigen/eigen/-/merge_requests/1041
- [x] ~Print a warning if an external fmt is used with nvhpc 22.3?~ --> https://github.com/BlueBrain/nmodl/pull/914
- [x] Check if https://github.com/neuronsimulator/nrn/issues/1505 has been addressed "by accident"

CI_BRANCHES:CORENEURON_BRANCH=pramodk/exclude-global-vars,NEURON_BRANCH=olupton/coreneuron-gpu-dynamic-loading


.gitmodules
```diff
@@ -12,7 +12,7 @@
 	url = https://github.com/gabime/spdlog
 [submodule "ext/fmt"]
 	path = ext/fmt
-	url = https://github.com/fmtlib/fmt.git
+	url = https://github.com/olupton/fmt.git
```
pramodk, line 15: Just fyi: for [similar reasons|https://github.com/fmtlib/fmt/pull/1581] in the past, we have created bbp fork: https://github.com/BlueBrain/fmt. If something doesn't merge sooner, we could use the bbp fork.

src/visitors/implicit_argument_visitor.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "ast/string.hpp"
+
+#include <cassert>
+#include <iostream>
+
+namespace nmodl {
+namespace visitor {
+
+// TODO should this also handle the case covered by
+// needs_neuron_thread_first_arg?
```
pramodk, line 18: Looking at `extern_definitions`, nothing comes to mind that needs to be changed (immediately).

.gitmodules
olupton: OK. Note that this fix is only needed for NVHPC 22.3, newer versions don't need it.

src/visitors/implicit_argument_visitor.cpp
olupton: I think what I meant here was that we now have two different ways of fudging extra arguments into function calls:
- this visitor, which I use for `nrn_ghk` and which actually modifies the AST
- this code in the code printing: https://github.com/BlueBrain/nmodl/blob/40f046c7aaf22af7e265efb52ea68f37195c4516/src/codegen/codegen_c_visitor.cpp#L1382-L1388

and I was wondering if we should consolidate and only use one approach. Originally I think I tried with the extra visitor because I was trying to only add `celsius` to the instance struct if it was actually used, so adding that information to the AST was useful.

.gitmodules
olupton: I pushed the workaround to https://github.com/BlueBrain/fmt; there's also a PR to upstream so we might be able to revert this part.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -2484,100 +2500,161 @@ void CodegenCVisitor::print_mechanism_global_var_structure() {
...
+                          float_type,
+                          info.thread_var_data_size);
         codegen_global_variables.push_back(make_symbol("thread_data_in_use"));
         auto symbol = make_symbol("thread_data");
         symbol->set_as_array(info.thread_var_data_size);
         codegen_global_variables.push_back(symbol);
     }
 
-    printer->add_line(fmt::format("{}int reset;", qualifier));
+    // TODO: remove this entirely?
```
pramodk, line 2530: Bit of background:

"thread" variables are a way in neuron to avoid race condition when variables are being updated by multiple threads (e.g. user declare something `GLOBAL` and nrnivmodl converts it to "thread variable" with `#threads` copies)

In context of coreneuron, this is not sufficient due to SIMD/SIMT execution. One solution I am thinking is that NEURON should convert those variables to "RANGE" in order to avoid race conditions.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -2949,34 +3018,41 @@ void CodegenCVisitor::print_thread_memory_callbacks() {
...
+    auto const value_initialise = print_initialisers ? "{}" : "";
     auto float_type = default_float_data_type();
     auto int_type = default_int_data_type();
     printer->add_newline(2);
-    printer->add_line("/** all mechanism instance variables */");
-    printer->start_block(fmt::format("struct {} ", instance_struct()));
+    printer->add_line("/** all mechanism instance variables and global variables */");
+    printer->fmt_start_block("struct {} ", instance_struct());
+    // TODO dynamically [don't] include this depending on whether it's used
+    printer->fmt_line("double celsius{};", print_initialisers ? "{celsius}" : "");
```
pramodk, line 3029: I think celsius should be in the symbol table and there is a "read/write" count per symbol that indicates if a variable is used in the mod file or not.

src/codegen/codegen_c_visitor.hpp
```diff
@@ -614,9 +614,11 @@ class CodegenCVisitor: public visitor::ConstAstVisitor {
     /**
      * Determine the variable name for a global variable given its symbol
      * \param symbol The symbol of a variable for which we want to obtain its name
+     * \param use_instance Should the variable be accessed via the (host-only)
+     * global variable or the instance-specific copy (also available on GPU).
```
pramodk, line 618: minor / could be a separate decision - as mentioned on slack, I was wondering if 2 copies is a good idea.

pramodk (commented):
I skim through this. Just few minor comments.

src/visitors/implicit_argument_visitor.cpp
olupton: Done.

src/codegen/codegen_c_visitor.hpp
olupton: This has now been changed. Global variables go in a plain data struct `Memb_list::global_variables[_size]` where CoreNEURON handles the host/device movement (as it does for `celsius` etc.). The instance struct now contains no data, only pointers to range variables (inside `ml->data`), true global variables (`celsius`), and mechanism global variables (i.e. `ml->global_variables`), and the translated mechanism .cpp copies this in/out lazily.

src/codegen/codegen_c_visitor.cpp
olupton: I think this comment was referring to `reset`, which never seems to be touched in code generated by NMODL.

src/codegen/codegen_c_visitor.cpp
olupton: Done in 07086dfa38e308699c86212cf98d5324217995bc.

src/visitors/implicit_argument_visitor.cpp
```diff
@@ -0,0 +1,56 @@
+/*************************************************************************
+ * Copyright (C) 2022 Blue Brain Project
+ *
+ * This file is part of NMODL distributed under the terms of the GNU
+ * Lesser General Public License. See top-level LICENSE file for details.
+ *************************************************************************/
+#include "visitors/implicit_argument_visitor.hpp"
+#include "ast/ast_decl.hpp"
```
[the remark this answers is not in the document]
olupton: I think by default it has special treatment for the header matching the current .cpp name. In any case, I wasn't trying to force a non-alphabetic sort order here.

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -461,6 +461,19 @@ void CodegenHelperVisitor::find_table_variables() {
     info.table_assigned_variables = psymtab->get_variables_with_properties(property);
 }
 
+void CodegenHelperVisitor::find_neuron_global_variables() {
+    // TODO: it would be nicer not to have this hardcoded list
+    using pair = std::pair<const char*, const char*>;
+    for (auto [var, type]: {pair{naming::CELSIUS_VARIABLE, "double"},
+                            pair{"secondorder", "int"},
+                            pair{"pi", "double"}}) {
```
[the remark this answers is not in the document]
olupton: I did try using that list, but (apart from the missing `pi` and `secondorder`) you then have to veto variables like `t` and `dt` that are typically there in the symbol table, but which are not globals that we need to include in the global structures that are being added in this PR.

In the end, the hardcoded list (which also provides the types, which we need to know) seemed simpler...

src/visitors/implicit_argument_visitor.cpp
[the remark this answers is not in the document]
pramodk: > I think by default it has special treatment for the header matching the current .cpp name

Ok, didn't know.

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1909,9 +1896,9 @@ void CodegenCVisitor::print_eigen_linear_solver(const std::string& float_type, i
 
 std::string CodegenCVisitor::internal_method_arguments() {
     if (ion_variable_struct_required()) {
-        return "id, pnodecount, inst, ionvar, data, indexes, thread, nt, v";
+        return "id, pnodecount, inst, ionvar, data, indexes, thread, nt, ml, v";
```
iomaganaris, line 1899: Since the `global` functions are accessible through the `inst`, do we need to also pass a pointer to `ml` as well? Maybe there is some other reason for this that I haven't understood.
Same question applies to other places were `ml` was added as well

src/codegen/codegen_c_visitor.cpp
```diff
@@ -3252,88 +3185,122 @@ std::string CodegenCVisitor::get_range_var_float_type(const SymbolType& symbol)
...
+                          global_struct_instance());
+        printer->fmt_line("assert(inst->{} == ml->global_variables);", naming::INST_GLOBAL_MEMBER);
+        printer->fmt_line("assert(ml->global_variables_size == sizeof({}));", global_struct());
+    };
+
+    printer->fmt_line(
+        "static inline void copy_instance_to_device(NrnThread* nt, Memb_list* ml, {} const* inst);",
+        instance_struct());
+    printer->fmt_line("static inline void delete_instance_from_device({}& inst);",
+                      instance_struct());
```
iomaganaris, line 3225: No strong opinion but maybe we could remove completely `device` related functions from the `C` codegen?

iomaganaris (commented):
Just few questions/comments while skimming through the changes

src/codegen/codegen_c_visitor.cpp
pramodk: As mentioned in slack,

When we have to access global or any other data, we have to get via ml->instance.
In nmodl, we have Instance struct but on the mod2c side, we have _threadargsproto_namespace as args to functions:

> #define _threadargsproto_namespace int _iml, int _cntml_padded, double* _p, coreneuron::Datum* _ppvar, coreneuron::ThreadDatum* _thread, coreneuron::NrnThread* _nt, double v

In order to reach to globals, we have now changed it to

> #define _threadargsproto_namespace int _iml, int _cntml_padded, double* _p, coreneuron::Datum* _ppvar, coreneuron::ThreadDatum* _thread, coreneuron::NrnThread* _nt, coreneuron::Memb_list* _ml, double v

As this prototype change is everywhere (including coreneuron side), we have to change this.

(at least this is what I remember)

src/codegen/codegen_c_visitor.cpp
pramodk: These are just empty definitions. This is then overridden in OpenACC visitor.

These are callback functions registered to coreneuron and hence they are defined in C visitor as well. Maybe we could register nullptr? But maybe they could be repurposed later for generic mechanism-related cleanup by just renaming.

PR #912, opened 2022-08-22
[LLVM] Instantiate CodegenLLVMVisitor of PyJIT with fast math and debug symbols options

iomaganaris:
- Set `add_debug_information` and `fast_math_flags` variables when constructing the `CodegenLLVMVisitor` of the PyJIT Driver according to the variables passed to the python object of `CodeGenConfig`
- Added related test


ohm314 (approved):
lgtm

Issue #888, opened 2022-06-30
Testing & Fixing issues with DBBS-Lab's MOD file collection


pramodk:
We have been working with Robin in order to run his model with CoreNEURON on GPU (see neuronsimulator/nrn/issues/1839). While testing the model, we have seen quite some issues in NMODL translation. The goal of this ticket is to compile mod files and fix translation+compilation issues.

### What to do?

* Install coreneuron + nmodl with sympy (`-DCORENRN_ENABLE_NMODL=ON "-DCORENRN_NMODL_FLAGS=sympy --analytic"`)
* Get mod file collection from DBBS Lab
```console
git clone https://github.com/dbbs-lab/dbbs-mod-collection.git
```
Few mod file has issues and we should do following:

1. following file is not required and incompatible with coreneuron

  ```console
  rm dbbs_mod_collection/dbbs_mod_collection/mod/glia__dbbs_mod_collection__gap_junction__parallel.mod 
  ```

2. `dbbs_mod_collection/mod/glia__dbbs_mod_collection__cdp5__CR.mod` file needs following change (state variable can't be in COMPARTMENT)

```console
diff --git a/dbbs_mod_collection/mod/glia__dbbs_mod_collection__cdp5__CR.mod b/dbbs_mod_collection/mod/glia__dbbs_mod_collection__cdp5__CR.mod
index 32df989..7edb9ef 100644
--- a/dbbs_mod_collection/mod/glia__dbbs_mod_collection__cdp5__CR.mod
+++ b/dbbs_mod_collection/mod/glia__dbbs_mod_collection__cdp5__CR.mod
@@ -100,6 +100,7 @@ ASSIGNED {
        cai       (mM)
        mgi     (mM)
        vrat    (1)
+       mg              (mM)^M
 }

 CONSTANT { cao = 2     (mM) }
@@ -110,7 +111,6 @@ STATE {
        : let it be ~1.5 - 2 orders of magnitude smaller than baseline level

        ca              (mM)    <1e-3>
-       mg              (mM)    <1e-6>

        Buff1           (mM)
        Buff1_ca        (mM)
@@ -223,7 +223,7 @@ LOCAL dsq, dsqvol  : can't define local variable in KINETIC block
                    :   or use in COMPARTMENT statement

 KINETIC state {
[...]

PR #913, opened 2022-08-23
Support for TABLE statements in FUNCTION blocks

alkino:
In mod2c a table works really differently from a function or from a procedure.
For example, `TABLE` in `FUNCTION` has an implicit variable name same as the function name.
In this PR we implement TABLE support for function same as mod2c.


pramodk (approved):
Overall the changes LGTM.

We will need some validation / integration test but that we could do in #888.

PR #919, opened 2022-08-30
Remove internal `ml` argument and move GPU code to Acc visitor

olupton:
Minor cleanup and reorganisation of code introduced in #904.


iomaganaris (approved):
LGTM 👍 
We should also update the `CoreNEURON` submodule

Issue #249, opened 2019-10-31
Add list of operators and functions supported in NMODL


pramodk:
- Updated documentation with the list of available operators and functions 'sin() cos() tan() .... + - * / % ...' supported in NMODL
- Extract this information from language grammar and lexer implementation

PR #920, opened 2022-09-01
Add a check for external definitions that are not thread safe

alkino:
Some functions are not thread safe and so not supported by CoreNeuron, handle them.


alkino:
And so we use some of them. What should we do?

src/visitors/semantic_analysis_visitor.cpp
```diff
@@ -78,5 +79,17 @@ void SemanticAnalysisVisitor::visit_destructor_block(const ast::DestructorBlock&
     /// -->
 }
 
+void SemanticAnalysisVisitor::visit_function_call(const ast::FunctionCall& node) {
+    /// <-- This code is for check 4
+    const auto& name = node.get_node_name();
+    if (details::is_external_definitions(name) && details::is_not_thread_safe(name)) {
+        logger->critical(
+            "SemanticAnalysisVisitor :: '{}' is not thread safe and incompatible with CoreNEURON",
```
iomaganaris, line 87: I'm wondering whether instead of this message we should print that the problem is that these functions are not implemented in CoreNEURON (which will need to change if at some point we change the `nmodl` translator of `NEURON` with this `NMODL`)

pramodk: I agree - the thread safety aspect is « implementation » aspect. I believe it’s not really about the method themselves.

Ioannis - as part of your work, we need to find out what needs to be done with these (legacy) functions.

And if you want to move this PR forward, throwing error as not implemented is better than saying not thread safe.

pramodk: @alkino @iomaganaris : something else relevant comes to my mind - if you look at the CI then it fails for [netstim.mod](https://github.com/BlueBrain/CoreNeuron/blob/master/coreneuron/mechanism/mech/modfile/netstim.mod) for the usage of `set_seed` and `exprand`. 

There are a few things to remember for thread safety: there are only certain blocks of the MOD file are called from the parallel region and they are: `INITIAL`, `BREAKPOINT`, `DERIVATIVE`, `KINETIC`, `NET_RECEIVE` (and something else?). If we are using non-thread-safe functions from these blocks (directly or indirectly) then only it's problem. Given this, I think we can bit improve this pass when to show a warning. For example:

*  `set_seed()`: this function is called from `PROCEDURE seed(x)` but this procedure is not called from `INITIAL`, `BREAKPOINT`, `DERIVATIVE`, `KINETIC` or `NET_RECEIVE`. This means this procedure is most likely called from the interpreter (hoc/python) side in a serial execution. Hence this usage is not really problematic.

* `exprand()` is called indirectly from  INITIAL/NET_RECEIVE->invl->erand so this usage is "potentially" problematic. But it's difficult to say for sure because `exprand` is preceded by the usage of VERBATIM block and hence this could be conditional usage for NEURON with ` #if !NRNBBCORE` etc.

What I am trying to say is that it might be possible to do this check bit better? I haven't thought about the implementation aspect. If it's difficult, we can leave this. But I thought it's worth mentioning.

alkino:
I found this quite related: https://github.com/BlueBrain/nmodl/issues/249

src/visitors/semantic_analysis_visitor.cpp
pramodk: @alkino : I am reviving this after 2 years! 

As NEURON code generation is ongoing, I think this PR is relevant and good to have!  If this PR is blocked for my earlier comment, especially `set_seed and exprand`, then I would say we can exclude these two functions and get the rest merged.

What do you think?

PR #922, opened 2022-09-01
Log a critical instead of an assert if we cannot find the block to so…

alkino:
…lved

Otherwise it only segmentation fault in Release

### DIFF
--- a/src/visitors/solve_block_visitor.cpp
+++ b/src/visitors/solve_block_visitor.cpp
@@ -11,6 +11,7 @@
 
 #include "ast/all.hpp"
 #include "codegen/codegen_naming.hpp"
+#include "utils/logger.hpp"
 #include "visitor_utils.hpp"
 
 namespace nmodl {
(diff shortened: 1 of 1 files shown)


alkino:
I'm not sure what we should do, how can we end the program there?
Is `exit(1)` good?

iomaganaris:
I was thinking that maybe based on [Exceptions versus assertions
](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170#exceptions_versus_assertions) an assertion fits here?

useranonym:
> I was thinking that maybe based on [Exceptions versus assertions](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170#exceptions_versus_assertions) an assertion fits here?

that's true but `assert` is removed in non-debug builds and hence the problem?

alkino:
@iomaganaris the problem of assert is that it is not executed with Release build. As this assertion is about conformance of mod files, it needs to be run all the time, as we cannot more trust the user input in Release that in Debug. I was running nmodl and i got a segfault without explanation on a buggy modfile.

pramodk (approved):

PR #923, opened 2022-09-01
Use C++ header instead of C one insite generated cpp files

alkino:


olupton (commented):
I believe the main difference here is that with the C++ versions the standard library functions are not required to be defined in the global namespace. So `std::log(...)` will definitely work, and `using std::log; log(...)` will definitely work, but `log(...)` might not any more.

So: do we always use `std::` when using functions from these headers?

See https://en.cppreference.com/w/cpp/header#C_compatibility_headers and https://stackoverflow.com/questions/15656434/include-cmath-vs-include-math-h-in-a-c-program

alkino:
On some platform functions are not defined in the global  namespace

PR #925, opened 2022-09-06
Codegen fixes for Eigen solvers with use of PROCEDURE using TABLE statement

pramodk:
* When we have TABLE statement used in a PROCEDURE and if such procedure
  is called from DERIVATIVE block then we end-up calling table statement
  related function from initialize() function:

  ```c++

        struct functor {
            NrnThread* nt;
            glia__dbbs_mod_collection__Ca__granule_cell_Instance* inst;
            int id, pnodecount;
            double v;
            Datum* indexes;
            double old_s, old_u;

            void initialize() {
                rate_glia__dbbs_mod_collection__Ca__granule_cell(id, pnodecount, inst, data, indexes, thread, nt, v, v);
                ...
  ```

  Here we are lacking `data` and `thread` variable as members in `struct functor`.

* This happens with the `glia__dbbs_mod_collection__Ca__granule_cell.mod` in #888

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -1848,8 +1848,10 @@ void CodegenCVisitor::visit_eigen_newton_solver_block(const ast::EigenNewtonSolv
     printer->end_block(2);
 
     printer->fmt_line(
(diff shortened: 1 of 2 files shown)


iomaganaris (approved):
LGTM 👌

PR #926, opened 2022-09-06
Remove token that have never be implemented in nrn: STEPPED / TERMINAL / PLOT / SENS / SECTION

alkino:
Those TOKENS are STEPPED / TERMINAL / PLOT / SENS / SECTION

Following https://github.com/neuronsimulator/nrn/pull/1974


pramodk (approved):
LGTM 🚀

PR #929, opened 2022-09-09
Fix duplicate declaration of conductance variable g

pramodk:
* when user defined g as a range variable in a non-threadsafe mod file, then instance structure was declaring g as a member variable twice
* now, check if variable is already defined before adding new declaration
* this is realted to #888 with mod file glia__dbbs_mod_collection__Nav1_1__0.mod

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -863,11 +863,10 @@ std::vector<SymbolType> CodegenCVisitor::get_float_variables() {
                                           : naming::CONDUCTANCE_VARIABLE;
 
         // make sure conductance variable like `g` is not already defined
-        auto iter =
-            std::find_if(variables.begin(), variables.end(), [&name](const SymbolType& sym) {
-                return name == sym->get_name();
-            });
-        if (iter == variables.end()) {
+        if (auto r = std::find_if(variables.cbegin(),
+                                  variables.cend(),
+                                  [&](const auto& s) { return name == s->get_name(); });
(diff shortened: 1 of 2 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -857,11 +857,21 @@ std::vector<SymbolType> CodegenCVisitor::get_float_variables() {
...
-        variables.push_back(make_symbol(name));
+
+        // make sure conductance variable like `g` is not already defined
+        auto iter =
+            std::find_if(variables.begin(), variables.end(), [&name](const SymbolType& sym) {
+                return name == sym->get_name();
+            });
+        if (iter == variables.end()) {
+            variables.push_back(make_symbol(name));
+        }
```
alkino, line 872: C++17 !

```c++
if (auto r = std::find_if(variables.cbegin(), variables.cend(), [&](const SymbolType& s) { return name == s->get_name(); }); r == variables.cend()) {
    variables.push_back(make_symbol(name));
};
```

alkino: We could maybe add range-v3 as a depends:

```c++
if (auto r = std::ranges::find_if(variables, [&](const SymbolType& s) { return name == s->get_name();}); r == variables.cend()) {
    variables.push_back(make_symbol(name));
}
```

pramodk: > C++17 !

thanks!

iomaganaris (approved):
LGTM 👍

PR #934, opened 2022-09-20
Store all nodes for a specific symbol

alkino:
This way we can support the anti-feature of nrn having a function and a range variable having the same name

### DIFF
--- a/src/language/templates/pybind/pysymtab.cpp
+++ b/src/language/templates/pybind/pysymtab.cpp
@@ -200,7 +200,7 @@ void init_symtab_module(py::module& m) {
         .def("get_id", &Symbol::get_id)
         .def("get_status", &Symbol::get_status)
         .def("get_properties", &Symbol::get_properties)
-        .def("get_node", &Symbol::get_node)
+        .def("get_node", [](const std::shared_ptr<Symbol>& s){ return s->get_nodes().front(); })
         .def("get_original_name", &Symbol::get_original_name)
         .def("get_name", &Symbol::get_name)
         .def("has_any_property", &Symbol::has_any_property)
--- a/src/language/templates/pybind/pysymtab.cpp
+++ b/src/language/templates/pybind/pysymtab.cpp
@@ -201,6 +201,7 @@ void init_symtab_module(py::module& m) {
         .def("get_status", &Symbol::get_status)
         .def("get_properties", &Symbol::get_properties)
         .def("get_node", [](const std::shared_ptr<Symbol>& s){ auto n = s->get_nodes(); return n.empty() ? nullptr : n.front(); })
+        .def("get_nodes", &Symbol::get_nodes)
         .def("get_original_name", &Symbol::get_original_name)
         .def("get_name", &Symbol::get_name)
         .def("has_any_property", &Symbol::has_any_property)
--- a/src/language/templates/pybind/pysymtab.cpp
+++ b/src/language/templates/pybind/pysymtab.cpp
@@ -200,7 +200,7 @@ void init_symtab_module(py::module& m) {
         .def("get_id", &Symbol::get_id)
         .def("get_status", &Symbol::get_status)
         .def("get_properties", &Symbol::get_properties)
-        .def("get_node", [](const std::shared_ptr<Symbol>& s){ return s->get_nodes().front(); })
+        .def("get_node", [](const std::shared_ptr<Symbol>& s){ auto n = s->get_nodes(); return n.empty() ? nullptr : n.front(); })
         .def("get_original_name", &Symbol::get_original_name)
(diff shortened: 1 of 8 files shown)


src/language/templates/pybind/pysymtab.cpp
```diff
@@ -200,7 +200,7 @@ void init_symtab_module(py::module& m) {
         .def("get_id", &Symbol::get_id)
         .def("get_status", &Symbol::get_status)
         .def("get_properties", &Symbol::get_properties)
-        .def("get_node", &Symbol::get_node)
```
alkino, line 203: we have to think about that

src/symtab/symbol.hpp
```diff
@@ -236,10 +239,17 @@ class Symbol {
...
-        return node;
+    void add_node(ast::Ast* node) noexcept {
+        nodes.push_back(node);
     }
 
+    std::vector<ast::Ast*> get_nodes() const noexcept {
+        return nodes;
+    }
+
+    std::vector<ast::Ast*> get_nodes_by_token(
```
pramodk, line 250: nitpick: reaming this to `get_nodes_by_type` would sound better?

pramodk (commented):
overall looks fine! as discussed, adding unit test under visitor would be great!

src/language/templates/pybind/pysymtab.cpp
```diff
@@ -200,7 +200,7 @@ void init_symtab_module(py::module& m) {
...
+        .def("get_node", [](const std::shared_ptr<Symbol>& s){ auto n = s->get_nodes(); return n.empty() ? nullptr : n.front(); })
```
ohm314, line 203:
```suggestion
        .def("get_nodes", [](const std::shared_ptr<Symbol>& s){ auto n = s->get_nodes(); return n.empty() ? nullptr : n.front(); })
```

src/visitors/inline_visitor.cpp
```diff
@@ -210,25 +211,21 @@ void InlineVisitor::visit_function_call(FunctionCall& node) {
         return;
     }
 
-    auto function_definition = symbol->get_node();
-    if (function_definition == nullptr) {
+    auto nodes = symbol->get_nodes_by_type(
```
ohm314, line 214: Since you can have multiple block assigned to one symbol, should you now not iterate through all of them and inline them all? or how do you make sure all gets inlined?

src/visitors/solve_block_visitor.cpp
```diff
@@ -47,7 +47,8 @@ ast::SolutionExpression* SolveBlockVisitor::create_solution_expression(
         throw std::runtime_error(
             fmt::format("SolveBlockVisitor :: cannot find the block '{}' to solve it", block_name));
     }
-    auto node_to_solve = solve_node_symbol->get_node();
+    auto nodes_to_solve = solve_node_symbol->get_nodes();
+    auto node_to_solve = nodes_to_solve.front();
```
ohm314, line 51:
```suggestion
    auto node_to_solve = solve_node_symbol->get_nodes().front();
```

ohm314 (commented):
Generally looks good, but maybe a few minor details to fix.

src/visitors/inline_visitor.cpp
alkino: If I try to add a procedure and a function with the same name, it is not possible, because when entering a scope (a block) we try to insert a new symbol only if it does not exist.

src/symtab/symbol.hpp
```diff
@@ -17,6 +17,7 @@
 
 #include "lexer/modtoken.hpp"
 #include "symtab/symbol_properties.hpp"
+#include <ast/ast_decl.hpp>
```
pramodk, line 20: just a nitpick: in order to prefer/prioritizes headers in the current working (than system dirs), we have been using `""` than `<>`. I think Tristan cleaned that in the past.

src/symtab/symbol_table.cpp
```diff
@@ -202,12 +200,14 @@ std::shared_ptr<Symbol> ModelSymbolTable::lookup(const std::string& name) {
...
                                     const std::shared_ptr<Symbol>& second,
                                     bool redefinition) {
-    auto node = first->get_node();
+    auto nodes = first->get_nodes();
     std::string name = first->get_name();
     auto properties = to_string(second->get_properties());
     std::string type = "UNKNOWN";
-    if (node != nullptr) {
-        type = node->get_node_type_name();
+    // Here we take the first one, because this is a redefinition
```
pramodk, line 207:
```suggestion
```

pramodk (commented):
few minor nitpicks otherwise good to go! 🚀

pramodk (approved):
(should have approved it with the previous review)

Issue #927, opened 2022-09-06
KINETIC block with function call result into sympy exception: SympySolverVisitor :: solve_non_lin_system python exception: 'Symbol' object is not callable


pramodk:
Consider a MOD file [glia__dbbs_mod_collection__Na__granule_cell.mod](https://github.com/dbbs-lab/dbbs-mod-collection/blob/master/dbbs_mod_collection/mod/glia__dbbs_mod_collection__Na__granule_cell.mod) which has:

```python

FUNCTION alfa(v(mV))(/ms){ 
	alfa = Q10*Aalfa*exp(v/Valfa) 
}

FUNCTION beta(v(mV))(/ms){ 
	beta = Q10*Abeta*exp(-v/Vbeta) 
}

FUNCTION teta(v(mV))(/ms){ 
	teta = Q10*Ateta*exp(-v/Vteta) 
}
 

KINETIC kstates {
	: 1 riga
	~ C1 <-> C2 (n1*alfa(v),n4*beta(v))
	~ C2 <-> C3 (n2*alfa(v),n3*beta(v))
	~ C3 <-> C4 (n3*alfa(v),n2*beta(v))
	~ C4 <-> C5 (n4*alfa(v),n1*beta(v))
	~ C5 <-> O  (gamma,delta)
	~  O <-> OB (epsilon,teta(v))
	
	: 2 riga
	~ I1 <-> I2	(n1*alfa(v)*a,n4*beta(v)*b)
	~ I2 <-> I3	(n2*alfa(v)*a,n3*beta(v)*b)
	~ I3 <-> I4	(n3*alfa(v)*a,n2*beta(v)*b)
	~ I4 <-> I5 (n4*alfa(v)*a,n1*beta(v)*b)
	~ I5 <-> I6 (gamma,delta)
	
	: connette 1 riga con 2 riga
	~ C1 <-> I1 (Con,Coff)
	~ C2 <-> I2 (Con*a,Coff*b)
	~ C3 <-> I3 (Con*a^2,Coff*b^2)
	~ C4 <-> I4 (Con*a^3,Coff*b^3)
	~ C5 <-> I5 (Con*a^4,Coff*b^4)
	~  O <-> I6 (Oon,Ooff)
	
	CONSERVE C1+C2+C3+C4+C5+O+OB+I1+I2+I3+I4+I5+I6=1
}
```

when we run this via NMODl we get:

```console
[...]

PR #935, opened 2022-09-22
Apply user_functions to vecFcode too

alkino:
When functions inside solve block are not inlined sympy cannot find them. The problem is already fixed for vecJcode, do the same.

Related to #927

### DIFF
--- a/nmodl/ode.py
+++ b/nmodl/ode.py
@@ -228,7 +228,7 @@ def solve_lin_system(eq_strings, vars, constants, function_calls, tmp_unique_pre
             )
             for var, expr in sub_exprs:
                 new_local_vars.append(sp.ccode(var))
-                code.append(f"{var} = {sp.ccode(expr.evalf())}")
+                code.append(f"{var} = {sp.ccode(expr.evalf(), user_functions=custom_fcts)}")
             solution_vector = simplified_solution_vector[0]
         for var, expr in zip(state_vars, solution_vector):
             code.append(f"{sp.ccode(var)} = {sp.ccode(expr.evalf(), contract=False, user_functions=custom_fcts)}")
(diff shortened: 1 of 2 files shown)


cattabiani:
may you add a CI test that crashes with the previous version? It should be simple: with sympy 1.9 `beta(x)` was transformed to `beta(x,x)`. This probably cannot compile.

alkino:
This patch is not fixing this bug. But the bug that not inlined functions are not found.

cattabiani:
ok, it was another bug. I sincerly do not know how this impacts nmodl. I do not know what it should do. I think you need someone else for the review

alkino:
Without the patch, the following file fail
```
NEURON {
    SUFFIX glia
}

STATE {
    C1
    C2
}

BREAKPOINT {
    SOLVE kstates METHOD sparse
}

FUNCTION alfa(v(mV)) {
    alfa = exp(v)
}

FUNCTION tau(v(mV)) {
    tau = exp(-v)
}

KINETIC kstates {
    ~ C1 <-> C2 (alfa(v), tau(v))
}
```

Because functions `alfa` and `tau` are not known, so sympy return
```
vecFcode=['F[0] = /* Not supported in C: */
/* alfa */
/* tau */
-X[0]*dt*alfa(v) - X[0] + X[1]*dt*pow(tgamma(v), 2)/tgamma(2*v) + old_C1',
'F[1] = /* Not supported in C: */
/* alfa */
/* tau */
X[0]*dt*alfa(v) - X[1]*dt*pow(tgamma(v), 2)/tgamma(2*v) - X[1] + old_C2']
```

And our internal parser don't know C comments

cattabiani:
it looks like it actually solved the problem. Are you sure to the left you do not have `F[0]` and `F[1]` and somewhere someone leaves the `'`?

pramodk:
> it looks like it actually solved the problem.

That's my understanding as well. By providing `user_functions`, sympy no loner produce C comments and avoids errors in NMODL parser.

@alkino : snout the failing mod file we had - maybe we can add that as a unit test? (in the unit test, we will not run Inline visitor and then it should cover the use case)

pramodk (approved):
we can merge #937 to this branch first?

alkino:
Those are 2 differents issues

iomaganaris (approved):
LGTM

PR #937, opened 2022-09-26
Rename protected functions for sympy

alkino:
Some functions are protected inside sympy, so if the user defined an own function with the same name we rename it

### DIFF
--- a/nmodl/ode.py
+++ b/nmodl/ode.py
@@ -8,6 +8,7 @@
 from importlib import import_module
 
 import sympy as sp
+import re
 
 # import known_functions through low-level mechanism because the ccode
 # module is overwritten in sympy and contents of that submodule cannot be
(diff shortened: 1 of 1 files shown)


nmodl/ode.py
```diff
@@ -25,9 +26,12 @@
     raise ImportError(f"Requires SympPy version >= 1.2, found {major}.{minor}")
 
 def _get_custom_functions(fcts):
+    forbidden_var = ["beta", "gamma"]
```
pramodk, line 29: this could be somewhere at the top level. And also, there more functions than 2 IIRC. (Alessandro listed those).

nmodl/ode.py
```diff
@@ -270,8 +287,15 @@ def solve_non_lin_system(eq_strings, vars, constants, function_calls):
...
     """
 
-    eqs, state_vars, sympy_vars = _sympify_eqs(eq_strings, vars, constants)
+    forbidden_var = ["beta", "gamma"]
+    print(function_calls)
+    for c in function_calls:
+        if c in forbidden_var:
+            r = re.compile(r"\b{}\b\s*\(".format(c))
+            f = f"_sympy_{c}_fun("
+            eq_strings = [re.sub(r, f, x) for x in eq_strings]
```
pramodk, line 296: nitpick: factor into a function to avoid duplication?

iomaganaris (commented):
Could we add a related test as well?

alkino:
Merge into #935

PR #939, opened 2022-09-29
Remove FIRST / LAST / PARTIAL / IFERROR

alkino:
* Remove unnecessary IFERROR production in SolveBlock (https://github.com/BlueBrain/nmodl/pull/940)
* See https://github.com/neuronsimulator/nrn/issues/1957


src/symtab/symbol_properties.hpp
```diff
@@ -180,9 +180,6 @@ enum class NmodlType : enum_type {
     /// constant variable
     constant_var = 1L << 21,
 
-    /// Partial Block
-    partial_block = 1L << 22,
```
pramodk, line 184: @alkino: as we removed 22, maybe we should shift the next symbols (i.e. avoid gap)? I will push the small change.

pramodk (approved):
LGTM

PR #942, opened 2022-10-01
Remove unnecessary NMODL constructs: MATCH, RESET, FORALL, MODEL_LEVEL, variables in THREADSAFE 

alkino:
See discussion in neuronsimulator/nrn/issues/1957


pramodk (approved):
LGTM

PR #961, opened 2022-10-11
Add support for MUTEXLOCK/MUTEXUNLOCK/PROTECT

alkino:


src/codegen/codegen_c_visitor.cpp
```diff
@@ -2176,6 +2197,13 @@ void CodegenCVisitor::print_nmodl_constants() {
     }
 }
 
+void CodegenCVisitor::print_mutex_decl() {
+    if (!info.mutex_used) {
+        return;
+    }
+
+    printer->add_line("std::unique_ptr<std::mutex> nmodlmutex;", 2);
```
pramodk, line 2205: @alkino : FOR PROTCT/MUTEX: I am thinking we should prefer [`omp critical mechanism_name`](https://www.openmp.org/spec-html/5.0/openmpsu89.html) (or omp mutex) here for two reasons 1) just for consistency with openmp used in coreneuron (even though we know it's compatible with std::mutex) 2) all profiling/analysis tools know how to report the performance metrics associated.

pramodk:
@alkino : I think there are simple changes remaining on this PR?

iomaganaris (approved):

pramodk (approved):
LGTM

test/unit/codegen/codegen_cpp_visitor.cpp
```diff
@@ -422,3 +422,34 @@ SCENARIO("Check code generation for TABLE statements", "[codegen][array_variable
...
+                MUTEXLOCK
+                tmp = 11
+                MUTEXUNLOCK
+                PROTECT tmp = 12
+            }
+        )";
+
+        THEN("Code with OpenMP critical sections is generated") {
+            auto const generated = get_cpp_code(nmodl_text);
+            std::string expected_code = R"(#pragma omp critical TEST {
```
pramodk, line 446: @alkino : we missed this in review! the [valid syntax](https://www.openmp.org/spec-html/5.0/openmpsu89.html) for named critical section is : `#pragma omp critical (name)` i.e. `( )` are missing and hence it's compilation error!

alkino: oupsy

alkino: is there no test for that?

iomaganaris: We should have added the keyword in one of the mod files here: https://github.com/BlueBrain/nmodl/tree/master/test/integration/mod that get compiled in the CI

PR #963, opened 2022-10-11
Remove ISPC codegen backend

alkino:
Here is a desc

CI_BRANCHES:CORENEURON_BRANCH=cornu/remove_ispc


pramodk:
Initially, we thought ISPC would be a good backend to target different platforms including GPUs and CPUs.
A few years ago NVIDIA backend got removed. When SVE becomes common on ARM HPC platforms, we asked about the possible support but nothing has happened on that front (https://github.com/ispc/ispc/issues/1947). So today, ISPC  backend could be useful on Intel platforms only. But this still excludes solvers based on Eigen etc.

With the free availability of intel-oneapi as well as [ARM](https://community.arm.com/arm-community-blogs/b/high-performance-computing-blog/posts/arm-compilers-and-libraries-for-hpc-now-free) compiler toolchains, the relevance of ISPC is reduced. Given this, I think it's not a bad idea to clean this up now.

pramodk (approved):
Good to merge once CI is green!

src/codegen/codegen_c_visitor.cpp
```diff
@@ -1131,20 +1119,6 @@ bool CodegenCVisitor::nrn_cur_reduction_loop_required() {
 }
 
 
-/**
- * \details For CPU backend we iterate over all node counts.
- */
-void CodegenCVisitor::print_channel_iteration_block_begin(BlockType type) {
```
ohm314, line 1137: I would not remove those - they can be useful in another case and they dont' hurt.

ohm314 (changes requested):
Looks all good, except for one thign that I woudl keep.
Additionally, I think in the CMakefiles there was some extra stuff to recognize the ISPC compiler that can be removed now.

src/codegen/codegen_c_visitor.cpp
alkino: @pramodk I told me that you prefer them to be remove. Let's debate (I don't care)

pramodk: For context - this `iteration block begin` was added at the beginning for **CPU OpenMP** backend where we were generating code with nested for loops. The idea was we can do nested-openmp parallelism i.e. once at top cell levels (NrnThreads) and then at individual mechanism levels (all instances of particular mechanism). This was never used/implemented.

I think it's a good idea to remove this now because 1) we removed OpenMP CPU backend months ago 2) if we will do fine-grained parallelism then (most likely) it will be more like multi-split that neuron does today 3) doing parallelism at the mechanism is a bit more complicated as we have to handle synapses sharing the same location or ion accumulation. And then `#pragma omp atomic` inside `#prgma simd / ivdep` doesn't work 😒.

Considering these aspects and simplifying the code, I suggested removing these empty functions. In case we will implement such a thing, it would be easy to bring back.

ohm314: OK, makes sense (we did use them also in the ISPC context), but as you said we don't have any use right now and they are unlikely to be used in the future - I just thought that it's just a couple lines of code that are generic enough that they "might" be useful.
Let's remove!

ohm314 (commented):
lgtm

PR #964, opened 2022-10-11
Remove Eigen::PartialPivLU and replace it with Crout LU decomposition

### LINKED ISSUE #943: Remove dependency with Eigen (and use internal Crout solver implementation instead)
We have seen quite some issues with Eigen-based solvers when it comes to OpenACC/OpenMP offload on GPUs. In order to simplify the implementation and reduce the maintenance burden, we should evaluate and then remove Eigen dependency from NMODL. 

As part of this ticket:

* Check the existing of Eigen
* Confirm if we can replace the usage by Crout solver
* Have implementation without Eigen
* Measure performance against standard benchmarks - channel-benchmark, olfactory-bulb and reduced-dentate
* Merge the implementation! 

Note that some was already done in #690.

kotsaloscv:
Remove Eigen::PartialPivLU and replace it with Crout LU decomposition

closes #943


src/solver/crout/crout.hpp
```diff
@@ -0,0 +1,147 @@
...
+        for (j = k + 1; j < n; j++) {
+            *(p_k + j) /= *(p_k + k);
+        }
+
+        // update remaining matrix
+        for (i = k + 1, p_row = p_k + n; i < n; p_row += n, i++)
+            for (j = k + 1; j < n; j++)
+                *(p_row + j) -= *(p_row + k) * *(p_k + j);
+    }
+    // return 0;
```
pramodk, line 83: there is no return, right?
```suggestion
```

src/solver/crout/crout.hpp
```diff
@@ -0,0 +1,147 @@
...
+ */
+#if defined(CORENEURON_ENABLE_GPU) && !defined(DISABLE_OPENACC)
+nrn_pragma_acc(routine seq)
+nrn_pragma_omp(declare target)
+#endif
+template <typename T>
+EIGEN_DEVICE_FUNC inline void Crout(int n, T* A, int* pivot) {
+    int i, j, k;
+    T *p_k, *p_row, *p_col;
+    T max;
```
pramodk, line 46: can I be `nitpicky`? : declaring these variables at local scope would be nice? (e.g. when I see `j, k` being at the top level, I was wondering whether their values are used across internal for loops)

src/solver/crout/crout.hpp
```diff
@@ -0,0 +1,147 @@
...
+ */
+#if defined(CORENEURON_ENABLE_GPU) && !defined(DISABLE_OPENACC)
+nrn_pragma_acc(routine seq)
+nrn_pragma_omp(declare target)
+#endif
+template <typename T>
+EIGEN_DEVICE_FUNC inline void solveCrout(int n, T* LU, T* B, T* x, int* pivot) {
+    int i, k;
+    T* p_k;
+    T dum;
```
pramodk, line 109: same

src/solver/crout/crout.hpp
```diff
@@ -0,0 +1,147 @@
...
+    // assumed to be 1.0.
+    for (k = n - 1, p_k = LU + n * (n - 1); k >= 0; k--, p_k -= n) {
+        if (pivot[k] != k) {
+            dum = B[k];
+            B[k] = B[pivot[k]];
+            B[pivot[k]] = dum;
+        }
+        for (i = k + 1; i < n; i++)
+            x[k] -= x[i] * *(p_k + i);
+        // if (*(p_k + k) == 0.0) return -1;
```
pramodk, line 137: remove or this requires some error propagation?

src/solver/newton/newton.hpp
```diff
@@ -73,13 +71,17 @@ EIGEN_DEVICE_FUNC int newton_solver(Eigen::Matrix<double, N, 1>& X,
...
-        // update X use in-place LU decomposition of J with partial pivoting
-        // (suitable for any N, but less efficient than .inverse() for N <=4)
-        X -= Eigen::PartialPivLU<Eigen::Ref<Eigen::Matrix<double, N, N>>>(J).solve(F);
-#endif
+        // In Eigen the default storage order is ColMajor.
+        // Crout's implementation requires matrices stored in RowMajor order (C-style arrays).
+        // Therefore, the transposeInPlace is critical such that the data() method to give the rows
+        // instead of the columns.
+        if (!J.IsRowMajor)
+            J.transposeInPlace();
```
pramodk, line 79: to clarify - when / how many times this transpose gets called?

pramodk (commented):
Some very minor comments otherwise look good to me!

One question for the performance aspect though - have we checked this via neuron paper models (channel-benchmark and olfactory-bulb model) and see if the execution time of respective MOD files block (the ones with sparse/derivimplicit solver) is comparable to old implementation?

docs/notebooks/nmodl-linear-solver.ipynb
```diff
@@ -39,7 +39,12 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.7"
+   "version": "3.6.8"
+  },
+  "vscode": {
+   "interpreter": {
+    "hash": "916dbcbb3f70747c44a77c7bcd40155683ae19c65e1c03b4aa3499c5328201f1"
```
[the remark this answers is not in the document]
kotsaloscv: the notebook was not cleaned at the first place and this is why there are these hashes. I am going to clean it :)

src/solver/crout/crout.hpp
kotsaloscv: exactly. This is intentional. Though, the algorithm is designed to return 0 for success and -1 for failure (singular matrix). Do you think that it is better to have them return something instead, even if we do not have any mechanism to catch errors?

src/solver/newton/newton.hpp
kotsaloscv: essentially every time, since the default storage order in Eigen is ColMajor.

kotsaloscv:
> Some very minor comments otherwise look good to me!
> 
> One question for the performance aspect though - have we checked this via neuron paper models (channel-benchmark and olfactory-bulb model) and see if the execution time of respective MOD files block (the ones with sparse/derivimplicit solver) is comparable to old implementation?

In terms of performance, I have just added timers in the unit test (measuring the time of eigen solvers and crout ones). These are the results I just got:
```
eigen_timing [ms] : 20.3689
crout_timing [ms] : 18.3714
```
Keep in-mid that crout includes the transposeInPlace from eigen.

Therefore, performance-wise I am expecting same or even better results. 

(I will push later my changes)

src/solver/crout/crout.hpp
kotsaloscv: I just wanted to be as close as possible to the legacy code. Therefore, I would prefer to leave it as is ;)

iomaganaris:
I run the CoreNEURON performance metrics CI and it seems like the time got reduced since the last run slightly but last CPU runtimes are a bit unstable. You can see the output here https://bbpgitlab.epfl.ch/hpc/cellular/coreneuron-perf-tests/-/jobs/391196. It run with commit https://github.com/BlueBrain/nmodl/pull/964/commits/3a2c7c27666a185f57c701f49a6c3c655b7cb28e. Maybe we should rerun it with the latest changes before merging and see whether the runtime is closer to the previous average CPU runtimes?

cmake/CompilerHelper.cmake
```diff
@@ -4,6 +4,9 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "PGI" OR CMAKE_CXX_COMPILER_ID MATCHES "NVHPC")
   # CMake adds standard complaint PGI flag "-A" which breaks compilation of of spdlog and fmt
   set(CMAKE_CXX14_STANDARD_COMPILE_OPTION --c++14)
 
+  # Avoid errors related to "excessive recursion at instantiation of function ..."
+  list(APPEND NMODL_EXTRA_CXX_FLAGS "-Wc,--pending_instantiations=0")
```
olupton, line 8: What code needs this, out of curiosity?

src/solver/crout/crout.hpp
```diff
@@ -0,0 +1,176 @@
...
+ * @return 0 for SUCCESS || -1 for FAILURE (The matrix A is singular)
+ */
+#if defined(CORENEURON_ENABLE_GPU) && !defined(DISABLE_OPENACC)
+nrn_pragma_acc(routine seq)
+nrn_pragma_omp(declare target)
+#endif
+template <typename T>
+EIGEN_DEVICE_FUNC inline int Crout(int n, T* A, int* pivot) {
+    int i, j, k;
+    T *p_k, *p_row, *p_col;
```
olupton, line 70:
```suggestion
    T *p_k{}, *p_row{}, *p_col{};
```

olupton (approved):
LGTM, just a few tiny comments!

cmake/CompilerHelper.cmake
kotsaloscv: it is needed for Eigen's transposeInPlace() when compiled with nvhpc, e.g.
https://github.com/BlueBrain/nmodl/blob/kotsalos/eigen_removal/src/solver/newton/newton.hpp#L79

src/solver/newton/newton.hpp
olupton: As we generate the code then couldn't we consistently use RowMajor when the Crout solver is (going to be) used?

kotsaloscv: for now I would leave Eigen in its default state. Once we decide to completely remove Eigen, then we make this change. Performance-wise, Crout-branch even with this transposeInPlace is consistently faster compared to Eigen.

docs/notebooks/nmodl-linear-solver.ipynb
```diff
@@ -39,7 +32,12 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.7"
+   "version": "3.9.7"
```
iomaganaris, line 35: FYI with
```
jupyter nbconvert --ClearOutputPreprocessor.enabled=True --ClearMetadataPreprocessor.enabled=True --clear-output --inplace docs/notebooks/nmodl-linear-solver.ipynb
```
you can clear all these python and vscode related details from jupyter notebooks

pramodk (approved):
LGTM

pramodk:
```console
PYTHONPATH=/home/vsts/work/1/s/nrn/build/lib::/home/vsts/work/1/s/nrn/build/lib/python MODLUNIT=/home/vsts/work/1/s/nrn/build/share/mod2c/nrnunits.lib /home/vsts/work/1/s/nrn/build/bin/nmodl x86_64/corenrn/mod2c/hh.mod -o x86_64/corenrn/mod2c/ host --c passes --inline sympy --analytic
Segmentation fault (core dumped)
make[3]: *** [/home/vsts/work/1/s/nrn/build/share/coreneuron/nrnivmodl_core_makefile:253: x86_64/corenrn/mod2c/halfgap.cpp] Error 139
make[3]: *** Waiting for unfinished jobs....
```

Is something worth looking for segfault?

kotsaloscv:
@pramodk shall we merge it?
I have added the error checking.
All the tests are passing, see as well: https://github.com/BlueBrain/CoreNeuron/pull/871

pramodk (approved):
LGTM. Good to go!  (as squash & merge)

Issue #850, opened 2022-04-27
Simplify C code generation backend and remove experimental/unnecessary options


pramodk:
While trying to redo code generation for `nrn_cur` and `nrn_state` in the LLVM branch, I notice that there are old/experimental options in the C backend that could be just removed at this stage. They will be useful `at some point in the future` but the current coreneuron setup doesn't allow us to use them anyway. So they are just making 🍝 now!

List of such options/flags (WIP)

- [x] Remove OpenMP backend ? #858 
- [x] channel_task_dependency_enabled #858 
- [x] print_channel_iteration_tiling_block_begin and print_channel_iteration_block_end #963 
- [x] Not sure if declarations using `k_const()` is really useful #882
- [x] Same situation with `ptr_type_qualifier()` #965 
- [x] Unused CUDA backend #962 
- [x] Unused ISPC backend #963 

The idea here is to make a list of all such cleanup and simplify the C visitor. This will make things cleaner, easy to understand and implement the code generation for NEURON easier.

alkino:
Everything has been removed. Let's close?

PR #965, opened 2022-10-17
Remove useless ptr_type_qualifier()

alkino:
Once ISPC backend have been removed (#963), this function is constant, so remove it.

Part of #850

CI_BRANCHES:CORENEURON_BRANCH=master

### DIFF
--- a/src/codegen/codegen_c_visitor.cpp
+++ b/src/codegen/codegen_c_visitor.cpp
@@ -3074,8 +3074,7 @@ void CodegenCVisitor::print_mechanism_range_var_structure(bool print_initialiser
         auto name = var.symbol->get_name();
         if (var.is_index || var.is_integer) {
             auto qualifier = var.is_constant ? "const " : "";
-            printer->fmt_line(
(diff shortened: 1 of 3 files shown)


src/codegen/codegen_c_visitor.cpp
```diff
@@ -3063,9 +3058,8 @@ void CodegenCVisitor::print_mechanism_range_var_structure(bool print_initialiser
 
     for (auto const& [var, type]: info.neuron_global_variables) {
         auto const name = var->get_name();
-        printer->fmt_line("{}* {}{}{};",
+        printer->fmt_line("{}* __restrict__ {}{};",
```
pramodk, line 3061: @alkino : I would say even get rid of this restrict. Nowadays we rely everywhere on something like `pragma ivdep` or `pragma simd`. The `restrict` part is not sufficient.

So if you prefer to remove it, I am Ok with it.

pramodk (approved):

src/codegen/codegen_c_visitor.cpp
alkino: done

PR #966, opened 2022-10-17
Independent is now useless, only check if variable is valid

alkino:
INDEPENDENT are now doing nothing in nrn, do the same here

### DIFF
--- a/test/unit/visitor/semantic_analysis.cpp
+++ b/test/unit/visitor/semantic_analysis.cpp
@@ -122,3 +122,27 @@ SCENARIO("Ion variable in CONSTANT block", "[visitor][semantic_analysis]") {
         }
     }
 }
+
+SCENARIO("INDEPENDENT block", "[visitor][semantic_analysis]") {
+    GIVEN("A mod file with Independent block with only t") {
+        std::string nmodl_text = R"(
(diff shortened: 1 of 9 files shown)


test/unit/visitor/semantic_analysis.cpp
```diff
@@ -122,3 +122,27 @@ SCENARIO("Ion variable in CONSTANT block", "[visitor][semantic_analysis]") {
...
+    }
+    GIVEN("A mod file with Independent block with something else than t") {
+        std::string nmodl_text = R"(
+            INDEPENDENT {
+                t FROM 0 TO 1 WITH 100
+                u FROM 0 TO 1 WITH 100
+            }
+        )";
+        THEN("Semantic analysis fails") {
+            REQUIRE_FALSE(run_semantic_analysis_visitor(nmodl_text));
```
alkino, line 145: How can I check the output of the visitor?

pramodk: catch2 provides an option to catch stderr/stdout?

pramodk (approved):
LGTM

I took the liberty do a minor change and adding `todo` as a reminder.

PR #970, opened 2022-10-26
Use the legacy Crout solver : nrn / scopmath / crout.c

kotsaloscv:
Experimentation with different versions of the Crout solver

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -30,8 +30,7 @@ namespace codegen {
  *      for(int id=0; id<nodecount; id++) {
  *
  */
-void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType type,
-                                                                    bool error_checking) {
+void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType type) {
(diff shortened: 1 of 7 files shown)


kotsaloscv:
@pramodk 
FYI, I decided to replace the previous crout solver that I have introduced with the legacy one found in NEURON: https://github.com/neuronsimulator/nrn/blob/master/src/scopmath/crout.c

I took this decision in order to have a way more bullet-proof solver and be sure that our tests do not fail because of some implementation error.

kotsaloscv:
@olupton @pramodk this PR passes all the tests, i.e.:
1. https://bbpgitlab.epfl.ch/hpc/coreneuron/-/pipelines/81127 (correctness)
1. https://bbpgitlab.epfl.ch/hpc/cellular/coreneuron-perf-tests/-/pipelines/81133 (good performance)
Therefore, it is safe to be merged and replace the previous and merged Crout implementation (with the bug in the error handling).

I have opened this issue #971 to enhance the error checking. Currently, the error handling is through an assert. Not ideal but for now it is a sufficient solution.

olupton (approved):
LGTM (well, a little sad to re-incorporate old C code... 🙈)

PR #973, opened 2022-11-23
Update Catch 2 to a v3 version

alkino:
A lot of changes are needed because they break how catch2 from v2 to v3.
First of all, this is no more header-only but as a static library now.
Headers are split for compilation speed.


test/unit/parser/parser.cpp
```diff
@@ -34,8 +33,8 @@ bool is_valid_construct(const std::string& construct) {
 }
 
 
-SCENARIO("NMODL can accept \\r as return char for one line comment", "[parser]") {
-    GIVEN("A comment defined with \\r as return char") {
+SCENARIO("NMODL can accept CR as return char for one line comment", "[parser]") {
+    GIVEN("A comment defined with CR as return char") {
```
iomaganaris, line 37: Maybe some mistake with the replace?

alkino: No, \r is no more accepted in title. CR is the name of \r

.github/workflows/nmodl-ci.yml
```diff
@@ -143,6 +143,7 @@ jobs:
         env:
           CCACHE_BASEDIR: ${{runner.workspace}}/nmodl
           CCACHE_DIR: ${{runner.workspace}}/ccache
+          PYTHONMALLOC: malloc
```
alkino, line 146: Changes to be discussed

CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 # =============================================================================
 option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybind11 based python bindings" ON)
 option(NMODL_ENABLE_LEGACY_UNITS "Use original faraday, R, etc. instead of 2019 nist constants" OFF)
+option(NMODL_ENABLE_TESTS "Enable build of tests" ON)
```
alkino, line 25: Changes to be discussed

setup.py
```diff
@@ -85,6 +85,7 @@ def run(self, *args, **kwargs):
 cmake_args = ["-DPYTHON_EXECUTABLE=" + sys.executable]
 if "bdist_wheel" in sys.argv:
     cmake_args.append("-DLINK_AGAINST_PYTHON=FALSE")
+    cmake_args.append("-DNMODL_ENABLE_TESTS=FALSE")
```
alkino, line 88: To be discussed

.github/workflows/nmodl-ci.yml
olupton: Seems OK to me.

Context: Catch2 does not set the full environment (https://github.com/catchorg/Catch2/issues/1810) when running test executables to discover the list of tests.

When LeakSanitizer is enabled, it flags leaks from Python if we do not set `PYTHONMALLOC=malloc`.

CMakeLists.txt
olupton: Seems reasonable.

setup.py
olupton: Seems reasonable to me.
We already did not run the CTest suite when building the wheels.

test/unit/newton/newton.cpp
```diff
@@ -36,7 +35,7 @@ SCENARIO("Non-linear system to solve with Newton Numerical Diff Solver", "[numer
             CAPTURE(iter_newton);
             CAPTURE(X);
             REQUIRE(iter_newton > 0);
-            REQUIRE(X[0] == Approx(1.0));
+            REQUIRE(X[0] == Catch::Approx(1.0));
```
olupton, line 38: Note that they recommend not using this: https://github.com/catchorg/Catch2/blob/06c0e1cfab2a2fd6d01527b92422df2a8c49a8d7/docs/comparing-floating-point-numbers.md#approx

olupton (commented):
I didn't spot anything in the diff, but the failing CI needs to be understood...

olupton:
> I didn't spot anything in the diff, but the failing CI needs to be understood...

(Slack is down)
I believe the problem with the CI is that your Spack branch makes `nmodl@develop` depend on `catch2@3`, but this conflicts with `libsonata` requiring `catch2@2`, so `nmodl@develop` is not used (even though it is marked as preferred) and we do not get the changes from this PR.
cc: @iomaganaris @alkino

alkino:
Alléluïa
So what next?
Merging the spack changes?

test/unit/newton/newton.cpp
alkino: I changed this.

iomaganaris (approved):

PR #983, opened 2022-12-26
[LLVM][refactoring] Preparing code for builders

georgemitenkov:
Since #911 was a huge change and difficult to review, it is better to split into multiple PRs. This one reorganises code, fixes some typos ad renames functions. Some code has now explicit `if (GPU) else if (SIMD) else` blocks which is what we need to have multiple builders.


iomaganaris:
FYI the Azure CI are failing due to incompatibilities between the `llvm` branch of `NMODL` and the latest version of `NEURON` being used. I would only focus on the `GitLab CI` that runs the CPU and GPU integration tests of the JIT workflow

iomaganaris (approved):

PR #990, opened 2023-01-12
Fix setup.py

alkino:
The depends sphinxcontrib-applehelp is badly created with the last version name is `sphinxcontrib.applehelp` that leads to something I don't really understand.
Let's use the last version.


pramodk (approved):
LGTM - I guess we don’t need to spend much time of newer version is working as expected ?

PR #858, opened 2022-05-03
Remove OMP backend

alkino:

PR #994, opened 2023-01-26
Improve / fix codegen support for PROTECT and MUTEX constructs

pramodk:
* typo/invalid syntax for mutex block from #961 : name for critical section should be in ( )
* ~avoid printing `ivdep` as on CPU also we have to respect PROTECT and MUTEX statements. So now onwards, just use OpenMP semantics rather than non-standard pragma like ivdep.~
* Print `ivdep` only if MUTEX or PROTECT constructs doesn't appear in that specific block.  
* protect statement now becomes semantically same as atomic statements. This adds some limitations but that's what allows us to rely on OpenMP / OpenACC for atomic udpates.
* PROTECT can be now used on GPU as well
* Update test

Related to https://github.com/neuronsimulator/nrn/issues/2191

### DIFF
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -9,6 +9,7 @@
 
 #include "ast/eigen_linear_solver_block.hpp"
 #include "ast/integer.hpp"
+#include "ast/protect_statement.hpp"
 
 
 namespace nmodl {
@@ -353,6 +354,13 @@ void CodegenAccVisitor::print_net_send_buf_count_update_to_device() const {
 void CodegenAccVisitor::print_dt_update_to_device() const {
     printer->fmt_line("#pragma acc update device({}) if (nt->compute_gpu)",
                       get_variable_name(naming::NTHREAD_DT_VARIABLE));
+}
+
+void CodegenAccVisitor::visit_protect_statement(const ast::ProtectStatement& node) {
+    print_atomic_reduction_pragma();
+    printer->add_indent();
+    node.get_expression()->accept(*this);
+    printer->add_text(";");
 }
 
 }  // namespace codegen
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -55,11 +55,9 @@ void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType ty
 }
 
 
-void CodegenAccVisitor::print_atomic_reduction_pragma(bool skip) {
-    if (!skip) {
-        printer->add_line("nrn_pragma_acc(atomic update)");
-        printer->add_line("nrn_pragma_omp(atomic update)");
-    }
+void CodegenAccVisitor::print_atomic_reduction_pragma() {
+    printer->add_line("nrn_pragma_acc(atomic update)");
+    printer->add_line("nrn_pragma_omp(atomic update)");
 }
 
 
@@ -357,13 +355,6 @@ void CodegenAccVisitor::print_net_send_buf_count_update_to_device() const {
 void CodegenAccVisitor::print_dt_update_to_device() const {
     printer->fmt_line("#pragma acc update device({}) if (nt->compute_gpu)",
                       get_variable_name(naming::NTHREAD_DT_VARIABLE));
-}
-
-void CodegenAccVisitor::visit_protect_statement(const ast::ProtectStatement& node) {
-    print_atomic_reduction_pragma();
-    printer->add_indent();
-    node.get_expression()->accept(*this);
-    printer->add_text(";");
 }
 
 }  // namespace codegen
--- a/src/codegen/codegen_acc_visitor.cpp
+++ b/src/codegen/codegen_acc_visitor.cpp
@@ -31,7 +31,8 @@ namespace codegen {
  *      for(int id=0; id<nodecount; id++) {
  *
  */
-void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType type) {
+void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType type,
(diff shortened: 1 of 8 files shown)


src/codegen/codegen_cpp_visitor.cpp
```diff
@@ -319,16 +319,16 @@ void CodegenCVisitor::visit_update_dt(const ast::UpdateDt& node) {
 }
 
 void CodegenCVisitor::visit_protect_statement(const ast::ProtectStatement& node) {
-    printer->fmt_start_block("#pragma omp critical {}", info.mod_suffix);
+    printer->add_line("#pragma omp atomic update");
```
pramodk, line 322: note: we will need to update NMODL docs to mention constraints to use `PROTECT` statement (i.e. they need to be valid automic statements for OpenMP)

olupton: If `CodegenCVisitor` implemented `print_atomic_reduction_pragma` then we wouldn't need to implement `visit_protect_statement` in `CodegenAccVisitor`..right?

src/codegen/codegen_cpp_visitor.cpp
```diff
@@ -319,16 +319,16 @@ void CodegenCVisitor::visit_update_dt(const ast::UpdateDt& node) {
...
     printer->add_text(";");
-    printer->add_newline();
-    printer->end_block(1);
 }
 
 void CodegenCVisitor::visit_mutex_lock(const ast::MutexLock& node) {
-    printer->fmt_start_block("#pragma omp critical {}", info.mod_suffix);
+    printer->fmt_line("#pragma omp critical ({})", info.mod_suffix);
+    printer->add_indent();
+    printer->start_block();
```
alkino, line 331: where is the end_block?

iomaganaris (commented):
Didn't look in detail at the code yet but could you also add these keywords to one of the mod files in https://github.com/BlueBrain/nmodl/tree/master/test/integration/mod to make sure that they are compiled properly?

src/codegen/codegen_cpp_visitor.cpp
pramodk: yes..needed some extra changes done in 98eb8488d3dda85bf1e2fcff7e628c43de1221fd (basically, we don't need to emit `pragma omp atomic` at all places where `print_atomic_reduction_pragma()` is used today).

pramodk:
> https://github.com/BlueBrain/nmodl/tree/master/test/integration/mod to make sure that they are compiled properly?

Added.

test/unit/codegen/codegen_cpp_visitor.cpp
```diff
@@ -637,28 +639,39 @@ SCENARIO("Check codegen for MUTEX and PROTECT", "[codegen][mutex_protect]") {
...
-                #pragma omp critical TEST {
-                    inst->tmp[id] = 12.0;
-                })";
-            REQUIRE_THAT(generated, Contains(expected_code));
+                #pragma omp atomic update
+                inst->tmp[id] = 12.0;)";
+
+            // atomic update for the PROTECT construct
+            std::string expected_code_proc = R"(#pragma omp atomic update
+        inst->foo[id] = 21.0;)";
```
olupton, line 671: This isn't an atomic update. [The OpenMP spec](https://www.openmp.org/spec-html/5.0/openmpsu95.html) says:
```
...
where expression-stmt is an expression statement with one of the following forms:
...
* If atomic-clause is update or not present:
  x++; 
  x--; 
  ++x; 
  --x; 
  x binop= expr; 
  x = x binop expr; 
  x = expr binop x;
```
I'm not sure if this test code is actually compiled (I think not?), but we should generate something correct.

pramodk: > if this test code is actually compiled

yes, it's not compiled (we just compare the code structure) but yeah it was my sloppiness here!

src/codegen/codegen_acc_visitor.cpp
```diff
@@ -355,5 +359,12 @@ void CodegenAccVisitor::print_dt_update_to_device() const {
                       get_variable_name(naming::NTHREAD_DT_VARIABLE));
 }
 
+void CodegenAccVisitor::visit_protect_statement(const ast::ProtectStatement& node) {
```
olupton, line 362: Why does this need to be overriden in `CodegenAccVisitor`?

src/codegen/codegen_acc_visitor.cpp
```diff
@@ -53,9 +55,11 @@ void CodegenAccVisitor::print_channel_iteration_block_parallel_hint(BlockType ty
 }
 
 
-void CodegenAccVisitor::print_atomic_reduction_pragma() {
-    printer->add_line("nrn_pragma_acc(atomic update)");
-    printer->add_line("nrn_pragma_omp(atomic update)");
+void CodegenAccVisitor::print_atomic_reduction_pragma(bool skip) {
+    if (!skip) {
```
olupton, line 59: I am struggling to understand the point of the `skip` parameter. When is it better to call `print_atomic_reduction_pragma(true)` instead of just deleting the line?

src/codegen/codegen_acc_visitor.cpp
pramodk: 👍  - was a just leftover!

src/codegen/codegen_acc_visitor.cpp
pramodk: you are right! now it is simplified! 

(your comment forced me to look into existing code and realised that #858 has removed OpeMP backend for which empty `print_atomic_reduction_pragma()` calls were added. I got rid of those and now it should be simpler)

pramodk:
Only docs was failing in previous CI run and it should be fixed by 1048d5eaf6b36745e25bce33da4f8d009ad57386. So will press the button!

PR #996, opened 2023-02-01
Remove unnecessary code after OpenMP backend removal

pramodk:
* #858 has removed OpenMP backend with mechanism level async execution
* shadow_statements vector is not used and always empty now
* this will help to cleanup some logic/confusion in #994

```cpp
$ grep -r shadow_statement *

codegen/codegen_cpp_visitor.cpp:    for (const auto& statement: shadow_statements) {
codegen/codegen_cpp_visitor.cpp:    shadow_statements.clear();
codegen/codegen_cpp_visitor.cpp:    if (!shadow_statements.empty()) {
codegen/codegen_cpp_visitor.hpp:    std::vector<ShadowUseStatement> shadow_statements;
```

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -1180,17 +1180,6 @@ void CodegenCVisitor::print_atomic_reduction_pragma() {
 }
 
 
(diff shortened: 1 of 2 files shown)


alkino (approved):

pramodk:
added in #994

PR #997, opened 2023-02-01
More CodegenCVisitor coverage

### LINKED ISSUE #957: Increase code coverage of CodegenCVisitor
Currently the code coverage of the https://github.com/BlueBrain/nmodl/blob/master/src/codegen/codegen_cpp_visitor.cpp responsible for printing almost all of the C++ code is around ~89% according to the code coverage report you can find attached.
[nmodl_coverage.zip](https://github.com/BlueBrain/nmodl/files/9728180/nmodl_coverage.zip)
The code coverage was generated using the following script https://github.com/BlueBrain/nmodl/blob/magkanar/codecov/codecov.sh.
This issue is about trying to increase it as much as possible to 100% with the addition of unit or integration tests.

alkino:
Related to #957

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -2932,7 +2932,7 @@ static size_t get_register_type_for_ba_block(const ast::Block* block) {
 void CodegenCVisitor::print_mechanism_register() {
     printer->add_newline(2);
     printer->add_line("/** register channel with the simulator */");
-    printer->fmt_start_block("void _{}_reg() ", info.mod_file);
+    printer->fmt_start_block("void _{}_reg()", info.mod_file);
 
     // type related information
     auto suffix = add_escape_quote(info.mod_suffix);
@@ -3626,7 +3626,7 @@ void CodegenCVisitor::print_nrn_destructor() {
 void CodegenCVisitor::print_nrn_alloc() {
     printer->add_newline(2);
     auto method = method_name(naming::NRN_ALLOC_METHOD);
-    printer->fmt_start_block("static void {}(double* data, Datum* indexes, int type) ", method);
+    printer->fmt_start_block("static void {}(double* data, Datum* indexes, int type)", method);
     printer->add_line("// do nothing");
     printer->end_block(1);
 }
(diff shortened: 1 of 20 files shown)


test/unit/codegen/codegen_cpp_visitor.cpp
```diff
@@ -638,6 +638,79 @@ SCENARIO("Check code generation for FUNCTION_TABLE block", "[codegen][function_t
...
+            foo_(nt);
+            &tqitem;
+            pnodecount+id;)";
+            REQUIRE_THAT(generated, Contains(expected_code));
+        }
+    }
+}
+
+
 SCENARIO("Check codegen for MUTEX and PROTECT", "[codegen][mutex_protect]") {
```
pramodk, line 714: nice! 

@alkino : if you are tackling this, could you check what more could be easily added?

src/codegen/codegen_cpp_visitor.cpp
```diff
@@ -4121,7 +4119,7 @@ void CodegenCVisitor::print_net_receive_kernel() {
     }
 
     printer->add_newline(2);
-    printer->fmt_start_block("static inline void {}({}) ", name, get_parameter_str(params));
+    printer->fmt_start_block("static inline void {}({})", name, get_parameter_str(params));
```
iomaganaris, line 4122: Looking at the code coverage report I saw that the code corresponding to `info.artificial_cell` is not covered and it should be easy to do by taking as an example: https://github.com/neuronsimulator/nrn/blob/master/test/coreneuron/mod%20files/invlfire.mod

iomaganaris (commented):
Is it possible to get a report of the whole `codegen_cpp_visitor.cpp` to see which parts are not covered after the changes in this PR?

iomaganaris (approved):
LGTM

src/codegen/codegen_cpp_visitor.cpp
iomaganaris: This can be done in another PR. #957 can stay open

test/unit/codegen/codegen_cpp_visitor.cpp
iomaganaris: Rest of additions can be handled by separate PRs. We can follow #957 for this

PR #998, opened 2023-02-06
Fix code generation for OpenMP GPU backend with NVHPC

iomaganaris:
- Solves the issues described in https://forums.developer.nvidia.com/t/issue-with-locally-defined-classes-in-openmp-offload-region-since-nvhpc-22-5/228735/7 by defining the `functor` structs in the beginning of the file
- Each time there is an `EigenNewtonSolverBlock` there will be another `functor` struct defined. This is to make sure that the proper structs are always defined regardless of the order and where they are defined (`nrn_init` block, `nrn_state` or function)
- Added integration and unit tests
- In BB5 we have sympy 1.9 and in the other CIs there was 1.11. These two different versions generated different code and the new unit tests will fail in one or the other. To solve this issue I created a `requirements.txt` which is used to install all the python packages needed in all of the CIs. Currently I have pinned the sympy version the same as BB5. It can be updated once we update the version in BB5 as well

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -39,14 +39,14 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: ${{ env.PYTHON_VERSION }}
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 2    
       - name: Install Python3 dependencies
         working-directory: ${{runner.workspace}}/nmodl
         run: |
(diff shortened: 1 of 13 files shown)


iomaganaris:
CI is green and the PR is now ready for review :rocket: 
Once this is merged I can update the Spack PR (https://github.com/BlueBrain/spack/pull/1654) with a new NMODL version to deploy NVHPC 23.1

olupton (approved):
🚀

PR #1000, opened 2023-02-07
Remove shadow variables that were only use with OpenMP backend

alkino:


pramodk (commented):
LGTM (if CI will be green)

pramodk (approved):

PR #1006, opened 2023-03-02
BAType: use name instead of values

alkino:
CI_BRANCHES:NEURON_BRANCH=named_before_after,


iomaganaris:
Could we add a `BEFORE/AFTER` statement in one of the mod files of https://github.com/BlueBrain/nmodl/tree/master/test/integration/mod just to make sure that everything compiles properly?

iomaganaris (approved):

PR #1010, opened 2023-03-06
sympy@1.11.1 yields (-A-B) not -(A+B)

olupton:
This version is included in the 2023 redeployment on BB5, see: https://bbpgitlab.epfl.ch/hpc/nmodl/-/jobs/586400

https://github.com/BlueBrain/spack/pull/1866 was needed in order for the tests to execute.
https://github.com/BlueBrain/spack/pull/1867 was needed in order for nvhpc compilation to succeed.

### DIFF
--- a/packaging/build_requirements.txt
+++ b/packaging/build_requirements.txt
@@ -1,4 +1,4 @@
 Jinja2>=2.9.3
 PyYAML>=3.13
 pytest
-sympy>=1.3,<1.6
+sympy>=1.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,5 +2,5 @@ Jinja2>=2.9.3
 PyYAML>=3.13
 pytest
 pytest-cov
-sympy==1.9
+sympy
 numpy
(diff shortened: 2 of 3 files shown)


iomaganaris:
We can also remove the version requirement in https://github.com/BlueBrain/nmodl/blob/master/requirements.txt

olupton:
> We can also remove the version requirement in https://github.com/BlueBrain/nmodl/blob/master/requirements.txt

We could try, but it might just lead to more issues 🤷 This PR just gets the tests to pass for one more specific version, rather than doing anything substantive towards getting **all** versions to pass. So it might still be good to specify a version that we know works.

iomaganaris:
> > We can also remove the version requirement in https://github.com/BlueBrain/nmodl/blob/master/requirements.txt
> 
> We could try, but it might just lead to more issues shrug This PR just gets the tests to pass for one more specific version, rather than doing anything substantive towards getting **all** versions to pass. So it might still be good to specify a version that we know works.

I can check again but IIRC this was the only failing test with the newest `sympy` version

iomaganaris (approved):

Issue #1016, opened 2023-03-31
(One More) Difference with UNITS precision in code generation 


pramodk:
Consider a UNITS block like:

```console
NEURON {
	SUFFIX cagk
}

UNITS {
	R = 8.313424 (joule/degC)
}
```

* NEURON generates

```cpp
  static double R = 8.313424;
```

* MOD2C generates

```cpp
 static double R = 8.313424;
```

* NMODL generates

```cpp
  static const double R = 8.31342;
```

alkino:
See https://github.com/neuronsimulator/nrn/pull/2315
https://github.com/BlueBrain/mod2c/pull/94
https://github.com/BlueBrain/nmodl/pull/1018

iomaganaris:
With the changes in the above PRs we get consistent units in all cases.
To validate this I used the following mod file: https://github.com/neuronsimulator/nrn/blob/377a1383bb46517c6f10b8fad5200db1cf0fc4d8/test/pynrn/unitstest.mod which was created to test the `UNITS`.
The `UNITS` of this file get compiled by `NEURON` to:
```
#define mol _nrnunit_mol[_nrnunit_use_legacy_]
static double _nrnunit_mol[2] = {0x1.fe185ca57c517p+78, 6.02217e+23};
 
#define E _nrnunit_E[_nrnunit_use_legacy_]
static double _nrnunit_E[2] = {0x1.7a4da290c1653p-63, 1.60219e-19};
 
#define FARADAY _nrnunit_FARADAY[_nrnunit_use_legacy_]
static double _nrnunit_FARADAY[2] = {0x1.78e555060882cp+16, 96485.3};
 
#define h _nrnunit_h[_nrnunit_use_legacy_]
static double _nrnunit_h[2] = {0x1.b860bde023111p-111, 6.626e-34};
 
#define hb _nrnunit_hb[_nrnunit_use_legacy_]
static double _nrnunit_hb[2] = {0x1.185a7057c690dp-113, 1.055e-34};
 
#define R _nrnunit_R[_nrnunit_use_legacy_]
static double _nrnunit_R[2] = {0x1.0a1013e8990bep+3, 8.3145};
 static double Rexact = 8.313424;
 
#define boltzmann _nrnunit_boltzmann[_nrnunit_use_legacy_]
static double _nrnunit_boltzmann[2] = {0x1.0b0e6d55e647cp-76, 1.38065e-23};
 
[...]

PR #1017, opened 2023-03-31
Bug fix in code generation in net_send() calll in INITIAL block under NET_RECEIVE

pramodk:
- when we emit a call to `net_send_buffering()`, the `weight_index` is either `0` or an argument `weight_index` itself depending on whether the code printed is for `NET_RECEIVE` block or a top-level `INITIAL` block.

```python
INITIAL {
    net_send(x, y)
}
```

vs

```python
NET_RECEIVE {
      INITIAL {
          net_send(x, y)
      }
}
```

- BUT, `NET_RECEIVE` block can contain `INITIAL` block itself (where we emit `net_init()` function).
- So, before using `weight_index=0` in `net_send_buffering()`, check if we are printing `INITIAL` block of `NET_RECEIVE` or top level `INITIAL` block.

NOTES:

* This issue was revealed by a hippocampus test of BBP BlueConfig where the delayed connection feature in `ProbAMPANMDA_EMS.mod` rely on this pattern: https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/pipelines/111363.
* The hippocampus test now passes for spike-level comparison. But note that there is a separate issue with UNITS precision: #1016.
* **Fixing the above issue will generate the exact same results for the hippocampus model!**
* Create a separate issue in NEURON if a manful test could be added to cover the

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -3771,8 +3771,9 @@ void CodegenCVisitor::print_net_send_call(const FunctionCall& node) {
     std::string weight_index = "weight_index";
     std::string pnt = "pnt";
 
-    // for non-net_receieve functions i.e. initial block, the weight_index argument is 0.
-    if (!printing_net_receive) {
+    // for functions not generated from NET_RECEIVE blocks (i.e. top level INITIAL block)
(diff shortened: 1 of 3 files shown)


ohm314 (approved):
lgtm

alkino (approved):

PR #1022, opened 2023-04-06
Fix parsing of double from the unit file

alkino:


iomaganaris (approved):

test/unit/units/parser.cpp
```diff
@@ -125,6 +125,11 @@ SCENARIO("Unit parser accepting valid units definition", "[unit][parser]") {
                 REQUIRE(is_valid_construct("dipotre\t\t\t/m\n"));
             }
         }
+        WHEN("A 's' is added") {
+            THEN("parser remove it to find the units") {
+                REQUIRE_NOTHROW(parse_string("pew 1 m\nfoo 2 pews\n"));
```
iomaganaris, line 130: I'm curious about this. Is there such case in the `nrnunits.lib`?

alkino: no, should I remove test and code related to that?

PR #1027, opened 2023-04-15
Use c++17 filesystem

alkino:


pramodk (commented):
LGTM - could you check what is causing the segfault in the failing CI?

alkino:
> LGTM - could you check what is causing the segfault in the failing CI?

Impossible to reproduce locally... @pramodk

src/units/units.cpp
```diff
@@ -314,10 +310,9 @@ void UnitTable::print_units_sorted(std::ostream& units_details) const {
                                                                                table.end());
     std::sort(sorted_elements.begin(), sorted_elements.end());
     for (const auto& it: sorted_elements) {
-        units_details << fmt::format("{} {:.{}f}: {}\n",
```
iomaganaris, line 317: Instead of being based to a default print scheme of fmt I think it would be better to use `{:g}` instead to be consistent with other places we print float numbers as well

PR #1028, opened 2023-04-17
Slowly try to remove things

alkino:


pramodk:
@alkino : I didn't get time to respond on #1027  - did you try to use Valgrind or address sanitiser locally? (if there is a really a bug, I wonder if valgrind/asan will catch it).

alkino:
The bug is the usage of filesystem in gcc-8

PR #1031, opened 2023-04-18
Fix euler method resolution by computing in two times 

alkino:
First time, compute the derivative part.
Second time, compute the value.
This way if there is inter-depends it will still be correct.

```
BREAKPOINT {
  SOLVE states METHOD euler
}

DERIVATIVE states {
   a' = a/n_tau
   n' = (n - a)/n_tau
}
```


src/parser/diffeq_context.cpp
```diff
@@ -140,7 +140,7 @@ std::string DiffEqContext::get_cnexp_solution() const {
  * Return solution for euler method
  */
 std::string DiffEqContext::get_euler_solution() const {
```
pramodk, line 142: `nitpick`: still this method name `get_euler_solution` make sense here? the Euler solution now is mostly in `src/visitors/neuron_solve_visitor.cpp` at line `94-96` ?

pramodk (commented):
Just a suggestion added. 

Also, an example from the PR description could be added in `test/unit/codegen/codegen_cpp_visitor.cpp` ?

src/visitors/neuron_solve_visitor.hpp
```diff
@@ -59,6 +59,8 @@ class NeuronSolveVisitor: public AstVisitor {
     /// the derivative name currently being visited
     std::string derivative_block_name;
 
+    std::vector<std::shared_ptr<ast::Statement>> euler_values_computation;
```
iomaganaris, line 62: Just a nitpick.. This variable could be renamed to `euler_solution_expressions`?

iomaganaris (commented):
LGTM overall
I launched the blueconfigs CI in https://bbpgitlab.epfl.ch/hpc/sim/blueconfigs/-/pipelines/116708

test/unit/codegen/codegen_cpp_visitor.cpp
```diff
@@ -1083,6 +1083,41 @@ SCENARIO("Some tests on derivimplicit", "[codegen][derivimplicit_solver]") {
...
+               m' = 2 * m
+               inf = inf * 3
+               n' = (2 + m - inf) * n
+            }
+        )";
+        THEN("Correct code is generated") {
+            auto const generated = get_cpp_code(nmodl_text);
+            std::string nrn_state_expected_code = R"(inst->Dm[id] = 2.0 * inst->m[id];
+            inf = inf * 3.0;
+            inst->Dn[id] = (2.0 + inst->m[id] - inf) * inst->n[id];
```
iomaganaris, line 1112: Just to note here that with this change we start using the `Dvar` variables which we didn't before. This will introduce an extra unnecessary write and the `Dvar`s are part of the InstanceStruct and increase the memory usage of the mechanism.
We could improve this by avoiding the use of these variables, stop adding them to the InstanceStruct and just using local variables.
For the moment we can keep it like this for compatibility with NEURON.
cc: @pramodk

iomaganaris (approved):

test/unit/codegen/codegen_cpp_visitor.cpp
pramodk: Yes, perfectly fine with how you formulated!

test/unit/utils/nmodl_constructs.cpp
```diff
@@ -1525,42 +1525,42 @@ std::vector<DiffEqTestCase> const diff_eq_constructs{
...
-                "Rho_GB = Rho_GB+dt*(( - Rho_GB*(1-Rho_GB)*(rho_star_GB-Rho_GB) + potentiate_GB*gamma_p_GB*(1-Rho_GB) - depress_GB*gamma_d_GB*Rho_GB ) / ((1e3)*tau_GB))",
+                "DRho_GB = ( - Rho_GB*(1-Rho_GB)*(rho_star_GB-Rho_GB) + potentiate_GB*gamma_p_GB*(1-Rho_GB) - depress_GB*gamma_d_GB*Rho_GB ) / ((1e3)*tau_GB)",
                 "euler"
         },
 
         {
                 "GluSynapse.mod",
                 "Use_GB' = (Use_d_GB + Rho_GB*(Use_p_GB-Use_d_GB) - Use_GB) / ((1e3)*tau_Use_GB)",
-                "Use_GB = Use_GB+dt*((Use_d_GB + Rho_GB*(Use_p_GB-Use_d_GB) - Use_GB) / ((1e3)*tau_Use_GB))",
+                "DUse_GB = (Use_d_GB + Rho_GB*(Use_p_GB-Use_d_GB) - Use_GB) / ((1e3)*tau_Use_GB)",
```
pramodk, line 1563: @alkino : correct me if I am wrong: I don't see the `euler solution` that we create like `create_statement(fmt::format("{} = {} + dt * D{}", n, n, n));` . Possible to test that part as well? 🤔

alkino: it is tested in codegen, because this line is in codegen

PR #1009, opened 2023-03-06
Various codegen fixes

iomaganaris:
Fixes compilation of mod files of `dbbs-mod-collection` (after small changes in mod files in https://github.com/iomaganaris/dbbs-mod-collection/tree/magkanar/nmodl)

- Move definition of functors after declaration of functions
- Fix issue with parsing `()` from `sympy` if `ode_rhs` is empty (Fixes #1008)

PR #1034, opened 2023-04-20
Print prototypes before functors

alkino:


pramodk (approved):

iomaganaris (approved):

iomaganaris:
I've also done an alternate fix for that in #1009 together with some other stuff. I have no problem merging either of them

iomaganaris (approved):

PR #1036, opened 2023-04-25
Remove Travis from build_wheels

alkino:


setup.py
```diff
@@ -75,29 +75,6 @@ def run(self, *args, **kwargs):
...
-        os.environ["NMODL_WRAPLIB"] = os.path.join(
-            NMODL_PREFIX_DATA, "libpywrapper.dylib"
-        )
-    else:
-        os.environ["NMODL_WRAPLIB"] = os.path.join(NMODL_PREFIX_DATA, "libpywrapper.so")
-
-    # find libpython*.so in the system
-    os.environ["NMODL_PYLIB"] = find_libpython()
-
-    return os.path.join(NMODL_PREFIX_DATA, exe_name)
```
pramodk, line 98: @alkino : are you about this? i.e. this is tested? (travis part is obvious but wondering about this!)

alkino: oh, this function is not used (or this is really black magic).

alkino: We can found a copy there: https://github.com/BlueBrain/nmodl/pull/1036 (and this one is used).

pramodk: Ok, I thought this must be used under binary shim but I see a copy here: https://github.com/BlueBrain/nmodl/blob/07c7267998e2066f744885e43b88a595aa5f5591/pywheel/shim/_binwrapper.py#L14

pramodk (approved):
I don't see why `_config_exe` was required in `setup.py` cc: @ohm314. Otherwise LGTM

setup.py
alkino: Sorry, my link was broken

packaging/build_wheels.bash
```diff
@@ -6,10 +6,10 @@ set -xe
...
 #
 # PREREQUESITES:
-#  - cmake (>=3.3)
-#  - flex
-#  - bison
-#  - python >= 3.6
+#  - cmake (>=3.15)
+#  - flex (>= 2.6)
+#  - bison (>=3.0)
+#  - python (>=3.7)
```
alexsavulescu, line 12: 3.8 to align with NEURON

alexsavulescu (approved):

PR #1041, opened 2023-05-02
Use NMODLHOME for NMODL_WRAPLIB

alkino:
Use NMODLHOME for wraplib and modlunit

Also fix that `libpywrapper` is written in `CMAKE_BINARY_PREFIX` but read by tests in `NMODL_PROJECT_BINARY_PREFIX`.

That is not the case for `nrnunits.lib` that is written in `NMODL_PROJECT_BINARY_PREFIX`.

The problem was that when installed through nrn cmake, both files appears in different path.


pramodk (approved):
LGTM

PR #1055, opened 2023-08-31
Rename CodegenCVisitor to CodegenCppVisitor

ohm314:
The file was renamed but not the class, even though that probably should have been part of that refactoring

(just getting my feet wet again :) )


tristan0x (approved):
LGTM
I cannot find an occurrence of `CodegenCVisitor` anymore.

PR #1056, opened 2023-08-31
Generate files as read-write.

1uc:
Prior to this commit generated files were created read-only. When updating files, one would add write permissions and then overwrite the file. Even if it was read-only. This commit changes the behaviour to generate files with regular (read & write) permissions and doesn't change the permissions of files, e.g. it wont overwrite read-only files.

### DIFF
--- a/src/language/code_generator.py
+++ b/src/language/code_generator.py
@@ -303,7 +303,7 @@ class JinjaTask(
         raise Exception("Unexpected output file extension: " + suffix)
 
     def render(self):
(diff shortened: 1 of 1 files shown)


1uc:
To check that the files are/should only generated in a build folder:
https://github.com/search?q=repo%3ABlueBrain%2Fnmodl%20code_generator.py&type=code

pramodk (approved):
LGTM!

Ioannis and Nico also say we don't need to worry much about build folder files.

PR #1057, opened 2023-08-31
Fix file copying for Python helper files.

1uc:


src/pybind/CMakeLists.txt
```diff
@@ -19,11 +19,6 @@ endif()
 # build nmodl python module under lib/nmodl
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/nmodl)
 
-foreach(file ast.py dsl.py ode.py symtab.py visitor.py __init__.py)
-  list(APPEND NMODL_PYTHON_FILES_IN ${NMODL_PROJECT_SOURCE_DIR}/nmodl/${file})
-  list(APPEND NMODL_PYTHON_FILES_OUT ${PROJECT_BINARY_DIR}/lib/nmodl/${file})
-endforeach()
```
1uc, line 25: These lines are redundant with copying the `*_HELPER_FILES`.

src/pybind/CMakeLists.txt
```diff
@@ -68,27 +63,20 @@ if(NMODL_ENABLE_PYTHON_BINDINGS)
   if(LINK_AGAINST_PYTHON)
     target_link_libraries(_nmodl PRIVATE pywrapper)
   endif()
-
-  add_custom_command(
-    OUTPUT ${PROJECT_BINARY_DIR}/lib/nmodl
-    COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:_nmodl> ${PROJECT_BINARY_DIR}/nmodl
-    DEPENDS ${NMODL_PYTHON_FILES_IN} _nmodl
-    COMMENT "-- COPYING NMODL PYTHON MODULE --")
```
1uc, line 76: AFAICT the file `$<TARGET_FILE:_nmodl>` (a `_nmodl*.so`) automatically ends up at that location. This also happens when using this branch in neuron.

Moreover, it not obvious why there's a target that has a dependency on required file/folder to trigger this "custom command" to be run.

src/pybind/CMakeLists.txt
```diff
@@ -68,27 +63,20 @@ if(NMODL_ENABLE_PYTHON_BINDINGS)
...
-
 # =============================================================================
 # Copy python binding components and examples into build directory
 # =============================================================================
-file(GLOB NMODL_PYTHON_HELPER_FILES "${NMODL_PROJECT_SOURCE_DIR}/nmodl/*.py")
-file(COPY ${NMODL_PYTHON_HELPER_FILES} DESTINATION ${CMAKE_BINARY_DIR}/lib/nmodl/)
+file(
+  GLOB NMODL_PYTHON_FILES
+  RELATIVE "${NMODL_PROJECT_SOURCE_DIR}/nmodl/"
+  CONFIGURE_DEPENDS "${NMODL_PROJECT_SOURCE_DIR}/nmodl/*.py")
```
1uc, line 74: Following the previous use of `file(GLOB` in NMODL.

We can discuss "hard-coding" the list of files. There's a similar `GLOB` for creating a test for each mod file.

src/pybind/CMakeLists.txt
pramodk: > Moreover, it not obvious why there's a target that has a dependency on required file/folder to trigger this "custom command" to be run.

I didn't check the details but it seems to be introduced in b2c7443dc25dcfb69304ab1561dec534a1c1d4c6 with a very relevant (?) comment: 

```
         - Fixed issue in build system that prevented pytest from working
           correctly. This was a regression introduced by the recent build-
           system changes.
```

@tristan0x - you might have a better memory about this?

> AFAICT the file $<TARGET_FILE:_nmodl> (a _nmodl*.so) automatically ends up at that location

I see the same.

pramodk (commented):
Just left one question for @tristan0x. Otherwise LGTM.

src/pybind/CMakeLists.txt
1uc: I've checked:
```
pip install -r requirements.txt
pip install scikit-build

pip install -e .
pytest
```
which runs without issues.

pramodk: Ok. My guess was that the shared library copying was triggered if any of the inputs are changed. But as you pointed out, if the library is created at the same location, not sure what was need.

Anyway, I will merge this. If we see some fishy behaviour, we know where to look at :)   (@alkino , @iomaganaris)

PR #1059, opened 2023-09-01
Add `scikit-build` to `requirements.txt`.

1uc:


requirements.txt
```diff
@@ -5,3 +5,4 @@ pytest-cov
 sympy
 numpy
 find_libpython
+scikit-build
```
pramodk, line 8: looks good - some scripts in the repo explicitly install skbuild and then requirements.txt. We can remove the first one.

1uc: done

pramodk (approved):

PR #1060, opened 2023-09-04
Fix reference to local variable.

1uc:
During CI we see the following compiler warning:
```
/root/nrn/external/nmodl/ext/pybind11/include/pybind11/cast.h:1054:35: note: declared here
 1054 |     return cast_op<T>(load_type<T>(handle));
      |                       ~~~~~~~~~~~~^~~~~~~~
```
e.g. https://app.circleci.com/pipelines/github/neuronsimulator/nrn/5301/workflows/c43f0a31-36ec-4e95-93df-7df650db01d4/jobs/2887?invite=true#step-102-127220_85

If we track the definition of `cast_op` and `load_type` we find:
```
template <typename T>
typename make_caster<T>::template cast_op_type<T> cast_op(make_caster<T> &caster) {
    return caster.operator typename make_caster<T>::template cast_op_type<T>();
}
```
and
```
template <typename T>
make_caster<T> load_type(const handle &handle) {
    make_caster<T> conv;
    load_type(conv, handle);
    return conv;
}
```
The suspicion is that `make_caster<T>` will strip the `&` and store a copy
```
template <typename type>
using make_caster = type_caster<intrinsic_t<type>>;
```
However, then the returned reference does point to a local variable.

We fix this by returning `std::shared_ptr` by value.

### DIFF
--- a/src/language/nodes.py
+++ b/src/language/nodes.py
@@ -235,10 +235,14 @@ class ChildNode(BaseNode):
         return type_name
 
     @property
+    def _is_member_type_wrapped_as_shared_pointer(self):
+        return not (self.is_vector or self.is_base_type_node or self.is_ptr_excluded_node)
+
+    @property
     def member_rvalue_typename(self):
         """returns rvalue reference type when used as returned or parameter type"""
         typename = self.member_typename
-        if not self.is_integral_type_node:
+        if not self.is_integral_type_node and not self._is_member_type_wrapped_as_shared_pointer:
             return "const " + typename + "&"
         return typename
 
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -4060,7 +4060,7 @@ void CodegenCppVisitor::visit_for_netcon(const ast::ForNetcon& node) {
     // to the next netcon.
     const auto& args = node.get_parameters();
     RenameVisitor v;
-    auto& statement_block = node.get_statement_block();
+    const auto& statement_block = node.get_statement_block();
     for (size_t i_arg = 0; i_arg < args.size(); ++i_arg) {
         // sanitize node_name since we want to substitute names like (*w) as they are
         auto old_name =
--- a/src/language/templates/ast/ast.cpp
+++ b/src/language/templates/ast/ast.cpp
@@ -30,7 +30,7 @@ std::string Ast::get_node_name() const {
   throw std::logic_error("get_node_name() not implemented");
 }
 
-const std::shared_ptr<StatementBlock>& Ast::get_statement_block() const {
+std::shared_ptr<StatementBlock> Ast::get_statement_block() const {
   throw std::runtime_error("get_statement_block not implemented");
 }
 
--- a/src/language/templates/ast/ast.hpp
+++ b/src/language/templates/ast/ast.hpp
@@ -273,7 +273,7 @@ struct Ast: public std::enable_shared_from_this<Ast> {
    *
    * \sa ast::StatementBlock
    */
-  virtual const std::shared_ptr<StatementBlock>& get_statement_block() const;
+  virtual std::shared_ptr<StatementBlock> get_statement_block() const;
 
   /**
    * \brief Set symbol table for the AST node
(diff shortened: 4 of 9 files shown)


1uc:
These functions is covered by tests:
```
  void SympyReplaceSolutionsVisitor::visit_diff_eq_expression(ast::DiffEqExpression& node)
  void SympyReplaceSolutionsVisitor::visit_non_lin_equation(ast::NonLinEquation& node)
```

1uc:
It's slightly more complicated. Pybind11 supports returning `const` references. I believe those usually take the first branch in:
```
#define PYBIND11_OVERRIDE_IMPL(ret_type, cname, name, ...)                                        \
    do {                                                                                          \
        pybind11::gil_scoped_acquire gil;                                                         \
        pybind11::function override                                                               \
            = pybind11::get_override(static_cast<const cname *>(this), name);                     \
        if (override) {                                                                           \
            auto o = override(__VA_ARGS__);                                                       \
            if (pybind11::detail::cast_is_temporary_value_reference<ret_type>::value) {           \
                static pybind11::detail::override_caster_t<ret_type> caster;                      \
                return pybind11::detail::cast_ref<ret_type>(std::move(o), caster);                \
            }                                                                                     \
            return pybind11::detail::cast_safe<ret_type>(std::move(o));                           \
        }                                                                                         \
    } while (false)
```
https://github.com/pybind/pybind11/blob/master/include/pybind11/pybind11.h#L2829

We can check that:
```
pybind11::detail::cast_is_temporary_value_reference<const std::string&>::value == true
```
but
```
pybind11::detail::cast_is_temporary_value_reference<const std::shared_ptr<T>&>::value == false
```
which explains why the warning point to code that's called via:
```
            return pybind11::detail::cast_safe<ret_type>(std::move(o));                           \
```
rather than
```
                return pybind11::detail::cast_ref<ret_type>(std::move(o), caster);                \
```
(I've tried adding `if constexpr` to make sure the other branch isn't considered.)

Given the documented limitations with returning const-references, see
https://pybind11.readthedocs.io/en/stable/advanced/classes.html#overriding-virtual-functions-in-python
(see Notes at the end of that paragraph), I'd still propose to avoid the issue by returning by value.

pramodk (approved):
LGTM. 

Just for clarity: `PyAst::get_statement_block()` is not a culprit here?

1uc:
Sorry, somehow only a "fixup" commit made it to github (bad rebase or something). The first commit is the important one (that was missing).

src/language/nodes.py
```diff
@@ -234,11 +234,15 @@ def member_typename(self):
...
+    @property
+    def _is_member_type_wrapped_as_shared_pointer(self):
+        return not (self.is_vector or self.is_base_type_node or self.is_ptr_excluded_node)
+
     @property
     def member_rvalue_typename(self):
         """returns rvalue reference type when used as returned or parameter type"""
         typename = self.member_typename
-        if not self.is_integral_type_node:
+        if not self.is_integral_type_node and not self._is_member_type_wrapped_as_shared_pointer:
```
pramodk, line 245: Just a nitpick: the check for « is_integral_type_node«  is not needed because we check for « is_base_type_node«  and that includes all integral types node?

1uc: I'm not sure I understand and there's too many `not`s for my brain. The logic is: We don't add `const&` to integral types and anything wrapped in a shared pointer. The first condition makes the first half of the sentence clear; and the second condition the second half. (For that reason alone I wouldn't golf this.)

Can it be golfed? I don't think so because an `int` is not wrapped in a shared pointer and hence  `_is_member_type_wrapped_as_shared_pointer` should be `False`.

Let's go slowly for the example of an `int` and check if it will be wrapped if the condition is only:
```
if not self._is_member_type_wrapped_as_shared_pointer:
   return "const " + typename + "&"
```
Let's start:
```
is_base_type_node == True
=> _is_member_type_wrapped_as_shared_pointer == False
=> not _is_member_type_wrapped_as_shared_pointer == True
=> add `const &` to the return type.
```

pramodk: my bad! missed the second `not` in `_is_member_type_wrapped_as_shared_pointer`!

pramodk (approved):

PR #1062, opened 2023-09-05
Revert order of CPU vectorization pragmas

iomaganaris:
Newest Intel oneAPI compiler doesn't support `#pragma ivdep` and in the following line another `#pragma`.
We get errors like the following like that:
```
error: expected a for, while, or do-while loop to follow '#pragma ivdep'
     521                #pragma omp simd
     522                ^
```
Instead revert the order of the pragmas


pramodk (approved):
@iomaganaris : did you verify if `#pragma omp simd` is respected then? (e.g. add some unknown pragma like instead of ivdep and see if compiler properly handled omp simd? Or, it also expects an immediate for loop)

PR #1067, opened 2023-09-06
Avoid duplicate calculation of number of float variables

iomaganaris:
This PR addresses 2 things:
- Count properly the number of float variables based on `codegen_float_variables` vector size
- Refactor the counting of the `int` variables using `STL`

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -535,54 +535,6 @@ std::string CodegenCppVisitor::breakpoint_current(std::string current) const {
 }
 
 
-int CodegenCppVisitor::float_variables_size() const {
-    auto count_length = [](int l, const SymbolType& variable) {
-        return l += variable->get_length();
-    };
-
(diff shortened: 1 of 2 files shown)


alkino (approved):
Nice!

pramodk:
I would be interested to look at this, especially because CI fails with:

```console
Error: ForNetConTest is a different MOD file than used by NEURON!
Error : NEURON and CoreNEURON must use same mod files for compatibility, 1 different mod file(s) found. Re-compile special and special-core!
Fatal Python error: Aborted
```

iomaganaris:
If the CI passes this should now be ready to merge

1uc:
You could implement one in terms of the other, then the symmetry between int and float variable remains; but the code duplication is gone.

iomaganaris:
@1uc shall I press the button? Wanna take a final look?

iomaganaris:
@1uc ready to merge? Do you want to give a final look?

1uc (approved):

PR #1069, opened 2023-09-07
Simplify flat index computation.

1uc:
I feel it's easier to not rely on the knowing ordering of `jacobian` to obtain a first flat index (row major), which then needs to be broken into `i` and `j` to reassemble the other flat index (colum major). Instead we express the same logic with a double loop over the two matrix indices; and then compute the required (colum major) flat index.

### DIFF
--- a/nmodl/ode.py
+++ b/nmodl/ode.py
@@ -7,6 +7,7 @@
 
 from importlib import import_module
 
+import itertools
 import sympy as sp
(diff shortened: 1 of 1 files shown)


pramodk (approved):

1uc:
The TODO comment got migrated to this ticket: #1072

PR #1071, opened 2023-09-08
make nmodl::stringutils methods more consistent

### LINKED ISSUE #1066: Bring more consistency to nmodl::stringutils utility functions
## Status
Here are some of the utility functions in `src/utils/string_utils.hpp`:

```c++
std::string& ltrim(std::string& s)
std::string& rtrim(std::string& s)
std::string& trim(std::string& s)
void remove_character(std::string& str, const char c)
std::string& trim_newline(std::string& s)
std::string escape_quotes(const std::string& before)
std::string align_text(const std::string& text, int width, text_alignment type)
std::string tolower(std::string text)
```

We can distinguish 2 categories:
1. the functions doing a side effect by accepting a reference to a `std::string`. Most of these functions also return the reference, probably to use them with a functional style (`std::cout << stringutils::trim(my_str)`). Note that the function `remove_character` does not follow that idiom by returning `void`.
1. The others returning a new instance of `std::string` with no side effect. Besides, `tolower` should accept a r-value, not an l-value :)

## Situation

The functions of the first category are a bit misleading, because the fact that they return a `std::string` may suggest that the parameter is not modified, when it actually is.

## Target

2 solutions:
1. either functions accept an l-value, do in-place edition of the string, and return nothing.
2. or adopt a full functional style by returning a copy of the input string with no side-effect

I prefer to sacrifice the functional style of (2) in the altar of performance by doing in-place edition of the strings with solution (1)

What do you think @ohm314 @pramodk ?

tristan0x:
* The prototypes of these functions now follow these 2 principles:
  1. the input string is kept intact
  2. the transformed text is returned in a new `std::string` instance
* Improve the Doxygen documentation

Also removed some useless work along the way while updating the calls to these functions.

Fixes #1066

### DIFF
--- a/src/utils/string_utils.hpp
+++ b/src/utils/string_utils.hpp
@@ -60,8 +60,8 @@ enum class text_alignment { left, right, center };
  * \param text the string to manipulate
  * \return a copy of the given string with both leading and trailing ASCII space characters removed
  */
-[[nodiscard]] static inline std::string trim(const std::string& text) {
-    return ltrim(rtrim(text));
+[[nodiscard]] static inline std::string trim(std::string text) {
+    return ltrim(rtrim(std::move(text)));
 }
 
 /**
@@ -80,8 +80,8 @@ enum class text_alignment { left, right, center };
  * \param text the string to manipulate
  * \return a copy of the given string with all occurrences of the ASCII newline character removed
  */
-[[nodiscard]] static inline std::string trim_newline(const std::string& text) {
-    return remove_character(text, '\n');
+[[nodiscard]] static inline std::string trim_newline(std::string text) {
+    return remove_character(std::move(text), '\n');
 }
 
 /**
--- a/src/utils/string_utils.hpp
+++ b/src/utils/string_utils.hpp
@@ -27,62 +27,95 @@ namespace nmodl {
 namespace stringutils {
 
 /**
- * @addtogroup utils
- * @{
+ * \addtogroup utils
+ * \{
  */
 
 /// text alignment when printing in the tabular form
 enum class text_alignment { left, right, center };
 
-/// Trim from start
-static inline std::string& ltrim(std::string& s) {
-    s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](int c) { return !std::isspace(c); }));
-    return s;
+/**
+ * \param text the string to manipulate
+ * \return a copy of the given string with leading ASCII space characters removed
+ */
+[[nodiscard]] static inline std::string ltrim(std::string text) {
+    text.erase(text.begin(),
+               std::find_if(text.begin(), text.end(), [](int c) { return !std::isspace(c); }));
+    return text;
 }
 
-/// Trim from end
-static inline std::string& rtrim(std::string& s) {
-    s.erase(std::find_if(s.rbegin(), s.rend(), [](int c) { return !std::isspace(c); }).base(),
-            s.end());
-    return s;
+/**
+ * \param text the string to manipulate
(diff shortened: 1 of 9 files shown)


src/utils/string_utils.hpp
```diff
@@ -27,62 +27,95 @@ namespace nmodl {
...
-/// Remove leading newline for the string read by grammar
-static inline std::string& trim_newline(std::string& s) {
-    remove_character(s, '\n');
-    return s;
+/**
+ *
+ * \param text the string to manipulate
+ * \return a copy of the given string with all occurrences of the ASCII newline character removed
+ */
+[[nodiscard]] static inline std::string trim_newline(const std::string& text) {
```
1uc, line 83: Most functions take a `std::string` this one takes a const reference, why?

src/visitors/rename_visitor.hpp
```diff
@@ -70,19 +70,19 @@ class RenameVisitor: public ConstAstVisitor {
...
-                  std::string old_name,
+                  const std::string& old_name,
                   std::string new_var_name_or_prefix,
                   bool add_prefix,
                   bool add_random_suffix)
         : ast(std::move(ast))
-        , var_name_regex(std::move(old_name))
-        , add_prefix(std::move(add_prefix))
-        , add_random_suffix(std::move(add_random_suffix)) {
+        , var_name_regex(old_name)
```
1uc, line 83: All the `var_name_regex` related changes are correct, because it doesn't have an overload for `std::string &&`.

src/utils/string_utils.hpp
tristan0x: This function is just a pass-through function `remove_character`, no copy required.

1uc: But `remove_character` accepts an `std::string`. If you pass it a const-ref to an `std::string` this will force a copy. If `trim_newline` accepts by value and moves, then you always pay for one move, sometimes a copy, but the copy can be avoided if you move into `trim_newline`.

tristan0x: To make sure I understand, your point is that a copy will be made anyhow. So let's keep the API consistent with functions accepting a `std::string`, and make the "efficient" move an implementation detail. Something like below @1uc :

```diff
diff --git a/src/utils/string_utils.hpp b/src/utils/string_utils.hpp
index 2a48fc24..c11ab18b 100644
--- a/src/utils/string_utils.hpp
+++ b/src/utils/string_utils.hpp
@@ -60,8 +60,8 @@ enum class text_alignment { left, right, center };
  * \param text the string to manipulate
  * \return a copy of the given string with both leading and trailing ASCII space characters removed
  */
-[[nodiscard]] static inline std::string trim(const std::string& text) {
-    return ltrim(rtrim(text));
+[[nodiscard]] static inline std::string trim(std::string text) {
+    return ltrim(rtrim(std::move(text)));
 }

 /**
@@ -80,8 +80,8 @@ enum class text_alignment { left, right, center };
  * \param text the string to manipulate
  * \return a copy of the given string with all occurrences of the ASCII newline character removed
  */
-[[nodiscard]] static inline std::string trim_newline(const std::string& text) {
-    return remove_character(text, '\n');
+[[nodiscard]] static inline std::string trim_newline(std::string text) {
+    return remove_character(std::move(text), '\n');
 }

 /**
```

tristan0x: Nice suggestion. added in [f3f3a70](https://github.com/BlueBrain/nmodl/pull/1071/commits/f3f3a700ce5c9efbe594338cf84e74ab50e796f9)

PR #1082, opened 2023-09-26
Remove legacy units

alkino:


CMakeLists.txt
```diff
@@ -220,21 +216,10 @@ include(${PROJECT_SOURCE_DIR}/src/language/code_generator.cmake)
...
-else()
-  set(LegacyY "/")
-  set(LegacyN "")
-endif()
-configure_file(share/nrnunits.lib.in ${CMAKE_CURRENT_BINARY_DIR}/share/nmodl/nrnunits.lib @ONLY)
-
 # =============================================================================
 # Install unit database to share
 # =============================================================================
+configure_file(share/nrnunits.lib ${CMAKE_CURRENT_BINARY_DIR}/share/nmodl/nrnunits.lib COPYONLY)
```
iomaganaris, line 222: I guess we don't need to configure the file anymore?

alkino: We need to install the file during cmake step to have it early.
If we only install we don't have it for tests and everything.

Here the `configure` is a `COPYONLY`

iomaganaris (approved):

PR #1085, opened 2023-09-28
Unconditionally format generated code but in CI

tristan0x:
Rely on utility `cmake/hpc-coding-conventions/bin/format`

### DIFF
--- a/src/language/code_generator.py
+++ b/src/language/code_generator.py
@@ -12,7 +12,6 @@ import math
 import os
 from pathlib import Path, PurePath
 import shutil
-import stat
 import subprocess
 import tempfile
 
@@ -25,6 +24,13 @@ import utils
 
 LOGGING_FORMAT = "%(levelname)s:%(name)s: %(message)s"
 LOGGER = logging.getLogger("NMODLCodeGen")
+CLANG_FORMAT_EXECUTABLE = (
+    Path(__file__).resolve().parent.parent.parent
+    / "cmake"
+    / "hpc-coding-conventions"
+    / "bin"
+    / "format"
+)
 
 
 class CodeGenerator(
@@ -32,7 +38,6 @@ class CodeGenerator(
         "Base",
         [
             "base_dir",
-            "clang_format",
             "jinja_env",
             "jinja_templates_dir",
             "modification_date",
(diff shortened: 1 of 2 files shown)


src/language/code_generator.py
```diff
@@ -25,14 +24,20 @@
 
 LOGGING_FORMAT = "%(levelname)s:%(name)s: %(message)s"
 LOGGER = logging.getLogger("NMODLCodeGen")
+CLANG_FORMAT_EXECUTABLE = (
```
pramodk, line 27: a pure nitpick: it's not  really a`clang-format` executable?

src/language/code_generator.py
```diff
@@ -98,6 +101,14 @@ def __new__(cls, base_dir, clang_format=None):
...
         return self
 
+    @property
+    def within_ci(self):
+        """Return True if code is running in a continuous-integration environment, else otherwise"""
+        for varname in ["CI", "CI_PIPELINE_ID", "AZURE_ENV_NAME"]:
+            if varname in os.environ:
+                return True
+        return False
+
```
pramodk, line 111: wondering why we need it i.e. what happens if the files are formatted in CI as well? 🤔

src/language/code_generator.py
```diff
@@ -287,9 +298,9 @@ def format_output(self, file):
             language: c++ or cmake
 
         """
-        if self.language == "c++" and self.app.clang_format:
-            LOGGER.debug("Formatting C++ file %s", str(file))
-            subprocess.check_call(self.app.clang_format + ["-i", str(file)])
+        if self.language == "c++" and not self.app.within_ci:
```
pramodk, line 301: A clarification question: if user doesn't have "compatible" clang format installed on his/her machine, what happens in this case? .i.e. if there is no clang-format installed or of a different version than we support, user will still be able to build the nmodl without any error?

pramodk (commented):
Just few clarification questions otherwise LGTM

src/language/code_generator.py
ohm314: `CLANG_FORMAT_COMMAND` then :)

src/language/code_generator.py
ohm314: I think this is taken care of from within the coding convention tool that wraps clang-format?

src/language/code_generator.py
ohm314: I tried to understand this but couldn't quite figure it out from looking at our current CMake files and CI scripts...

PR #1088, opened 2023-10-16
Bugfix: ASSIGNED variables are not considered for Localizer

pramodk:
- previously only RANGE and GLOBAL variables were considered for localiser pass
- exclude STATE variables as they are not usually a good candidate for such optimisations


azure-pipelines.yml
```diff
@@ -136,7 +136,7 @@ stages:
         fi
         ./bin/nrnivmodl-core $(Build.Repository.LocalPath)/test/integration/mod
       env:
-        CMAKE_PKG: 'cmake-3.15.0-Linux-x86_64'
+        CMAKE_PKG: 'cmake-3.17.0-Linux-x86_64'
```
iomaganaris, line 139: This can now be removed since NEURON is fixed

test/unit/visitor/defuse_analyze.cpp
```diff
@@ -445,6 +445,36 @@ SCENARIO("Perform DefUse analysis on NMODL constructs") {
...
+            NEURON {
+                GLOBAL x
+            }
+
+            ASSIGNED {
+                y
+            }
+
+            DERIVATIVE states {
+                x = 1
```
iomaganaris, line 459: Nitpick.. `x=1` could be removed from the `DERIVATIVE` block to make the MOD file valid for code generation?

iomaganaris (approved):

PR #1090, opened 2023-10-17
Remove mol definition from nrnunits.lib

iomaganaris:
Unfortunately I don't remember why I have explicitly added this definition but at the moment it makes the translation of mod files that redefine `(mol)` to fail.
NEURON's version of `nrnunits.lib` doesn't define this unit so I suppose that the right thing is to avoid defining it.


pramodk (approved):
LGTM

PR #1098, opened 2023-11-21
Semantic: check there is at most one DERIVATIVE block

alkino:
This is not forbidden to have more, but due to a bug this fail in nrn.


pramodk (approved):
LGTM

PR #1099, opened 2023-11-21
Semantic: check there is at most one DERIVATIVE block

pramodk:
This is not forbidden to have more, but due to a bug this fail in nrn.

Just to re-trigger CI testing for #1098


1uc:
@pramodk this looks useful. It contains unrelated code which is now outdated. Should this be revived and merged? Alternatively we can close it as too old.

PR #1110, opened 2023-12-04
Refactor slist and dlist according to NEURON

iomaganaris:


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -286,6 +286,13 @@ void CodegenNeuronCppVisitor::print_neuron_includes() {
 
 
 void CodegenNeuronCppVisitor::print_sdlists_init(bool print_initializers) {
+    printer->add_newline(2);
+    printer->add_line("static void _initlists() {");
+    printer->increase_indent();
+    printer->add_multi_line(R"CODE(
+        static int _first = 1;
+        if (!_first) return;
```
1uc, line 294: Why is this check required? Shouldn't this happen once in `_{mod_name}_reg`?

1uc:
How did you check it roughly does what you want it to?

test/unit/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -227,10 +227,7 @@ void _nrn_mechanism_register_data_fields(Args&&... args) {
         THEN("Placeholder registration function is printed") {
             std::string expected_placeholder_reg = R"(/** register channel with the simulator */
     void __test_reg() {
-        /* s */
-        _slist1[0] = {4, 0};
-        /* Ds */
-        _dlist1[0] = {7, 0};
+        _initlists();
```
1uc, line 230: The values inside should be stable. For one or two concrete cases we can look for the string:
```
_slist1[0] = {4, 0}
```
and
```
_dlist1[0] = {7, 0}
```

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Do you see it in other generated code?

iomaganaris: No, I don't find any other usage for it so the check should probably go away

iomaganaris:
> How did you check it roughly does what you want it to?

I just compared the code generated between `nocmodl` and `nmodl`. To actually test this change we need to move forward with code generated for the solvers and find specific MOD files that use those variables

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Thank you, we can add it back if we notice it's needed :)

test/unit/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -227,13 +227,21 @@ void _nrn_mechanism_register_data_fields(Args&&... args) {
...
-            std::string expected_placeholder_reg = R"CODE(/** register channel with the simulator */
-    extern "C" void __test_reg() {
+        THEN("Initialization function for slist/dlist is printed correctly") {
+            std::string expected_initlists = R"(
+    static void _initlists() {
         /* s */
         _slist1[0] = {4, 0};
         /* Ds */
         _dlist1[0] = {7, 0};
+    };)";
```
1uc, line 237: Why is there a semi-colon at the end of the function? More on point, could we check less? Just:
`"_slist1[0] = {4, 0}"` and `"_dlist1[0] = {7, 0}"`. I feel renaming `_initlists` should be permissible, and updating comments shouldn't require updating the tests.

src/codegen/codegen_neuron_cpp_visitor.cpp
iomaganaris: For reference this is now generated with NMODL:
```
    static void _initlists() {
        /* m */
        _slist1[0] = {13, 0};
        /* Dm */
        _dlist1[0] = {16, 0};
        /* h */
        _slist1[1] = {14, 0};
        /* Dh */
        _dlist1[1] = {17, 0};
        /* n */
        _slist1[2] = {15, 0};
        /* Dn */
        _dlist1[2] = {18, 0};
    };
```

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -286,6 +286,10 @@ void CodegenNeuronCppVisitor::print_neuron_includes() {
 
 
 void CodegenNeuronCppVisitor::print_sdlists_init(bool print_initializers) {
```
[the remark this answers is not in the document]
iomaganaris: For the moment it's not necessary but maybe it's used for other parameters. I can add a `[[maybe_unused]]` to show the intention and avoid compilation warning in some compilers possibly.

PR #1115, opened 2023-12-08
Move `print_function_call`.

1uc:
Additionally, adds stubs for `print_net_{send,move,event}_call` to `CodegenNeuronCppVisitor`.


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -743,6 +737,18 @@ void CodegenNeuronCppVisitor::print_codegen_routines() {
     codegen = false;
 }
 
+void CodegenNeuronCppVisitor::print_net_send_call(const ast::FunctionCall& node) {
+    throw std::runtime_error("Not implemented.");
```
iomaganaris, line 741: I wouldn't throw an error at the moment here. This means that the translation of a MOD file that has a `NET_SEND` block will fail even if we want to translate the file to test some other part of it like some part of the `NET_RECEIVE` block only. Same for the rest of the similar functions

iomaganaris (changes requested):

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Having just spent many hours finding out which empty stub function is preventing `exp` from getting printed, I disagree. The first time it throws one can decide that to do, stub it, implement it, etc.

iomaganaris: I'm just afraid that this is going to slow down the development progress. For example when we try to test a new mod file with a new feature we might come across this function and not be able to translate the mod file because of that. An example is the `NET_RECEIVE` block where we can add it piece by piece instead of having to do everything together to avoid adding throws in related functions not yet implemented.
If a an  empty stub is easy to write then we can just do that and if not then it can be left with a `TODO:` comment.
I'm not sure how throwing errors in these particular functions can help debugging why another part of the code isn't printed.

1uc: No it won't, if someone needs it empty, they delete that line. It helps because had `CodegenNeuronCppVisitor::print_function_call` thrown, I would have instantly know which empty stub that needs to be implemented.

Now, first time someone runs NRN with code that requires these blocks, they are immediately told, hey, this is missing; and then they implement it to the degree needed.

iomaganaris: Okay, we can get this in and test out how intrusive this take can be

PR #1118, opened 2023-12-08
Remove `codegen` flag from `Codegen` visitors

iomaganaris:


iomaganaris:
Since most of the CI has passed here: https://github.com/neuronsimulator/nrn/pull/2635 I think that this can be merged. (The only failing CI was related to MPI_Initi issue in MacOS)

PR #1119, opened 2023-12-08
Solve simple ODE.

1uc:
Implements the functionality required to use NEURON as an ODE solver for the state `x`.


1uc:
Missing tests.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -221,7 +236,70 @@ std::string CodegenNeuronCppVisitor::global_variable_name(const SymbolType& symb
...
+    if (g != codegen_global_variables.end()) {
+        return global_variable_name(*g, use_instance);
+    }
+
+    if (varname == naming::NTHREAD_DT_VARIABLE) {
+        return std::string("_nt->_") + naming::NTHREAD_DT_VARIABLE;
+    }
+
+    // t in net_receive method is an argument to function and hence it should
+    // ne used instead of nt->_t which is current time of thread
```
pramodk, line 280:
```suggestion
    // be used instead of nt->_t which is current time of thread
```

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -443,49 +538,78 @@ void CodegenNeuronCppVisitor::print_mechanism_register() {
...
+            auto type = var.is_vdata ? "void*" : default_float_data_type();
+            printer->fmt_line("{}{}* {}{};", qualifier, type, name, value_initialize);
         }
     }
+
+    // printer->fmt_line("{}* {}{};",
+    //                   global_struct(),
+    //                   naming::INST_GLOBAL_MEMBER,
+    //                   print_initializers ? fmt::format("{{&{}}}", global_struct_instance())
+    //                                      : std::string{});
```
pramodk, line 578: a nitpick for clarity: you are keeping commented-out code as those are something `todos` / `next steps` for the implementation?

test/unit/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -181,22 +181,6 @@ void _nrn_mechanism_register_data_fields(Args&&... args) {
...
-    #define s(id) _ml->template fpfield<4>(id)
-    #define ena(id) _ml->template fpfield<5>(id)
-    #define ina(id) _ml->template fpfield<6>(id)
-    #define Ds(id) _ml->template fpfield<7>(id)
-    #define v_unused(id) _ml->template fpfield<8>(id)
-    #define g_unused(id) _ml->template fpfield<9>(id))";
-
-            REQUIRE_THAT(generated,
-                         ContainsSubstring(reindent_and_trim_text(expected_range_macros)));
-        }
```
pramodk, line 199: "logically" macros should have been replaced with checking instance struct check? But with yesterday's discussion, do you plan to not put such checks here but somewhere else?

test/usecases/run_test.sh
```diff
@@ -0,0 +1,14 @@
...
+
+nmodl="$1"
+usecase_dir="$2"
+
+pushd "${usecase_dir}"
+
+rm -r x86_64 tmp || true
+
+nrnivmodl -nmodl "${nmodl}"
+x86_64/special simulate.py
```
pramodk, line 12: as some people develop on apple-m1, would be nice to change it.

pramodk (commented):
Just few minor comments otherwise looks ok.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -443,49 +538,78 @@ void CodegenNeuronCppVisitor::print_mechanism_register() {
...
-        method_name(naming::NRN_INIT_METHOD));
+    printer->push_block("for (int id = 0; id < nodecount; id++)");
+    print_initial_block(info.initial_node);
+    printer->pop_block();
 
-    codegen = false;
+    printer->pop_block();
+}
+
+void CodegenNeuronCppVisitor::print_initial_block(const InitialBlock* node) {
```
iomaganaris, line 607: Should be moved above `print_global_function_common_code`. Let's try to keep the order of declaration of functions between `CodegenCoreneuronCppVisitor` and `CodegenNeuronCppVisitor` the same to easily diff them

test/usecases/leonhard/simulate.py
```diff
@@ -0,0 +1,27 @@
+import numpy as np
+
+from neuron import h, gui
+from neuron.units import ms
+
+nseg = 1
+
+s = h.Section()
+s.insert("leonhard")
+s.nseg = nseg
```
iomaganaris, line 10: The `Section` creation (apart from the mechanism insertion code) could be moved in a separate function and file since this is going to be used in multiple test files. Feel free to do this in a separate PR if you prefer

iomaganaris (commented):
Nice and welcome changes :+1: Few suggestions here and there

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Kinda, yes. That function is almost identical to the cnrn one. Right now I don't need globals. That will change. Next mini-task is either ODEs with array range variables or with global variables

test/unit/codegen/codegen_neuron_cpp_visitor.cpp
1uc: I don't see how the mod file can compile if the instance is missing. Therefore, since I'm having difficulty pinning down what this test contributed, I simply removed it.

CMakeLists.txt
```diff
@@ -19,6 +19,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/bin)
 # =============================================================================
 option(NMODL_ENABLE_PYTHON_BINDINGS "Enable pybind11 based python bindings" ON)
 option(NMODL_ENABLE_TESTS "Enable build of tests" ON)
+option(NMODL_ENABLE_USECASES
```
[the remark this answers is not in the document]
1uc: I'd keep it simple and use `$PATH`. If we need complicated we can do that later. Currently there's also nothing you can do with `NEURON_DIR`. There's no place you can send it, is simply assumes that `nrnivmodl` is in the path and `neuron` can be imported from Python.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -199,7 +199,22 @@ void CodegenNeuronCppVisitor::print_namespace_stop() {
...
+    auto name = symbol->get_name();
+    auto dimension = symbol->get_length();
+    // auto position = position_of_float_var(name);
+    // clang-format off
+    if (symbol->is_array()) {
+        if (use_instance) {
+            return fmt::format("(inst.{}+id*{})", name, dimension);
+        }
+        throw std::runtime_error("Not implemented.");
+        // return fmt::format("(data + {}*pnodecount + id*{})", position, dimension);
```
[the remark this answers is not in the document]
1uc: It'll come back the very instant the throw goes away. The idea is first person to trigger the assert gets to check that there are proper tests, etc. I've already started on checking that ODEs with array range variables work. This is just a way of creating merge conflicts :)

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -384,6 +462,23 @@ void CodegenNeuronCppVisitor::print_global_variables_for_hoc() {
     printer->add_line("};");
 }
 
+void CodegenNeuronCppVisitor::print_make_instance() const {
```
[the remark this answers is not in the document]
1uc: This is what I don't understand about the testing strategy. What do we need to test that isn't checked by compiling something like the `leonhard.mod`?

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -443,49 +538,78 @@ void CodegenNeuronCppVisitor::print_mechanism_register() {
...
+        printer->fmt_line("{}* {}{};",
+                          type,
+                          name,
+                          print_initializers ? fmt::format("{{&coreneuron::{}}}", name)
+                                             : std::string{});
+    }
+    for (auto& var: codegen_float_variables) {
+        const auto& name = var->get_name();
+        // auto type = get_range_var_float_type(var);
+        // auto qualifier = is_constant_variable(name) ? "const " : "";
```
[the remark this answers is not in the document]
1uc: Right the two lines can be deleted. When we get globals to run the function will likely become identical the the cnrn variant again and we can lift it.

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: done

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -221,7 +236,70 @@ std::string CodegenNeuronCppVisitor::global_variable_name(const SymbolType& symb
 /// TODO: Edit for NEURON
 std::string CodegenNeuronCppVisitor::get_variable_name(const std::string& name,
                                                        bool use_instance) const {
-    return name;
+    // const std::string& varname = update_if_ion_variable_name(name);
+    const std::string& varname = name;
```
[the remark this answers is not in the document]
1uc: Absolutely, but I can't test ion related stuff right now. Here's an issue for that: #1122.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -221,7 +236,70 @@ std::string CodegenNeuronCppVisitor::global_variable_name(const SymbolType& symb
 /// TODO: Edit for NEURON
```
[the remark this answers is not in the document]
1uc: done

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -199,7 +199,22 @@ void CodegenNeuronCppVisitor::print_namespace_stop() {
...
-    return symbol->get_name();
+    auto name = symbol->get_name();
+    auto dimension = symbol->get_length();
+    // auto position = position_of_float_var(name);
+    // clang-format off
+    if (symbol->is_array()) {
+        if (use_instance) {
+            return fmt::format("(inst.{}+id*{})", name, dimension);
+        }
+        throw std::runtime_error("Not implemented.");
```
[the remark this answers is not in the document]
1uc: The effective way of finding these is `gdb` followed by `bt`. If it doesn't print the backtrace automatically.

src/codegen/codegen_neuron_cpp_visitor.cpp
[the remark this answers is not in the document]
iomaganaris: Okay. I would prefer having a TODO which would do the same thing instead of the commented code that isn't immediately understood why it's there but if you're working on that at the moment I'm fine with it

test/unit/codegen/codegen_neuron_cpp_visitor.cpp
pramodk: Oh! I remember the reason for such type of test (from coreneuron codegen) - the order in which the variables were printed was introducing subtle bugs and it was pain to debug! i.e. nocmodl was printing `ena, ina` and new nmodl was printing `ina, ena`.

But with neuron and coreneuron codegen coming from new NMODL, we will have no/less issues. It's ok if you remove them from here. The ordering aspect could be tested in other ways.

src/codegen/codegen_neuron_cpp_visitor.cpp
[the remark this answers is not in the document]
1uc: It's break for uninteresting reasons (renaming `_ml`), it's very weak and requires manual changes regularly. Then there's the readability aspect IMO code inside a string isn't top. Then when it fails it's one semi long string followed by "not found in: ${REALLY_LONG_STRING}" and as a human I instantly find the line: right there. Naturally the computer is correct, so need to copy both string into a editor to align them so one can find a surplus space.

test/unit/codegen/codegen_neuron_cpp_visitor.cpp
1uc: That does sounds like very limited fun.

test/usecases/run_test.sh
1uc: If you have one please try, documentation suggests that it should be `uname -m`.

1uc:
@pramodk @iomaganaris IFAICT this should address all concerns.

pramodk (approved):

test/usecases/run_test.sh
iomaganaris: I can confirm it works on the M1 Mac

test/usecases/leonhard/leonhard.mod
```diff
@@ -0,0 +1,16 @@
+NEURON {
```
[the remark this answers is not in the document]
1uc: Good now we have twice as many tests!

test/usecases/run_test.sh
1uc: ty

iomaganaris (approved):

PR #1124, opened 2023-12-14
Skip printing of INDEPENDENT block from AST to NMODL

### LINKED ISSUE #1123: Parsing and writing `INDEPENDENT` block is broken
```python
import nmodl

print(nmodl.to_nmodl(nmodl.NmodlDriver().parse_string("INDEPENDENT {t FROM 0 TO 1 WITH 1 (ms)}")))
```

outputs an invalid `INDEPENDENT` block:

```
INDEPENDENT {
t}
```

which `nrnivmodl` gives a syntax error about:

```
syntax error:
 name FROM number TO number WITH integer
 at line 23 in file Cav2_1__0.mod
t}
 ^
Translating Ca__granule_cell.mod into /home/robin/.cache/Glia/92db3105/x86_64/Ca__granule_cell.c
make: *** [makemod2c_inc:35: Cav2_1__0.c] Error 1
make: *** Waiting for unfinished jobs....
Translating Cav2_2__0.mod into /home/robin/.cache/Glia/92db3105/x86_64/Cav2_2__0.c
Thread Safe
Thread Safe
```

pramodk:
* As INDEPENDENT is deprecated and has no impact/role in NEURON/NMODL models, we have removed the AST nodes related to this construct.
* When user provide an input with INDEPENDENT {} block then just add a comment and simply skip printing it.
* Update test accordingly

fixes #1123

With this

```console
$ cat test.py

import nmodl
print(nmodl.to_nmodl(nmodl.NmodlDriver().parse_string("NEURON {} INDEPENDENT {t FROM 0 TO 1 WITH 1 (ms)}")))
```

and it will give

```console
NEURON {
}

: INDEPENDENT block is deprecated and has no effect in the NEURON model. Skipped!
```

We can think of alternatives: bring back some part of the old implementation and reproduce the same `INDEPENDENT` block that the user provides. But I am not sure there is any more value to it as the `INDEPENDENT` is not useful in the MOD file.

### DIFF
--- a/src/language/node_info.py
+++ b/src/language/node_info.py
@@ -155,6 +155,7 @@ PROGRAM_BLOCK = "Program"
 STATEMENT_BLOCK_NODE = "StatementBlock"
 STRING_NODE = "String"
 UNIT_BLOCK = "UnitBlock"
+INDEPENDENT_BLOCK_NODE = "IndependentBlock"
 
 # name of variable in prime node which represent order of derivative
 ORDER_VAR_NAME = "order"
--- a/src/language/nodes.py
+++ b/src/language/nodes.py
@@ -60,6 +60,10 @@ class BaseNode:
         return self.class_name == node_info.PRIME_NAME_NODE
 
     @property
+    def is_independent_node(self):
+        return self.class_name == node_info.INDEPENDENT_BLOCK_NODE
+
+    @property
     def is_program_node(self):
         """
         check if current node is main program container node
--- a/src/language/templates/visitors/nmodl_visitor.cpp
+++ b/src/language/templates/visitors/nmodl_visitor.cpp
@@ -89,6 +89,10 @@ using namespace ast;
 
 {%- for node in nodes %}
 void NmodlPrintVisitor::visit_{{ node.class_name|snake_case}}(const {{ node.class_name }}& node) {
+    {% if node.is_independent_node %}
+        printer->add_element(": INDEPENDENT block is deprecated and has no effect in the NEURON model. Skipped!");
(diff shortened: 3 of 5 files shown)


alkino (approved):
Way better than my solution

Helveg:
What about code generations for versions of NEURON that still supported the syntax, and more generally for the future, how will the package `nmodl` deal with different NMODL (the language) versions. Is there an official spec yet for NMODL?

pramodk:
> What about code generations for versions of NEURON that still supported the syntax,

Just to clarify: NEURON **as well as this new NMODL** accept the `INDEPENDENT` block as input. But, internally it's unused and doesn't have any impact on the generated code. So it's basically `no-op`.

> and more generally for the future, how will the package nmodl deal with different NMODL (the language) versions

Valid question but I don't have a good answer. Until today, everything is backward compatible and from the user's perspective, nothing has been removed.

In this PR I am taking a shortcut to skip INDEPENDENT blocks because they are not useful anyway.

> Is there an official spec yet for NMODL?

`spec` is a strong word and following pages are definitely incomplete:
* https://nrn.readthedocs.io/en/latest/hoc/modelspec/programmatic/mechanisms/nmodl.html
* https://nrn.readthedocs.io/en/latest/hoc/modelspec/programmatic/mechanisms/nmodl2.html

But in terms of "technical structure", full language AST is expressed in this (incompletely documented) YAML file: https://github.com/BlueBrain/nmodl/blob/master/src/language/nmodl.yaml

And my hope was/is that with enough documentation, this can serve as something close to spec.

PR #1130, opened 2023-12-20
Enable trivial cases of global state variables.

1uc:
Adds the functionality required to make trivial cases involving GLOBAL variables work.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -418,7 +418,7 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
     printer->fmt_push_block("struct {}", global_struct());
 
     if (!info.ions.empty()) {
-        throw std::runtime_error("Not implemented, global ions.");
+        // TODO implement these when needed.
     }
 
     if (info.point_process) {
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -225,10 +225,13 @@ std::string CodegenNeuronCppVisitor::int_variable_name(const IndexVariableInfo& 
 }
 
 
-/// TODO: Edit for NEURON
 std::string CodegenNeuronCppVisitor::global_variable_name(const SymbolType& symbol,
                                                           bool use_instance) const {
-    return symbol->get_name();
+    if (use_instance) {
+        return fmt::format("inst.{}->{}", naming::INST_GLOBAL_MEMBER, symbol->get_name());
+    } else {
+        return fmt::format("{}.{}", global_struct_instance(), symbol->get_name());
+    }
 }
 
 
@@ -391,6 +394,8 @@ void CodegenNeuronCppVisitor::print_sdlists_init([[maybe_unused]] bool print_ini
 
 
 void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_initializers) {
+    const auto value_initialize = print_initializers ? "{}" : "";
+
     /// TODO: Print only global variables printed in NEURON
     printer->add_newline(2);
     printer->add_line("/* NEURON global variables */");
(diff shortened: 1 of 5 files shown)


1uc:
This PR includes preparatory commits to move some printing code. Which I'd like to split away once accepted.

It also includes the generated `.cpp` file as a trial run to see if it helps with understanding what's happening.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -405,6 +410,73 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
...
+    }
+
+    if (info.table_count > 0) {
+        throw std::runtime_error("Not implemented, global table count.");
+    }
+
+    for (const auto& var: info.state_vars) {
+        auto name = var->get_name() + "0";
+        auto symbol = program_symtab->lookup(name);
+        if (symbol == nullptr) {
```
iomaganaris, line 443:
```suggestion
        if (!symbol) {
```

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -405,6 +410,73 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
...
+            printer->fmt_line("{} {}{};", float_type, name, value_initialize);
+            codegen_global_variables.push_back(make_symbol(name));
+        }
+    }
+
+    for (const auto& var: info.global_variables) {
+        auto name = var->get_name();
+        auto length = var->get_length();
+        if (var->is_array()) {
```
iomaganaris, line 452: I think we're missing the proper implementation for this in the CoreNEURON codegen as well. Could you create an issue about this? Could be a `good first issue` for newcomers

test/usecases/global_breakpoint/leonhard.cpp
```diff
@@ -0,0 +1,163 @@
+/*********************************************************
```
iomaganaris, line 1: I think that before merging this PR this file should be deleted since it's not contributing to any test. If #1131 gets merged first with proper tests on those "golden" reference files then it can be added

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -405,6 +410,73 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
...
+
+    if (!info.thread_variables.empty()) {
+        throw std::runtime_error("Not implemented, global thread variables.");
+    }
+
+    if (info.table_count > 0) {
+        throw std::runtime_error("Not implemented, global table count.");
+    }
+
+    for (const auto& var: info.state_vars) {
```
iomaganaris, line 440: I'm wondering whether those parts of this function (or others) that are the same as `CodegenCoreneuronCppVisitor` should be moved into some function that be called by both visitors instead of duplicating the code

test/usecases/global_breakpoint/leonhard.cpp
1uc: Probably, yes. However, did it help with reviewing this PR?

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: What is "this"? Adding an initializer for a const array?

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Midterm, yes. The hope is that as we add more and more, the two implementations will automatically become more similar again. Once we fill out (and test) all of the

        throw std::runtime_error("Not implemented, global thread variables.");

we'd be in a better place to figure out the differences between NRN and CNRN (I think it's `inst->` vs `inst.` or something similarly minor) and solve it neatly.

test/usecases/global_breakpoint/simulate.py
```diff
@@ -0,0 +1,27 @@
...
+h.run()
+
+x = np.array(x_hoc.as_numpy())
+t = np.array(t_hoc.as_numpy())
+
+x_exact = 2.0 * np.ones_like(t)
+x_exact[0] = 42;
+abs_err = np.abs(x - x_exact)
+
+assert np.all(abs_err < 1e-12), f"{abs_err=}"
```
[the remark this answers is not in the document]
1uc: No amount of testing is overkill. Plus it uncovered something surprising to me.

src/codegen/codegen_neuron_cpp_visitor.cpp
iomaganaris: Yes that's what I meant. However looking at this again I am not sure what the use case that isn't working. I can't find how a `GLOBAL` array can be const or initialized in a MOD file except inside a `FUNCTION/PROCEDURE/INITIAL/STATE/BREAKPOINT` block which should work properly.
I saw that the `TODO` was added in https://github.com/BlueBrain/nmodl/pull/904. I am wondering whether @pramodk has any idea what's missing?

src/codegen/codegen_neuron_cpp_visitor.cpp
iomaganaris: I hope the same as well. We can keep it as it is for now and hopefully we can get rid of more stuff in the end

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: I wasn't able to guess the correct syntax for global array variables. Meaning I tried a couple of sensible looking ideas against NRN + nocmodl, but they didn't work.

iomaganaris (approved):

PR #1135, opened 2024-01-09
Improve run_test.sh to check against NRN.

1uc:
We can run `run_test.sh` with a `special` generated by `nocmodl`. This helps debug `simulate.py` itself, which is useful when things get very subtle.


1uc:
Merge after #1134.

1uc:
An example of when is this useful is:
```
INITIAL {
  x = 42
}

BREAKPOINT {
  x = c
}
```
and now the question is should `x[0] == 42.0` or `x[0] == x`? This is risky to answer without checking against NRN + nocmodl.

PR #1138, opened 2024-01-11
Add semantic analysis check for infinite loops with function calls

iomaganaris:
- First step towards support in NEURON code generation for `FUNCTION` and `PROCEDURE`
- Related to the changes in https://github.com/neuronsimulator/nrn/pull/2475 that require that there is no loop between functions and procedures
- Added token in `ast::FunctionCall` parsing for printing nice error
- Added unit test
- Improved a bit the error for other semantic checks to show the problematic function/procedure name
- The usage of this feature can be seen in https://github.com/BlueBrain/nmodl/compare/magkanar/nrn_proc_func

### DIFF
--- a/src/visitors/semantic_analysis_visitor.cpp
+++ b/src/visitors/semantic_analysis_visitor.cpp
@@ -108,19 +108,19 @@ void SemanticAnalysisVisitor::visit_function_call(const ast::FunctionCall& node)
 
 void SemanticAnalysisVisitor::visit_procedure_block(const ast::ProcedureBlock& node) {
     /// <-- This code is for check 1
-    visited_function_or_procedure_blocks.push_front(&node);
+    visited_function_or_procedure_blocks.push_back(&node);
     one_arg_in_procedure_function = node.get_parameters().size() == 1;
     node.visit_children(*this);
-    visited_function_or_procedure_blocks.pop_front();
+    visited_function_or_procedure_blocks.pop_back();
     /// -->
 }
 
 void SemanticAnalysisVisitor::visit_function_block(const ast::FunctionBlock& node) {
     /// <-- This code is for check 1
-    visited_function_or_procedure_blocks.push_front(&node);
+    visited_function_or_procedure_blocks.push_back(&node);
     one_arg_in_procedure_function = node.get_parameters().size() == 1;
     node.visit_children(*this);
-    visited_function_or_procedure_blocks.pop_front();
+    visited_function_or_procedure_blocks.pop_back();
     /// -->
 }
 
--- a/src/visitors/semantic_analysis_visitor.cpp
+++ b/src/visitors/semantic_analysis_visitor.cpp
@@ -23,6 +23,8 @@
 namespace nmodl {
 namespace visitor {
 
+using symtab::syminfo::NmodlType;
+
 bool SemanticAnalysisVisitor::check(const ast::Program& node) {
     check_fail = false;
 
@@ -93,8 +95,10 @@ void SemanticAnalysisVisitor::visit_function_call(const ast::FunctionCall& node)
         return;
     }
     auto func_symbol = psymtab->lookup(func_name);
-    // If symbol is not found or there are no AST nodes for it return
-    if (!func_symbol || func_symbol->get_nodes().empty()) {
+    // If symbol is not found or there are no AST nodes for it or it's not a function or procedure
+    // return
+    if (!func_symbol || func_symbol->get_nodes().empty() ||
+        !func_symbol->has_any_property(NmodlType::function_block | NmodlType::procedure_block)) {
         return;
     }
     const auto func_block = func_symbol->get_nodes()[0];
--- a/src/visitors/semantic_analysis_visitor.cpp
+++ b/src/visitors/semantic_analysis_visitor.cpp
@@ -8,6 +8,7 @@
 #include "visitors/semantic_analysis_visitor.hpp"
 #include "ast/breakpoint_block.hpp"
 #include "ast/function_block.hpp"
+#include "ast/function_call.hpp"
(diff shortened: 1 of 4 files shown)


src/visitors/semantic_analysis_visitor.cpp
```diff
@@ -71,27 +73,56 @@ void SemanticAnalysisVisitor::visit_program(const ast::Program& node) {
...
+                                                   });
+    if (func_already_visited != visited_function_or_procedure_blocks.end()) {
+        logger->critical(
+            fmt::format("SemanticAnalysisVisitor :: Recursive function call of \"{}\" in {}",
+                        func_name,
+                        node.get_token()->position()));
+        check_fail = true;
+        return;
+    }
+    auto func_symbol = psymtab->lookup(func_name);
```
matz-e, line 95: is there a big penalty to have this just be
```suggestion
    auto func_symbol = node.get_symbol_table()->lookup(func_name);
```
and get around having a member pointer to the program table in the visitor class?

iomaganaris: Here we need the global symbol table that has all the symbols in the mod file to be able to search whether there is a `FUNCTION` or `PROCEDURE` defined with the `func_name` so we need to set `psymtab` when we visit the `ast::Program` which is the top level symbol table

test/unit/visitor/semantic_analysis.cpp
```diff
@@ -192,3 +192,21 @@ SCENARIO("At most one DERIVATIVE block", "[visitor][semantic_analysis]") {
...
+                a = 42
+            }
+            PROCEDURE b() {
+                LOCAL x
+                x = a()
+            }
+        )";
+        THEN("Semantic analysis should fail") {
+            REQUIRE(run_semantic_analysis_visitor(nmodl_text));
+        }
```
1uc, line 210: This reads like the analysis succeeded on something that contains an infinite loop. Shouldn't there be a `REQUIRE_THROWS` or a `!` something to make it assert failure?

1uc: Thank you!

src/visitors/semantic_analysis_visitor.hpp
```diff
@@ -43,14 +46,14 @@ class SemanticAnalysisVisitor: public ConstAstVisitor {
...
     /// true if accelerator backend is used for code generation
     bool accel_backend = false;
     /// true if the procedure or the function contains only one argument
     bool one_arg_in_procedure_function = false;
-    /// true if we are in a procedure block
-    bool in_procedure = false;
-    /// true if we are in a function block
-    bool in_function = false;
+    /// list of visited Function or Procedure Blocks
+    std::forward_list<const ast::Block*> visited_function_or_procedure_blocks;
```
1uc, line 56: What's the reason for an `std::forward_list`?

1uc:
What happens to recursive or mutually recursive code? The MODL equivalent of:
```
def f(i):
  if i == 0:
    return 0
  else:
    return f(i-1) + f(i-2)
```
and
```
def f(x):
  y = 0.5 * x
  shrink_to_fit()

def shirk_to_fit():
  if y > 0.1:
    f(y)
```

src/visitors/semantic_analysis_visitor.hpp
iomaganaris: It's a container that allows fast access to the first element and searching in it with `std::find_if`

iomaganaris:
> What happens to recursive or mutually recursive code? The MODL equivalent of:
> 
> ```
> def f(i):
>   if i <= 1:
>     return 1
>   else:
>     return f(i-1) + f(i-2)
> ```
> 
> and
> 
> ```
> def f(x):
>   y = 0.5 * x
>   shrink_to_fit()
> 
> def shirk_to_fit():
>   if y > 0.1:
>     f(y)
> ```

Good point even if I believe it's a very rare scenario in NMODL files. I have to check how NEURON handles that

src/visitors/semantic_analysis_visitor.hpp
1uc: Sounds like a `std::vector`. To me picking an exotic container, i.e. anything not `std::vector` reads like a lot of thought went into the choice, and even if I suspect the choice is suboptimal, I can't change it without thorough testing. The rumour is that `std::vector` is usually faster due to better cache locality and constant time random access to individual elements.

Lists have a nice property that references/pointers aren't invalidated as the container grows. Hence the question when changing this is: are parts of the code holding references, pointers to the `ast::Block*`.

1uc:
My understanding is that after visiting a procedure or function we push the node into a list. Then, when we enter a function-call node we check it's name against the names of already visited nodes. What I don't understand is why:
```
def shirk_to_fit():
  f(1)
  f(2)
```
would pass, because in my head, the second time we call `f` it would detect that we've visited `f` already and mark the analysis as failed.

src/visitors/semantic_analysis_visitor.hpp
iomaganaris: > The rumour is that std::vector is usually faster due to better cache locality and constant time random access to individual elements.

That's mostly true I believe. The only reason why I selected the `std::forward_list` is that because it might be faster than `std::vector` pushing and poping elements from its edges (to be more specific, its front). `std::vector` allows pushing and poping elements on its back but I believe that there might be cases that this will be slower than pushing and poping elements from it. I was looking to using a container that resembles a stack that is however searchable with `std::find_if`.
To increase the readability and because I believe that the performance benefit of `std::forward_list` will be insignificant compared to `std::vector` since the number of elements they hold should be really small I can revert to using an `std::vector`

iomaganaris:
Superseded by https://github.com/BlueBrain/nmodl/pull/1139

PR #1106, opened 2023-12-04
Add more logic for POINT_PROCESS registration function

iomaganaris:
- Added call to `point_register_mech()` needed for `POINT_PROCESS`es and related function definitions
- Added logic in `nrn_alloc` needed to create reasonable `POINT_PROCESS` tests
- Added test for `POINT_PROCESS`
- Added test for `PARAMETER`
- Added small logic for `ION`s as well in `_<mech>_reg()`
- Added small logic for `diam` as well in `_<mech>_reg()`

PR #1141, opened 2024-01-23
FUNCTION and PROCEDURE codegen for NEURON

iomaganaris:
- Based on https://github.com/BlueBrain/nmodl/pull/1139 and https://github.com/BlueBrain/nmodl/pull/1106
- This branch is based on the branches of the above two PRs so when these two are merged to master the diffs of this PR will be reduced


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -456,6 +821,100 @@ void CodegenNeuronCppVisitor::print_global_variables_for_hoc() {
     printer->add_line("{nullptr, nullptr, 0}");
     printer->decrease_indent();
     printer->add_line("};");
+
+    printer->add_newline(2);
+    printer->add_line("/* declaration of user functions */");
+    for (const auto& procedure: info.procedures) {
+        const auto proc_name = procedure->get_node_name();
+        if (proc_name[0] == '_') {
```
pramodk, line 829: what this means? i.e. NMODL doesn't allow methods/variables with `_`? and if its allowed, why they need to be skipped declarations?

test/usecases/func_proc_pnt/simulate.py
```diff
@@ -0,0 +1,15 @@
...
+
+pnt_proc.set_x_42()
+
+assert pnt_proc.x == 42
+
+pnt_proc.set_x_a(13.7)
+
+assert pnt_proc.x == 13.7
+
+assert pnt_proc.get_a_42(42) == 84
```
pramodk, line 15: just to understand the current convention (/ philosophy): a test like this could also easily cover the purpose of `test/usecases/point_process/simulate.py`. And this could be easily also combined in `test/usecases/global_breakpoint/simulate.py`.

 Do we try to minimise the number of test cases wherever possible? (of course, without making them too much complicate)

pramodk (commented):
@iomaganaris: I went through this, just to be sure that if there are any major questions then I can post them / get them clarified. 

Few minor questions/comments otherwise it's OK and on our side we can move forward this.

test/usecases/func_proc_pnt/simulate.py
iomaganaris: As Luc suggested in Slack having small and independent tests help with the actual testing and understanding of how NMODL features work. I agree with this for the moment and while the code is actively being developed.

src/codegen/codegen_neuron_cpp_visitor.cpp
iomaganaris: This is case is related to https://github.com/neuronsimulator/nrn/blob/d7750508d713877bd55e25186fa3b56df615ae16/src/nmodl/nocpout.cpp#L222. I think there is some special handling if the `SUFFIX` of a mechanism is set to `nothing`.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -129,13 +377,19 @@ void CodegenNeuronCppVisitor::print_function(const ast::FunctionBlock& node) {
 
 /// TODO: Edit for NEURON
 std::string CodegenNeuronCppVisitor::internal_method_arguments() {
-    return {};
+    return "_ml, id, _ppvar, _thread, _nt";
 }
 
 
 /// TODO: Edit for NEURON
```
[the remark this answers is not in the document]
iomaganaris: It's true that there were some unnecessary `TODO`s. I've now removed them

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -89,36 +92,281 @@ void CodegenNeuronCppVisitor::print_atomic_reduction_pragma() {
...
+        printer->add_multi_line(R"CODE(
+            static double _hoc_has_loc(void* _vptr) {
+                return has_loc_point(_vptr);
+            }
+        )CODE");
+        printer->add_multi_line(R"CODE(
+            static double _hoc_get_loc_pnt(void* _vptr) {
+                return (get_loc_point_process(_vptr));
+            }
+        )CODE");
```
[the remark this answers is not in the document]
iomaganaris: I combined some of them related to this PR. I think there are some more opportunities to combine `add_line`s and `add_multi_line`s across the `CodegenNeuronCppVisitor` but maybe this can come in another PR

pramodk (approved):
LGTM.

- I have merged the master and reviewed the changes.
- Only noticeable thing I have changes is in c3edb59e7488b1ca6be6392da3e6940173cefb1b 
```diff
@@ -853,31 +852,19 @@ void CodegenNeuronCppVisitor::print_global_variables_for_hoc() {
     printer->add_line("/* declaration of user functions */");
     for (const auto& procedure: info.procedures) {
         const auto proc_name = procedure->get_node_name();
-        if (proc_name[0] == '_') {
-            continue;
-        }
```
i.e. removal of check of `_` in the function/procedure name. NMODL language doesn't allow it and I believe the `info.procedures` & `info.functions` don't contain any methods with `_` at the beginning. So I am removing that check.

Also, I check the test with a `GLOBAL` variable usage in efe963c67738d2d7350fd69205329da52d61153e and it's working fine.

PR #1144, opened 2024-01-29
Make sonarcloud checks non-failing

ohm314:
We want sonarcloud runs to just inform developers on the state of their PR rather than stopping them from merging.


1uc (approved):
Seems plausible :)

PR #1145, opened 2024-02-01
Update CMake for testing

JCGoran:
The pytest package should only be required if we configure the project with the `NMODL_ENABLE_TESTS` flag. Additionally, if we configure it with `NMODL_ENABLE_USECASES`, we need to have numpy and NEURON installed.


CMakeLists.txt
```diff
@@ -213,6 +216,9 @@ if(NOT NMODL_AS_SUBPROJECT AND NMODL_ENABLE_TESTS)
 
   if(NMODL_ENABLE_USECASES)
     add_subdirectory(test/usecases)
+    # usecases require numpy and neuron
+    cpp_cc_find_python_module(numpy REQUIRED)
+    cpp_cc_find_python_module(neuron REQUIRED)
```
1uc, line 221: Does it matter that we only need them at runtime? It's possible to build NMODL without. Only by the time one runs the tests does one need NEURON and numpy.

JCGoran: In that case, should I set them as `OPTIONAL`, and post a warning message that the usecases may not run if those are not installed, or just remove these lines altogether?

1uc: What would you like to accomplish with those lines?

ohm314: I would not make them optional. These tests are run by developers who need the packages if they want to run these types of tests. It's true that technically they are not needed for building, but they are needed right after for testing - so I think this is a useful addition.

ohm314 (approved):

JCGoran:
Closing as we removed checking for Python packages via CMake in #1147

PR #1147, opened 2024-02-05
Transition to `pyproject.toml`

### LINKED ISSUE #462: Move nmodl python files into subdirectory
Currently the nmodl python module files are in a top level directory called `nmodl`. Having the same name as the package itself it often leads to confusion especially when testing the python package. Various scenarios fail because python uses the current working dir as a search dir for modules. This way the local `nmodl` directory will take precedence over an installed `nmodl` python module (eg. in `site-packages`). We have several workarounds to address this, but we keep finding scenarios, where confusing errors show up.
A proper fix is needed. For example the `nmodl` directory should be moved into another dir, maybe named `pkg`. It is not sufficient to move the directory as this will entail a number of changes in the build system and CI:

- CMakeLists.txt need to be checked and fixed
- setup.py needs to be checked and fixed
- azure CI will need checking for both the testing and wheel building pipelines
- travis CI might also need a check

This list is non-exhaustive but should cover most of what is needed.

### LINKED ISSUE #1153: fmtlib places files outside of `.data` when not linking against Python
When using `-DLINK_AGAINST_PYTHON=FALSE` (for instance, for building wheels), Eigen and related libs are placed in `nmodl/.data`, but fmtlib doesn't respect this and stubbornly puts items in other places. Below is an excerpt from the `install_manifest.txt` file (tested with latest master):

```plaintext
/lib/libfmt.a
/include/fmt/args.h
/include/fmt/chrono.h
/include/fmt/color.h
/include/fmt/compile.h
/include/fmt/core.h
/include/fmt/format.h
/include/fmt/format-inl.h
/include/fmt/os.h
/include/fmt/ostream.h
/include/fmt/printf.h
/include/fmt/ranges.h
/include/fmt/std.h
/include/fmt/xchar.h
/lib/cmake/fmt/fmt-config.cmake
/lib/cmake/fmt/fmt-config-version.cmake
/lib/cmake/fmt/fmt-targets.cmake
/lib/cmake/fmt/fmt-targets-release.cmake
/lib/pkgconfig/fmt.pc
/nmodl/.data/lib/libpywrapper.dylib
/nmodl/_nmodl.cpython-311-darwin.so
/nmodl/.data/include/newton/newton.hpp
/nmodl/.data/include/Eigen/CholmodSupport
/nmodl/.data/include/Eigen/StdVector
/nmodl/.data/include/Eigen/StdDeque
... other files in /nmodl/.data/
```

I don't know if this is intended or not, but it's somewhat annoying since the wheel built with `scikit-build-core` (see #1147) has `lib` (or `lib64` on Linux) and `include` directories along with the `nmodl` one.

JCGoran:
## Overview and context of changes

This PR moves all of the build logic from the dynamic `setup.py` into the new static `pyproject.toml` configuration file, as described in [PEP 621](https://peps.python.org/pep-0621/). Since NMODL has a handy Python wrapper, it makes sense to use a modern Python build system.

Some of the notable changes and explanations for them:

- wheels can now be built using `pip wheel . --no-deps`, with an optional `--wheel-dir [DIR]` for output
- the actual building is done by `scikit-build-core`, and any options can be passed by adding `-C [OPTION]=[VALUE]` to the above call (on MacOS, this is needed to specify the build dir as it fails in a temp dir due to a bug in cmake)
- remove `setup.py` and all various requirements files in favor of `pyproject.toml` (except a single `requirements.txt` which has all of them too)
- move some `setup.cfg` options to `pyproject.toml` as well (notably flake8 and sphinx do not support it, so we leave those out for now)
- ~~due to the fact that `pip` doesn't provide a way of parsing the `pyproject.toml` file using the CLI, to install _all_ dependencies (build, run, optional), we need to do `pip install 'pip-tools>=7.4.0'`, followed by `pip install -r <(pip-compile --all-build-deps --all-extras --no-strip-extras 2>&1)` Note that this will also leave a `requirements.txt` in your current dir (not sure if this is a feature or a bug of `pip-compile`)~~ use `pip install -r requirements.txt` to install requirements
- move documentation generation from `setup.py` to a standalone script `packaging/generate_docs.sh` which generates the documentation (see docs for usage)
- `pip install -e .` (i.e. editable install) doesn't quite work due to the strange way we bundle everything, but `pip install -C build-dir=[SOMEDIR]` is fast enough so there isn't much of a difference. If we later merge #1179 we can also sidestep any issues with imports
- add `NMODL_BUILD_WHEEL` to cmake config so we can signal we are building a wheel (this sets the appropriate flags automatically)
- remove random `include` and `lib` stuff from wheel because they're not needed (looking at you, fmtlib and eigen). Also fixes #1153
- due to the limitations of PEP 621, it's not allowed to change the name of the package dynamically (we build `nmodl-nightly` wheels sometimes), so as a workaround there is the `packaging/change_name.py` script, which can be used for changing the name of the package (only relevant in the CI, so users shouldn't touch it)
- `pyproject.toml` doesn't allow an empty version (we now call cmake _after_ Python reads `pyproject.toml`), so we can set it dynamically using `setuptools_scm` (uses git tags), or even statically using `SETUPTOOLS_SCM_PRETEND_VERSION` (we seem to use something like `git describe --tags | cut -d '-' -f 1,2 | tr - .` to set it) in the CI (I didn't want to do too much refactoring with versions)
- `pywheel` dir is gone, and the `_binwrapper` script is moved to `nmodl` dir with minimal changes so we can actually set the right script target
- Python bindings are now in `python/nmodl` instead of `nmodl`. Fixes #462.
- `NMODL_WHEEL_VERSION` was removed; if you want to specify a custom version of a wheel, use `SETUPTOOLS_SCM_PRETEND_VERSION` instead
- finally, update docs with all of the above changes

## Stuff that works

- `pip wheel .` (and `python -m build --wheel`) actually builds an installable wheel. Note that the latter needs the `build` package to work properly (but this should only be ran in the CI anyway, most developers should use `pip install --config-setting="build-dir=_build" -e .` or similar for development and testing).

## Stuff that needs work

- [x] figure out how to use a single source of truth for the version, and where to put it; `setuptools-scm` seems to be able to do it without any file (see [here](https://setuptools-scm.readthedocs.io/en/latest/usage/)), which is nice, but is Python-specific. Maybe [this project](https://github.com/LecrisUT/CMakeExtraUtils)? (**EDIT**: as a workaround, we can set the env variable `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_NMODL="$(git describe --tags | cut -d '-' -f 1,2 | tr - .)"` to mimic the current behavior)
- [x] the keys `build-system.requires` and `project.dependencies` have some duplication: figure out which ones are required when (build-time vs. run-time) (**EDIT**: separated the two into disjoint sets now, works as intended)
- [x] possibly related to the above: for some reason, most of the other `build-system.requires` dependencies are also built in `wheelhouse/` when running `pip wheel` (**EDIT**: it seems `pip wheel` can be replaced by `python -m build --wheel`, which doesn't have this issue **EDIT 2**: the `--no-deps` flag does exactly that so we don't need `build` at all)
- [x] the documentation used to be built by `setup.py`, figure out how to build it without burdening `setup.py` with it. `sphinx` comes with the `sphinx-build` CLI utility, but is a bit cumbersome to use (**EDIT**: the docs are a bit difficult to refactor since they mix the version and NMODL installation. **EDIT 2**: preliminary docs seem to work)
- [x] a superset of the documentation issue: where do we put all of the "other" (testing, etc.) dependencies? (**EDIT**: seems this should go somewhere like `project.optional-dependencies.test` and `project.optional-dependencies.docs`)
- [x] bundle various mod files and other package data when creating the wheel
- [x] figure out how to change the name of the package dynamically so we can have both NMODL and NMODL-nightly on PyPI. Note that [PEP 621](https://peps.python.org/pep-0621/), which defines the `pyproject.toml` spec, explicitly mentions: "A build back-end MUST raise an error if the metadata specifies the name in dynamic.", so we can't use something that respects that PEP. A possible solution is to have a script in the CI which edits `pyproject.toml` to change the name of the project to `nmodl-nightly` (**EDIT**: as a PoC workaround, added a `change_name.py` script to the `packaging/` subdirectory, that only runs in the CI, which can change the name of the project before we actually build it).
- [x] figure out whether editable installs are supported (which would allow incremental builds) (**EDIT**: seems that `pip install -e .` works, and `setuptools` has support for PEP 660) (**EDIT 2**: `scikit-build` [does not support](https://scikit-build.readthedocs.io/en/latest/index.html#known-issues) editable installs, but `scikit-build-core` does. Note that one should explicitly set `--config-setting="build-dir=[PATH]"` to re-use the build directory since build frontends delete them).
    - [x] ~~TODO: maybe pin the earliest version of `setuptools` which has support for PEP 660?~~ we don't need this if using `scikit-build-core`
- [x] for some reason, if `build-system.build-dir` is not specified on MacOS, the build consistently fails (reproducible both in the CI and locally), with the complaint from bison that one of the files cannot be found. Not sure if this is a huge problem (we can always override this by specifying the `SKBUILD_BUILD_DIR` env variable), but it's a bit annoying that it doesn't work OOTB (**EDIT**: this is possibly due to [this bug](https://gitlab.kitware.com/cmake/cmake/-/issues/20107) in cmake itself; as a workaround, if specifying a temporary dir as a build dir, one can use `realpath $(mktemp -d)` instead of just `mktemp -d`)
- [x] the cmake file apparently puts all of the `fmt` library-related stuff in the `install_manifest.txt`, which is impossible to ignore using either `sdist.exclude` or `wheel.exclude`, so, right now, `scikit-build-core` puts spurious `include` and `lib` (or `lib64` on Linux) directories in the wheel which cannot be removed (ticket: #1153) (**EDIT**: we can disable this using `set(FMT_INSTALL OFF)` in `CMakeLists.txt`. **EDIT 2**: c.f. https://github.com/fmtlib/fmt/pull/3264 maybe there's a more clever way of doing this?)
- [x] when it comes to editable installs, it would be nice if there was a way to install the `build-system.requires` dependencies using pip (since it vendors its own version of a TOML parser), but I am not very keen on being dependent on dubious enhancements to upstream. Of course, for Python 3.11+, it's possible to make a script which parses `pyproject.toml` natively that can then feed a list to either stdout or to a requirements.txt file (**EDIT**: this is now possible with a PoC script in `packaging/show_deps.py`, which uses the vendored `tomli` package in pip itself, or the native `tomllib` if running on Python 3.11. Note that there's also [pip-tools](https://pypi.org/project/pip-tools/), but I haven't found a way for it to output the `build-system.requires` dependencies. **EDIT 2**: once [this commit](https://github.com/jazzband/pip-tools/commit/bb289c49f19408c9afa633537cac6df284d11da7) gets into a release, then we should be able to use `--all-build-deps` to get those as well. **EDIT 3**: `pip-tools` 7.4.0 and above can now use `pip-compile --all-build-deps` to get the build-time dependencies. This can also be used with the `--all-extras` to get _all_ dependencies installed)
    - to clarify the above, to install _all_ of the dependencies using `pip-tools`, one must run `pip install 'pip-tools>=7.4.0'` followed by `pip install -r <(pip-compile --all-build-deps --all-extras --no-strip-extras 2>&1)`
- [ ] figure out whether pybind can be removed as an external dependency that we keep track of here (see https://pybind11.readthedocs.io/en/stable/compiling.html#pep-518-requirements-pip-10-required)
- [x] maybe fix Spack recipe as the gitlab pipeline is now failing (**EDIT**: after https://github.com/BlueBrain/spack/pull/2320 gets merged we should be good to go)


docs/generate_docs.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# script for generating documentation for NMODL
+# note that the NMODL Python wheel must be installed
+# for the script to work properly
+
+set -xe
```
1uc, line 7:
```suggestion
set -xeu
```

INSTALL.rst
```diff
@@ -219,9 +245,12 @@ In order to build the documentation you must have additionally
...
 
-You can build the entire documentation simply by using sphinx from
-``setup.py``:
+You can build the entire documentation simply by using the ``generate_docs.sh``
+script:
 
 .. code:: sh
 
-   python3 setup.py build_ext --inplace docs
+   bash docs/generate_docs.sh
```
1uc, line 253: This is incorrect. I cloned the repo and ran the command:
```
$ bash docs/generate_docs.sh
+++ dirname docs/generate_docs.sh
++ realpath docs
+ script_dir=/home/lucg/sandbox/nmodl/pr-1147/nmodl/docs
+ cd /home/lucg/sandbox/nmodl/pr-1147/nmodl/docs
+ doxygen Doxyfile
error: configuration file Doxyfile not found!
```

INSTALL.rst
```diff
@@ -74,11 +75,13 @@ installed along with the system toolchain:
...
 
-The Python dependencies are installed using:
+All of the Python dependencies (build, run, and development) can be installed
+using:
 
 .. code:: sh
 
-   pip3 install --user -r requirements.txt
+   pip3 install --user pip-tools
+   pip3 install --user -r <(pip-compile --all-build-deps --all-extras --no-strip-extras 2>&1)
```
1uc, line 84: This looks non-standard. The problem seems to be that we're not using a `requirements.txt` anymore. Probably because the dependencies are specified elsewhere.

Would it be worth automatically generating `requirements.txt` and add it to the Git repo? One then adds CI to check that the file is up to date. After doing so, installing the dependencies becomes rather standard:
```
pip install -r requirements.txt
```
(assuming a venv is active).

INSTALL.rst
JCGoran: Ah, good catch! Apparently we use CMake to generate it:
https://github.com/BlueBrain/nmodl/blob/25eb4a055fee4ba7a2e77a621b36834f9d028312/CMakeLists.txt#L194
so I guess we can just use `Doxygen.in` instead of `Doxygen`, and get rid of the useless invocation in `CMakeLists.txt`? Note that the CI doesn't pick this up because `pip wheel . --no-deps` runs CMake

1uc: I think lines like this:
https://github.com/BlueBrain/nmodl/blob/25eb4a055fee4ba7a2e77a621b36834f9d028312/docs/Doxyfile.in#L867

are why you can't just use `Doxygen.in`. I was expecting this to also configure the version for Doxygen.

INSTALL.rst
JCGoran: All of the requirements are in `pyproject.toml` now, either as build-time, run-time, or optional (testing and docs for now, we can also add development-centric stuff like linters or type checkers in another category for instance).
Note that `pip-compile --all-build-deps --all-extras --no-strip-extras` actually generates a `requirements.txt`, but it's a very lenghty one (has all of the transitive dependencies I think).
On the other hand, `pip` has no way of actually _showing_ the dependencies/generating a `requirements.txt` from the TOML file, which is IMO a design issue on their end.
I've contemplated rolling our own script which just uses `pip`s internally vendored TOML reader to read the requirements, but it seems a bit convoluted for what should've been basic built-in functionality.

1uc: Side note: It took 30s or more to run. It also creates a `requirements.txt` with precise version numbers.

JCGoran: Yes, it pins everything and there isn't a way to remove this behavior. I don't really know of a good way to solve this except rolling our own parser, none of the standard tools seem to have a way of just parsing the TOML file and outputting the various dependencies in a `requirements.txt` format.

INSTALL.rst
JCGoran: Ah, so in order to document the project, we first need to build it? Should the documentation then run `pip wheel . --no-deps` to get around this? Nothing else really comes to mind...

1uc: Try not running `doxygen` manually. It should be taken care of here:
```
https://github.com/BlueBrain/nmodl/blob/fbf9de61d02b252a1c8da54ef3c7e98175920cdf/docs/conf.py#L36
```

JCGoran: Now `bash docs/generate_docs.sh` should work more or less without issues, and you can also specify a build directory (so you can reuse it) using `bash docs/generate_docs.sh DIRECTORY` (it defaults to a temporary dir).

1uc: It doesn't work on my machine:
```
+ sphinx-build . /home/lucg/sandbox/nmodl/pr-1147/nmodl/docs/../public
Running Sphinx v7.2.6

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/sphinx/config.py", line 358, in eval_config_file
    exec(code, namespace)  # NoQA: S102
    ^^^^^^^^^^^^^^^^^^^^^
  File "/home/lucg/sandbox/nmodl/pr-1147/nmodl/docs/conf.py", line 21, in <module>
    import nmodl  # isort:skip
    ^^^^^^^^^^^^
ModuleNotFoundError: No module named 'nmodl'
```

JCGoran: Ahh sorry, I forgot about the state of my venv so it worked locally... For some reason we import NMODL in `conf.py`, so we need to first build the module. I think we can just mimic the behavior of `packaging/test_wheels.bash` and install it in a separate venv by default (with the option to do it in the current env), then build the docs using the module just installed.

docs/generate_docs.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# in order to create the docs, we first need to build NMODL in some dir
+build_dir="${1:-"$(mktemp -d)"}"
+if ! [ -d "${build_dir}" ]
+then
+    mkdir -p "${build_dir}"
+fi
+
+# the Python executable to use (default: whatever `python3` is)
+python_exe="${2:-"$(command -v python3)"}"
```
1uc, line 15: Ignored, because it's not used before it's overwritten.

docs/generate_docs.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# should we use a separate venv?
+use_venv="${3:-}"
+
+if [ "${use_venv}" != "false" ]
+then
+    venv_name="$(mktemp -d)"
+    $(command -v python3) -m venv "${venv_name}"
```
1uc, line 23: Ignores the choice made in `${python_exe}`.

docs/generate_docs.sh
```diff
@@ -0,0 +1,43 @@
...
+    . "${venv_name}/bin/activate"
+    python_exe="$(command -v python)"
+    ${python_exe} -m pip install -U pip
+else
+    python_exe="$(command -v python3)"
+fi
+${python_exe} -m pip install ".[docs]" -C build-dir="$(cd "${build_dir}"; pwd -P)"
```
1uc, line 30: How are the following different?
```
$(cd "${build_dir}"; pwd -P)
$(realpath "${build_dir}")
```

docs/generate_docs.sh
```diff
@@ -0,0 +1,43 @@
...
+script_dir="$(cd "$(dirname "$0")"; pwd -P)"
+
+cd "${script_dir}"
+doxygen Doxyfile
+sphinx-build . "${script_dir}/../public"
+cd -
+
+if [ "${use_venv}" != "false" ]
+then
+    deactivate
```
1uc, line 42: We leave the temporary directory in place, creating a new one each time we build the docs. If it's temporary, maybe we should delete it?

docs/generate_docs.sh
JCGoran: `realpath` isn't available on some older versions of MacOS (see [here](https://apple.stackexchange.com/a/450116)), so the first one should work on both Linux and any version of MacOS (assuming the dir has the right permissions, which, for a temp one, should always be the case).

nmodl/__init__.py
```diff
@@ -34,7 +36,12 @@
 try:
     # Try importing but catch exception in case bindings are not available
     from ._nmodl import NmodlDriver, to_json, to_nmodl  # noqa
-    from ._nmodl import __version__
+
+    try:
+        __version__ = version("nmodl")
+    except PackageNotFoundError:
+        # package is not installed
+        pass
```
1uc, line 44: I don't follow, why is it okay, that the package is not installed?

packaging/change_name.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+"""
+Barebones utility for changing the name of the package in `pyproject.toml`.
+
+Usage
+-----
+[SCRIPT] [PYPROJECT_FILE] [NAME]
```
1uc, line 7: The script can print its own, accurate, help message.

packaging/change_name.py
```diff
@@ -0,0 +1,53 @@
...
+
+Notes
+-----
+Should only be used by the CI
```
1uc, line 11: This isn't helpful to because I don't know if it's telling me that it'll only work during CI; or it's just a hint that it's mostly used during CI. This can be folded into the first sentence.

docs/generate_docs.sh
JCGoran: You mean the `venv` dir? I prefer to leave the cleanup of those up to the OS, no need to run `rm -fr` unless absolutely necessary (while I quote my variables judiciously, I'm always paranoid of something like [this](https://github.com/MrMEEE/bumblebee-Old-and-abbandoned/issues/123) happening).

src/CMakeLists.txt
```diff
@@ -49,4 +49,6 @@ endif()
 # Install executable
 # =============================================================================
 install(TARGETS nmodl DESTINATION ${NMODL_INSTALL_DIR_SUFFIX}bin)
-install(FILES nmodl.hpp DESTINATION ${NMODL_INSTALL_DIR_SUFFIX}include)
+if(NOT NMODL_BUILD_WHEEL)
+  install(FILES nmodl.hpp DESTINATION ${NMODL_INSTALL_DIR_SUFFIX}include)
```
1uc, line 53: Why are we only installing `nmodl.hpp` when we're not building a wheel?

JCGoran: What's the usecase for distributing them in the wheel (genuinely asking)?

packaging/change_name.py
JCGoran: Good spot, I was writing this in an ad-hoc fashion and didn't bother changing it once it worked.

src/CMakeLists.txt
1uc: I wish `h5py` would distribute the headers for the HDF5 library they install; if they would, I could build python packages that link against the version of HDF5 they use; because they don't I would need to figure out which the correct version is, guess how to build it; and hope I end up with compatible headers. Then I could try to find their `libhdf5.so` and build a package that links to it.

Since HDF5 is a library with a lot of global state, if two applications don't use the same library they can't talk to each other, e.g. pass the integer identifier of a dataset from one library to the next.

For NMODL, I'd just like less special cases. If you don't want headers in the wheel you can strip them at a later stage. Much like the manifest dictates what goes into the source distribution.

JCGoran: I mean, the user still then needs to get the paths right, as the wheel is self-contained and doesn't put anything in `/usr/[local]/include`, `pip install` basically just extracts the files from the wheel into `site-packages`.

> If you don't want headers in the wheel you can strip them at a later stage.

Not very easily (i.e. I would need to remove stuff from the already-built wheel) as `scikit-build-core` doesn't have an option to change the `install_manifest.txt`, or exclude any files from it (also the reason why I use the same flag to disable `fmtlib` from installing itself).

All of the above being said, I can leave this header in as its default install location is fine (unlike `fmtlib`).

docs/generate_docs.sh
1uc: Okay, then remove the `mktemp` and require users to pass in a valid build directory. In CI you can use `$(mktemp -d)` or something else.

The reasons is that starting from nothing each time is unacceptably slow (multiple minutes). Meaning the default should reuse an existing build directory. (Or create one the first time it runs.)

Also it doesn't tell where to find the docs:
```
generating indices... genindex py-modindex done
copying notebooks ... [100%] notebooks/nmodl-sympy-solver-sparse.ipynb                                                                                        
highlighting module code... [100%] nmodl.ode                                                                                                                  
writing additional pages... search done
copying images... [100%] images/ion_storage-opt.svg                                                                                                           
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded, 13 warnings.

The HTML pages are in ../public.
+ cd -
/home/lucg/sandbox/nmodl/pr-1147/nmodl
+ '[' '' '!=' false ']'
+ deactivate
+ '[' -n /usr/local/sbin:/usr/local/bin:/usr/bin:/opt/cuda/bin:/opt/cuda/nsight_compute:/opt/cuda/nsight_systems/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/lucg/.local/bin:/home/lucg/third_party/gmsh-4.8.3/bin:/opt/Xilinx/Vivado/2020.2/bin:/home/lucg/.local/share/gem/ruby/2.7.0/bin ']'
+ PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/opt/cuda/bin:/opt/cuda/nsight_compute:/opt/cuda/nsight_systems/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/lucg/.local/bin:/home/lucg/third_party/gmsh-4.8.3/bin:/opt/Xilinx/Vivado/2020.2/bin:/home/lucg/.local/share/gem/ruby/2.7.0/bin
+ export PATH
+ unset _OLD_VIRTUAL_PATH
+ '[' -n '' ']'
+ '[' -n /usr/bin/bash -o -n '' ']'
+ hash -r
+ '[' -n '' ']'
+ unset VIRTUAL_ENV
+ unset VIRTUAL_ENV_PROMPT
+ '[' '!' '' = nondestructive ']'
+ unset -f deactivate
```
are the last couple of times of output. The location: `../public` is definitely wrong.

1uc: I'm not sure why it's very careful about not choosing a build directory, but boldly claims `public` as it's own. One should also check that if there already content in `public` from a prior build that it'll be replaced correctly. A typical issue is when a file is removed from the documentation, it might still be present in `public`.

Would it make sense to move both the build dir and the directory `public` into the output directory which one passes to `generate_docs.sh`? That way if I want to delete everything I can simply `rm -r OUTPUT_DIR` and all generated content is gone, including the `venv`.

src/CMakeLists.txt
1uc: That's unfortunate, maybe it's the lesser evil. I was hoping we could completely eliminate the need for a special wheel mode.

packaging/change_name.py
JCGoran: Removed the comment because we can use the script wherever we want, it's just that it's kinda pointless to use it outside of the CI.

docs/generate_docs.sh
JCGoran: Updated, docs now build in `[DIRECTORY]/docs`, where `[DIRECTORY]` is whatever dir you want, while the temporary build files are in `[DIRECTORY]/build` (see updated docs for usage).
On my machine, if re-building the docs, `bash docs/generate_docs.sh mydocs python false` runs in < 10 seconds.

INSTALL.rst
```diff
@@ -219,9 +245,16 @@ In order to build the documentation you must have additionally
...
 
-You can build the entire documentation simply by using sphinx from
-``setup.py``:
+You can build the entire documentation simply by using the ``generate_docs.sh``
+script:
 
 .. code:: sh
 
-   python3 setup.py build_ext --inplace docs
+   bash docs/generate_docs.sh DIRECTORY PYTHON_EXECUTABLE [use_virtual_env]
```
1uc, line 253: Please make it:
```
bash docs/generate_docs.sh DIRECTORY [PYTHON_EXECUTABLE] [use_virtual_env]
```
Hardly ever is `python3`, i.e. the one in the `PATH` not the one wants.

docs/generate_docs.sh
```diff
@@ -0,0 +1,51 @@
...
+
+build_dir="build"
+docs_dir="docs"
+
+echo "== Building documentation files in: ${output_dir}/${docs_dir} =="
+echo "== Temporary project build directory is: ${output_dir}/${build_dir} =="
+
+if [ "${use_venv}" != "false" ]
+then
+    venv_name="$(mktemp -d)"
```
1uc, line 32: Does implies you need to reinstall the `venv` every time even if you're reusing the directory.

docs/generate_docs.sh
1uc: Doesn't work:
```
$ docs/generate_docs.sh build-docs python3
/home/lucg/sandbox/nmodl/pr-1147/nmodl
Running Sphinx v5.3.0
[NMODL] [warning] :: Python bindings are not available

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/tmp/tmp.2no3mfNQJo/lib/python3.11/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/home/lucg/sandbox/nmodl/pr-1147/nmodl/docs/conf.py", line 83, in <module>
    version = nmodl.__version__
              ^^^^^^^^^^^^^^^^^
AttributeError: module 'nmodl' has no attribute '__version__'
```

src/CMakeLists.txt
1uc: After merging #1193, I think we no longer need special cases for what's installed when. The tree is now:
```
build-debug/install
├── bin
│   ├── lexer
│   │   ├── c_lexer
│   │   ├── nmodl_lexer
│   │   └── units_lexer
│   ├── nmodl
│   ├── parser
│   │   ├── c_parser
│   │   ├── nmodl_parser
│   │   └── units_parser
│   └── visitor
│       └── nmodl_visitor
├── include
│   ├── crout
│   │   └── crout.hpp
│   ├── Eigen
...
│   ├── newton
│   │   └── newton.hpp
│   └── nmodl.hpp
├── lib
│   └── nmodl
│       ├── ast.py
│       ├── dsl.py
│       ├── ext
│       │   ├── example
│       │   │   ├── exp2syn.mod
│       │   │   ├── expsyn.mod
│       │   │   ├── hh.mod
│       │   │   └── passive.mod
│       │   └── viz
│       │       ├── css
│       │       │   └── tree.css
│       │       ├── index.html
│       │       └── js
│       │           ├── d3.min.js
│       │           └── tree.js
│       ├── __init__.py
│       ├── _nmodl.cpython-311-x86_64-linux-gnu.so
│       ├── ode.py
│       ├── symtab.py
│       └── visitor.py
└── share
    └── nmodl
        └── nrnunits.lib
```

JCGoran: Do we need these two parts in the wheel? They get copied in `nmodl/.data/include` if I don't disable it with the `NMODL_BUILD_WHEEL` flag:

https://github.com/BlueBrain/nmodl/blob/85a4906f3458f29843ba57381a946aa4fccefdb1/src/solver/CMakeLists.txt#L12
https://github.com/BlueBrain/nmodl/blob/85a4906f3458f29843ba57381a946aa4fccefdb1/src/solver/CMakeLists.txt#L18

1uc: With the right MOD file I can generate code that looks like this:
```
#include <coreneuron/utils/randoms/nrnran123.h>
#include <newton/newton.hpp>


namespace coreneuron {
```
The file `<newton/newton.hpp> is supposed to be provided by NMODL which I think is why you'll find it in the tree. It in turn uses Eigen:
```
#include <crout/crout.hpp>

#include <Eigen/Dense>
#include <Eigen/LU>
```

Therefore when `nrnivmodl` does (the equivalent of)
```
g++ -c kinetic01.cpp -o kinetic01.o
```
it needs to know where to find `<newton/newton.hpp>` and Eigen. The way this is solved is by installing the required files. Now `nrnivmodl` can do:
```
g++ -I ${NMODL_INSTALL_DIR}/include -c kinetic01.cpp -o kinetic01.o
```

Since NMODL generates the calls to the functions inside of `newton.hpp` it seems to convenient to include `newton.hpp` in NMODL rather than ship it with NEURON. This way changing how we generate code that needs to solve a non-linear system is contained inside NMODL. (This isn't true if one uses `scopmath/newton_thread.hpp`; the common case.)

1uc: When checking the output of `nrnivmodl` we find:
```
/usr/bin/c++ -g  -O0 -fstack-protector   -fopenmp-simd -std=c++17 -DUSE_PYTHON -DNRN_ENABLE_THREADS -DMPI_NO_CPPBIND=1 -DOMPI_SKIP_MPICXX=1 -DMPICH_SKIP_MPICXX=1 -DHAVE_CONFIG_H  -I".." -I.   -I/home/lucg/git/bbp/nrn/build-nmodl/install/include  -I/usr/include -fPIC -c "/home/lucg/git/bbp/nmodl_benchmark/kinetic/x86_64/kinetic01.cpp" -o kinetic01.o
```
Meaning it doesn't in the slightest care where NMODL installs the sources, it'll always use the headers (and version of Eigen) that was installed with NEURON.

.github/workflows/nmodl-doc.yml
```diff
@@ -58,46 +58,31 @@ jobs:
...
-          pip3 install -U pip setuptools
-          pip3 install --user -r requirements.txt
-
-      # This step will set up an SSH connection on tmate.io for live debugging.
-      # To trigger it, simply add 'live-debug-docs' to your last pushed commit message.
-      - name: live debug session on failure
-        if: failure() && contains(github.event.head_commit.message, 'live-debug-docs')
-        uses: mxschmitt/action-tmate@v3
+          pip3 install -U pip setuptools 'pip-tools>=7.4.0'
+          pip3 install --user -r <(pip-compile --all-build-deps --all-extras --no-strip-extras 2>&1)
```
1uc, line 62: Isn't installing these dependencies part of `generate_docs.sh`?

JCGoran: Indeed.

docs/generate_docs.sh
1uc: It seems to build docs and after installing via `pip install .` it seems like it may have created a functional module (import and printing `nmodl.__version__` don't fail).

python/nmodl/__init__.py
```diff
@@ -3,8 +3,10 @@
 
 if sys.version_info >= (3, 9):
     from importlib.resources import files
+    from importlib.metadata import version, PackageNotFoundError
 else:
     from importlib_resources import files
+    from importlib_metadata import version, PackageNotFoundError
```
1uc, line 9:
```suggestion
    from importlib_metadata import version
```
would this work?

python/nmodl/__init__.py
```diff
@@ -33,8 +35,7 @@
 
 try:
     # Try importing but catch exception in case bindings are not available
-    from ._nmodl import NmodlDriver, to_json, to_nmodl  # noqa
-    from ._nmodl import __version__
+    from ._nmodl import NmodlDriver, to_json, to_nmodl, __version__  # noqa
```
1uc, line 38: Why is this happening in a `try` block? What's the difference between a Python package `nmodl` with python binding and without? Could we require the "bindings" to always be present and simply error out?

python/nmodl/__init__.py
JCGoran: `importlib_metadata` stuff is now irrelevant in the init because nothing uses it anymore, my linter just didn't report it so I forgot to remove it.

CMakeLists.txt
```diff
@@ -168,9 +174,12 @@ find_package(PythonInterp 3.8 REQUIRED)
...
 include(cmake/hpc-coding-conventions/cpp/cmake/bbp-find-python-module.cmake)
 cpp_cc_find_python_module(jinja2 2.9.3 REQUIRED)
-cpp_cc_find_python_module(pytest 3.3.0 REQUIRED)
-cpp_cc_find_python_module(sympy 1.3 REQUIRED)
-cpp_cc_find_python_module(textwrap 0.9 REQUIRED)
+if(NMODL_ENABLE_TESTS)
+  cpp_cc_find_python_module(pytest 3.3.0 REQUIRED)
+endif()
+if(NOT NMODL_BUILD_WHEEL)
+  cpp_cc_find_python_module(sympy 1.3 REQUIRED)
```
1uc, line 181: I guess they're split into build, run and test dependencies? If yes shuffle and please add a comment.

Alternatively, should we be checking these things in a CMakeLists.txt or can we all checks?

python/nmodl/__init__.py
JCGoran: Apparently removing it breaks this CI: https://github.com/BlueBrain/nmodl/actions/runs/8138354896/job/22238986595?pr=1147

INSTALL.rst
```diff
@@ -45,12 +45,13 @@ of all dependencies we recommend using `homebrew <https://brew.sh/>`__:
...
-The necessary Python packages can then easily be added using the pip3
-command.
+All of the Python dependencies (build, run, and development) can be installed
+using:
 
 .. code:: sh
 
-   pip3 install --user -r requirements.txt
+   pip3 install --user pip-tools
+   pip3 install --user -r <(pip-compile --all-build-deps --all-extras --no-strip-extras 2>&1)
```
1uc, line 54: The question is if it's worth duplicating some information to avoid these instructions and instead have the standard, easily discoverable:
```
pip install --user -r requirements.txt
```

1uc:
I get the following version information for this branch:
```
$ pip install .
$ python -c "import nmodl; print(nmodl.__version__)"
0.6
```
is that intended?

python/nmodl/__init__.py
1uc: Okay, might be a bug in the tests.

JCGoran:
> I get the following version information for this branch:
> 
> ```
> $ pip install .
> $ python -c "import nmodl; print(nmodl.__version__)"
> 0.6
> ```
> 
> is that intended?

We set `__version__` in Python from whatever CMake says the version is, I haven't touched that; if this is not intended, I think it should go in a separate issue/PR.

INSTALL.rst
JCGoran: Yeah it's a bit annoying, I've put back requirements now.

CMakeLists.txt
JCGoran: CMake doesn't make the distinction between build, run, or other dependencies, so I would be in favor of removing all of the checks here since `pip install -r requirements.txt` installs all of them anyway.

python/nmodl/__init__.py
JCGoran: I've reverted back the commit; if it's a bug in the tests, I think it's best we investigate in a separate issue/PR.

CMakeLists.txt
JCGoran: Removed discovery of all Python packages in CMakeLists.txt now so we don't need to keep track of the info in two places.

packaging/build_wheels.bash
```diff
@@ -47,14 +47,13 @@ build_wheel_linux() {
...
-    pip install -r packaging/build_requirements.txt
+    pip install pip auditwheel
 
     echo " - Building..."
-    rm -rf dist _skbuild
+    rm -rf dist _build
     # Workaround for https://github.com/pypa/manylinux/issues/1309
     git config --global --add safe.directory "*"
-    python setup.py bdist_wheel
+    python -m pip wheel . --wheel-dir dist/ --no-deps
```
1uc, line 56: Might be missing the `-C build-dir=_build` which we deleted a few lines prior.

JCGoran: On Linux the issue is not present because the temporary dir is not a symlink (unlike on MacOS, see [this ticket](https://gitlab.kitware.com/cmake/cmake/-/issues/20107)), so it's safe to build it in the temp dir (default if none is specified). If for some reason you need to build a wheel to test locally, you can build it with a persistent dir with just `python -m pip wheel . --wheel-dir dist/ --no-deps -C build-dir=somedir` (the `build_wheels.bash` script is used only in the CI anyway, and will be removed once #1155 is merged).

1uc: Okay, doesn't answer the question why we're deleting `_build` two lines further up.

JCGoran: Guess I forgot to delete that line after copy-pasting from the MacOS one, it doesn't really do anything.

1uc:
Let's summarize: 

* It installs on two desktops/laptops (Mac and Linux) and CI.
* Docs build on the same two machines and CI.
* There's the surprise that tests use a partial installation of NMODL; and we don't know if it must be like that because in reality NMODL actual is two different Python modules; or if NMODL is one, regular Python module but there's a bug/shortcut in the tests leading to the strange behaviour.
* Certain things were postponed, have the corresponding issues been created?
* Does it build on BB5?
* IIUC the spack recipe has been updated, does it still work?

JCGoran:
To answer your questions:

> Certain things were postponed, have the corresponding issues been created?

For the version issue, see #1199.
Regarding the test failing if we explicitly load the Python module, I don't think it's an issue with NMODL, but rather I with the compiler flags: we try to use the address sanitizer library, but cannot load it dynamically for some reason:

```plaintext
ImportError while importing test module '/home/runner/work/nmodl/nmodl/test/unit/ode/test_ode.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
../../../test/unit/ode/test_ode.py:6: in <module>
    from nmodl.ode import differentiate2c, integrate2c
../../lib/nmodl/__init__.py:34: in <module>
    from ._nmodl import NmodlDriver, to_json, to_nmodl, __version__  # noqa
E   ImportError: libclang_rt.ubsan_standalone-x86_64.so: cannot open shared object file: No such file or directory
```

We don't ship asan in the distributed wheels, so I don't think it's a big issue. Also, the failing test (ODE) is actually tested in the CI as part of the Pytest suite.

> Does it build on BB5?

Yes.

> IIUC the spack recipe has been updated, does it still work?

Ongoing discussion at https://github.com/BlueBrain/spack/pull/2332, but I have no strong opinion whether we should still support Python 3.8 or not.

1uc:
> whether we should still support Python 3.8 or not.

BlueBrain/spack (as opposed to spack/spack) is my and large a bespoke version of Spack for BB5 (and maybe some EPFL controlled laptops). Therefore, there's a reasonably chance that avoiding the issue now mean never encountering it. When someone does need Python 3.8, they'll first be sent to the archive of module and if that doesn't work one can put in the time to support 3.8.

W.r.t the PR itself: nice work! Unless there's reasons to do otherwise, I propose we wait until the Spack PR is ready, then we can merge them shortly after each other. That way we retain the ability to release a version of NMODL at short notice by simply tagging `master` (followed by a trivial update of the spack recipe).

1uc (approved):

PR #1151, opened 2024-02-07
Start documenting pointers.

1uc:


1uc:
Probably not complete, and might become part of some other section; but I'll stop here. Hence, I propose reviewing, improving, merging and improving again when my understanding improves.

pramodk (approved):
Excellent! thanks for starting this @1uc !

PR #1154, opened 2024-02-13
Update dependencies to last versions

alkino:


pramodk:
Did you check why sympy solver test is failing? Does this locally fail as well? (we don't change Sympy here and hence a bit confusing?)

1uc:
Implemented via different PR.

PR #1155, opened 2024-02-15
Add building wheels using cibuildwheel

JCGoran:
This is a follow-up of #1147 to greatly simplify and speed-up the whole build CI pipeline.

## Overview of changes

* building a (redistributable!) wheel for a platform is now as simple as `pip install cibuildwheel && cibuildwheel`. The resulting wheel will be available in the `wheelhouse` subdirectory. Some notes about this:
    * by default, it builds wheels for all Python versions supported by NMODL (currently 3.8-3.12); for quick testing and development, set the env variable `CIBW_BUILD` to something like `cp311*` (for Python 3.11) to only build a wheel for a specific Python version
    * the wheel built is native to the platform (on Linux, or if specifying `--platform linux` in the above, Docker is needed; if you have podman instead, set `CIBW_CONTAINER_ENGINE=podman`)
    * on MacOS, you need the installer from python.org (the Homebrew-installed one won't cut it)
    * on MacOS, we also run some tests once the wheel is built
    * on Linux, we only run tests in the CI (why? Because the `manylinux` Docker image used for building the container does not come with the Python shared library); if you want to test the built wheel locally on Linux, run `bash packaging/test_wheel.bash $(command -v python3) WHEEL`, where `WHEEL` is the wheel you just built
* as a result of all of the above, removed `packaging/build_wheels.bash`
* simplified CI to use `cibuildwheel` everywhere. This comes with a big reduction in the CI build time (see for instance [old](https://dev.azure.com/pramodskumbhar/nmodl/_build/results?buildId=6211&view=results) vs. [new](https://dev.azure.com/pramodskumbhar/nmodl/_build/results?buildId=6214&view=results), 16 minutes per Python version vs. 7 minutes). By default we build wheels for all Python versions, which is probably not optimal, and I don't have strong opinions on this, so whatever the popular vote decides is fine by me (build all/build specific version/build oldest/build newest/build oldest _and_ newest)
* so far ARM64 builds work without issues, it's just a matter of setting up the proper CI pipeline (Azure doesn't have ARM64 MacOS runners, and GitHub doesn't have ARM64 Linux runners; for the former we can use GitHub, and for the latter we can use CircleCI), which, following some internal discussions, I am leaving out of this PR
* update docs

## What works

- manylinux on x86_64 and ARM
- MacOS on x86_64 and ARM

# What needs work

- [x] after a bunch of Windows-specific fixes, jinja on Windows still has some issues with discovery of files, see https://github.com/BlueBrain/nmodl/actions/runs/7915828954/job/21608472715 **EDIT**: leave Windows support for later
- [x] ~~build time on Linux per Python version can take a while (5 minutes) on x86_64, and, due to the need for QEMU, is even longer on aarch64. If we only want x86_64 builds though, we should be fine (**EDIT**: added CircleCI to build aarch64 Linux since it has a native runner)~~ **EDIT**: it now builds wheels much faster than `master`
- [x] similar to the above, but with Python 3.8 on ARM on MacOS (however, there is a [workaround](https://cibuildwheel.readthedocs.io/en/stable/faq/#macos-building-cpython-38-wheels-on-arm64) for that one available) **EDIT**: ARM64 wheels are left for later
- [x] testing of the wheels is not implemented yet (**EDIT**: done)
- [x] we should update and/or build our own custom Docker images that have flex and bison to speed-up the building process; the current custom x86_64 latest image is 2 years old
- [x] apparently on manylinux images, the `libpython.so` library is not available (as described [here](https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode) and in https://github.com/scikit-build/scikit-build-core/issues/173) so some workarounds may be needed for testing the `nmodl` binary in a portable way. The Python module tests work like a charm though!


.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
+version: 2
+
+jobs:
+  linux-aarch64-wheels:
+    working_directory: ~/linux-aarch64-wheels
+    machine:
+      image: ubuntu-2004:2022.04.1
```
alkino, line 7:
```suggestion
      image: default
```

.circleci/config.yml
```diff
@@ -0,0 +1,30 @@
...
+    # resource_class is what tells CircleCI to use an ARM worker for native arm builds
+    # https://circleci.com/product/features/resource-classes/
+    resource_class: arm.medium
+    steps:
+      - checkout
+      - run:
+          name: Build the Linux aarch64 wheels.
+          environment:
+            CIBW_SKIP: "pp* *-musllinux_aarch64"
+            CIBW_BEFORE_BUILD_LINUX: bash packaging/install_dependencies.sh
```
alkino, line 17:
```suggestion
            CIBW_BEFORE_ALL_LINUX: yum install -y flex-devel bison-devel
```

JCGoran: This was my initial idea, however, the version of flex (or bison?) in the repos for the `manylinux` image is too old for NMODL (disclaimer: I don't know why we require those specific minimum versions in the first place), so we need to install them from source.

alkino:
```suggestion
            CIBW_BEFORE_ALL_LINUX: bash packaging/install_dependencies.sh
```

JCGoran: Ahhh right, good catch! I moved everything to `pyproject.toml` to keep the CI clutter-free (options can still be overridden with env variables), so I'll put this there as well.

1uc (approved):

PR #1160, opened 2024-02-20
Remove `clang-format`ing of generated sources.

1uc:
NMODL should generate code that's sensible formatted.


ohm314:
So, here's the thing with this PR. I remember specifically there has been already plenty of back-and-forth about the question of formatting generated code. I feel like about once every year someone decides that it needs to be done differently. I _thought_ with Tristan we had said that "no, we will not care too much about directly generating correctly formatted code, it should be of course readable, but we leave consistent formatting up to a tool that knows well how to do this, hence `clang-format`. So now I wonder, why should we change this policy?
I'm not fundamentally against, but I want to avoid us to do changes that are unnecessary.

PR #1161, opened 2024-02-21
Myers diff algorithm for unit tests

ohm314:
This is a small new feature for a bit of convenience. This PR adds a proper diff implementation that can be invoked when a unit test fails. Instead of outputting the the raw generated strings and painstakingly comparing them, we can now simply diff them and print formatted output.

For now just one test case was updated as an example for initial feedback.


test/unit/visitor/loop_unroll.cpp
```diff
@@ -88,7 +89,16 @@ SCENARIO("Perform loop unrolling of FROM construct", "[visitor][unroll]") {
         )";
         THEN("Loop body gets correctly unrolled") {
             auto result = run_loop_unroll_visitor(input_nmodl);
-            REQUIRE(reindent_text(output_nmodl) == reindent_text(result));
+            if (reindent_text(output_nmodl) != reindent_text(result)) {
+              auto diff = MyersDiff(reindent_text(output_nmodl), reindent_text(result));
+              diff.do_diff();
```
1uc, line 94: This pattern rather reliably tells you that `MyersDiff` is either not a class or not initializing itself correctly. Probably, the easiest would be to just call `do_diff` from inside its ctor.

Then provide means for creating a string; either `MyersDiff::to_string()` or overload `operator<<`.

Finally, probably provide a function that combines the two so this can be at most just:
```
auto diff = myers_diff(output_nmodl, result);
if(diff) {
  FAIL(diff);
}
```

ohm314: OK, thanks. Yes, this all makes sense.
On the last point, though.. unfortunately the algorithm (or at least my implementation) currently still sometimes shows some stupid diffs, e.g. a line that is the same in `a` and `b` can sometimes show up as deleted and added.
This might be really just a bug that I need to address, but it could be also that some of those corner cases will remain, which would show as false-positives in the unit test.

1uc: The entire thing is complex, mostly C and doesn't have a single test. Personally, I would be unable to implement this algorithm even broadly correctly without quite thorough testing.

It's all rather a lot of work for a `subprocess.run(["diff", a, b])` (in tests we can assume we have access python).

ohm314:
I'll close this. As discussed, it's not very high priority and unless someone has interest and finds the time this can just stay in a branch.

PR #1165, opened 2024-02-22
Fix Python discovery issue when calling `sympy` solver from wheel

JCGoran:
(Splitting this up from #1147 as it was getting too large).

As an example (stolen from one of the notebooks):

```python
# example.py
import nmodl.dsl as nmodl


def run_conductance_visitor_and_return_breakpoint(mod_string):
    # parse NMDOL file (supplied as a string) into AST
    driver = nmodl.NmodlDriver()
    AST = driver.parse_string(mod_string)
    # run SymtabVisitor to generate Symbol Table
    nmodl.symtab.SymtabVisitor().visit_program(AST)
    # constant folding, inlining & local variable renaming passes
    nmodl.visitor.ConstantFolderVisitor().visit_program(AST)
    nmodl.visitor.InlineVisitor().visit_program(AST)
    nmodl.visitor.LocalVarRenameVisitor().visit_program(AST)
    # run CONDUCTANCE visitor
    nmodl.visitor.SympyConductanceVisitor().visit_program(AST)
    # return new BREAKPOINT block
    return nmodl.to_nmodl(
        nmodl.visitor.AstLookupVisitor().lookup(
            AST, nmodl.ast.AstNodeType.BREAKPOINT_BLOCK
        )[0]
    )


ex1 = """
NEURON {
    USEION na READ ena WRITE ina
    RANGE gna
}
BREAKPOINT {
    ina = gna*(v - ena)
}"""
print(run_conductance_visitor_and_return_breakpoint(ex1))
```

Running this with a freshly-installed wheel on latest master (built with `python setup.py bdist_wheel`) gives:

```sh
$ python example.py
[NMODL] [critical] :: NMODL_PYLIB environment variable must be set to load embedded python
Traceback (most recent call last):
  File "/private/var/folders/r_/1lcd8hvd6nzgtzcy8wwq9slx186djn/T/tmp.vzUItTIK/example.py", line 32, in <module>
    print(run_conductance_visitor_and_return_breakpoint(ex1))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/var/folders/r_/1lcd8hvd6nzgtzcy8wwq9slx186djn/T/tmp.vzUItTIK/example.py", line 15, in run_conductance_visitor_and_return_breakpoint
    nmodl.visitor.SympyConductanceVisitor().visit_program(AST)
RuntimeError: NMODL_PYLIB not set
```

Basically, due to the way loading of Python is handled, we should set the various env variables as soon as we load the module, otherwise the user may find mysterious errors.

Other changes:
- updated requirements and setup.py (for `importlib`)
- added a test for env variables (needs a `BREAKPOINT` statement to activate the sympy solver)
- added automatic setting of `NMODL_PYLIB` and `NMODLHOME` env variables (only if they are not already set!)


JCGoran:
Need https://github.com/BlueBrain/spack/pull/2320 to get merged, then the gitlab pipeline should be ✅

nmodl/__init__.py
```diff
@@ -1,7 +1,25 @@
...
+
+if sys.version_info >= (3, 9):
+    from importlib.resources import files
+else:
+    from importlib_resources import files
+
+from find_libpython import find_libpython
+
+# add libpython*.so path to environment
+os.environ["NMODL_PYLIB"] = find_libpython()
```
1uc, line 12: Is setting these unconditionally desired? Should we allow a use to pick the value of those variables?

requirements.txt
```diff
@@ -6,3 +6,4 @@ sympy
 numpy
 find_libpython
 scikit-build
+importlib_resources;python_version<"3.9"
```
1uc, line 9: Has something, e.g. CI, or someone, e.g. you, run this with Python 3.8?

test/unit/pybind/test_env_variables.py
```diff
@@ -0,0 +1,28 @@
...
+    driver = nmodl.NmodlDriver()
+    AST = driver.parse_string(mod_string)
+    # run SymtabVisitor to generate Symbol Table
+    nmodl.symtab.SymtabVisitor().visit_program(AST)
+    # constant folding, inlining & local variable renaming passes
+    nmodl.visitor.ConstantFolderVisitor().visit_program(AST)
+    nmodl.visitor.InlineVisitor().visit_program(AST)
+    nmodl.visitor.LocalVarRenameVisitor().visit_program(AST)
+    # run CONDUCTANCE visitor
+    nmodl.visitor.SympyConductanceVisitor().visit_program(AST)
```
1uc, line 28: Is this the first unit-test that runs NMODL as a module? If no why do the others not fail?

1uc (approved):
Looks very sensible to me. My comment are all optional, please feel free to ignore them, if you think they don't make sense.

requirements.txt
JCGoran: The CI builds it on 3.8 in multiple places, for instance:
https://github.com/BlueBrain/nmodl/blob/c0eb1ac07bcfcb80189170819c52e4d7645f7c72/azure-pipelines.yml#L205
https://github.com/BlueBrain/nmodl/blob/c0eb1ac07bcfcb80189170819c52e4d7645f7c72/.github/workflows/nmodl-ci.yml#L19

1uc: Perfect!

test/unit/pybind/test_env_variables.py
JCGoran: I'm not 100% sure, but I think this has to do with the `BREAKPOINT` statement, as the following code demonstrates (tested on latest master):

```python
# example.py
import nmodl
from nmodl.dsl import visitor


driver = nmodl.NmodlDriver()
bad_string = """
NEURON {
    USEION na READ ena WRITE ina
    RANGE gna
}
BREAKPOINT {
    ina = gna*(v - ena)
}"""

good_string = """NEURON  {
    SUFFIX NaTs2_t
    RANGE mInf, hInf
}

ASSIGNED {
    mInf
    hInf
}

STATE   {
    m
    h
}

DERIVATIVE states   {
    m' = mInf-m
    h' = hInf-h
}"""

visitor.SympyConductanceVisitor().visit_program(driver.parse_string(good_string))
# call order w.r.t. above doesn't matter, below will always raise an error
visitor.SympyConductanceVisitor().visit_program(driver.parse_string(bad_string))
```
This seems to give:
```plaintext
[NMODL] [critical] :: NMODL_PYLIB environment variable must be set to load embedded python
Traceback (most recent call last):
  File "example.py", line 37, in <module>
    visitor.SympyConductanceVisitor().visit_program(driver.parse_string(bad_string))
RuntimeError: NMODL_PYLIB not set
```

I am guessing we are only calling sympy (and hence the embedded interpreter) when the parser processes `BREAKPOINT`.

1uc: Thank you!

nmodl/__init__.py
JCGoran: `os.environ["NMODL_PYLIB"] = os.environ.get("NMODL_PYLIB", find_libpython())` would admittedly be more flexible. Note that `find_libpython` either returns a string or `None` instead of raising an or returning an empty string, so maybe wrapping the above in a `try..except TypeError` block and raising a `RuntimeError`?

1uc: Build system related bugs are horrible to debug. Therefore, yes, that might save someone some time.

JCGoran: Updated the env mechanism now

JCGoran:
Minor annoyance: if we merge this, we need to add [py-importlib-resources](https://github.com/BlueBrain/spack/blob/develop/var/spack/repos/builtin/packages/py-importlib-resources/package.py) to the NMODL Spack recipe.

1uc:
That doesn't seem wrong. The dependencies changed, hence the Spack recipe needs to change.

1uc (changes requested):

docs/contents/currents.rst
```diff
@@ -0,0 +1,61 @@
+Currents
```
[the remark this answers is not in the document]
JCGoran: Should be fixed now, merged `master` cleanly.

PR #1166, opened 2024-02-22
Remove deprecated `pkg_resources`

JCGoran:
The `pkg_resources` submodule of `setuptools` is deprecated, and `importlib` should be used instead (or its backports, `importlib_metadata` and `importlib_resources`).


1uc (approved):

pramodk (approved):
If the example from README still works then this is all good!

JCGoran:
@pramodk I've added unit tests for the example from the README and caught a few issues along the way, should be okay now.

test/unit/pybind/data/exp2syn.json
```diff
@@ -0,0 +1 @@
...: [{"Name": [{"String": [{"name": "tau2"}]}]}]}]}]}]}]}]}]}]}, {"NetReceiveBlock": [{"Argument": [{"Name": [{"String": [{"name": "weight"}]}]}, {"Unit": [{"String": [{"name": "uS"}]}]}]}, {"StatementBlock": [{"ExpressionStatement": [{"BinaryExpression": [{"VarName": [{"Name": [{"String": [{"name": "A"}]}]}]}, {"BinaryOperator": [{"name": "="}]}, {"WrappedExpression": [{"BinaryExpression": [{"VarName": [{"Name": [{"String": [{"name": "A"}]}]}]}, {"BinaryOperator": [{"name": "+"}]}, {"WrappedExpression": [{"BinaryExpression": [{"VarName": [{"Name": [{"String": [{"name": "weight"}]}]}]}, {"BinaryOperator": [{"name": "*"}]}, {"VarName": [{"Name": [{"String": [{"name": "factor"}]}]}]}]}]}]}]}]}]}, {"ExpressionStatement": [{"BinaryExpression": [{"VarName": [{"Name": [{"String": [{"name": "B"}]}]}]}, {"BinaryOperator": [{"name": "="}]}, {"WrappedExpression": [{"BinaryExpression": [{"VarName": [{"Name": [{"String": [{"name": "B"}]}]}]}, {"BinaryOperator": [{"name": "+"}]}, {"WrappedExpression": [{"BinaryExpression": [{"VarName": [{"Name": [{"String": [{"name": "weight"}]}]}]}, {"BinaryOperator": [{"name": "*"}]}, {"VarName": [{"Name": [{"String": [{"name": "factor"}]}]}]}]}]}]}]}]}]}]}]}]}
```
1uc, line 1: Rather pedantic: Since this is a single line, changing a single character will replace the whole line. Makes it harder for Git to function correctly; and harder for us to merge conflicts.

nmodl/dsl.py
```diff
@@ -39,6 +39,6 @@ def load_example(example):
     """
     path = files("nmodl") / RESOURCE_DIR / example
     if path.exists():
-        return str(path)
+        return path.read_text()
```
1uc, line 42: The docstring for "Returns:" seems to describe something different, i.e. return a path, not the contents of the file.

1uc (approved):

nmodl/dsl.py
JCGoran: Ah yes, I forgot to change the docstring in the last revision.
While we're on the subject of the return type, the docs for `pkg_resources.resource_string` say:

> Return the specified resource as bytes. The resource is read in binary fashion, such that the returned string contains exactly the bytes that are stored in the resource.

and while I can just replace the call with `read_bytes`, does it make any tangible difference for later parsing by the C++ module?

test/unit/pybind/data/exp2syn.json
JCGoran: I'm not a huge fan either, if we change the output format/API at some point, we'd need to change all of the test data as well...should I just remove the test data for now and only make sure the call to `driver.parse_string(nmodl_string)` succeeds (before it was failing because I erroneously returned the path to the file instead of its contents)?

nmodl/dsl.py
1uc: My guess is that pybind11 is responsible for converting string or byte to `std::string` correctly.

test/unit/pybind/data/exp2syn.json
JCGoran: The files are input for the test, but I generated them with NMODL itself, so if we change any part of the API of NMODL, we'd need to regenerate them.

1uc: Then the, pedantic, suggestion is to:
```
json.dump(f, json.load(f), index=2)
```
To have them formatted in a more human readable manner. This might help Git produce better diffs, which likely only matters if there's ever a conflict; and we want to check something, instead of just brutally regenerating and overwriting.

nmodl/dsl.py
JCGoran: Indeed, the reference says it will always use UTF-8: https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html
so I think it's a non-issue.

test/unit/pybind/data/exp2syn.json
JCGoran: The (possibly minor) problem is that the PR then adds about 8000 extra lines of code, so I am actually not that enthusiastic about including the files in the first place 😅 but if you think it makes sense I can add them. Note that the only functionality these new tests should bring is to try to parse the example mod files, testing the contents themselves of any mod files should probably be done in a separate test.

1uc: I don't think they're big enough to substantially slow down `git clone`, so it's not as much of an issue. OTOH, if you really don't want them in here, there's a subrepository dedicated to "reference" files, called `nmodl-references` (now I which it were `nmodl-data`).

We can move the location of the to make it work for both usecases.

JCGoran:
Minor annoyance (similar to #1165): if we merge this, we need to update the Spack recipe with [py-importlib-metadata](https://github.com/BlueBrain/spack/blob/develop/var/spack/repos/builtin/packages/py-importlib-metadata/package.py) and [py-importlib-resources](https://github.com/BlueBrain/spack/blob/develop/var/spack/repos/builtin/packages/py-importlib-resources/package.py).

test/unit/pybind/data/exp2syn.json
JCGoran: I've removed the actual comparison of the files for now (in the tests we just make sure we can read them), and opened up a separate issue for this https://github.com/BlueBrain/nmodl/issues/1187

PR #1170, opened 2024-02-23
Blame NMODL for lines it's generating.

1uc:
Adds

    nmodl MOD_FILE blame --line LINE

which will cause NMODL to print a backtrace every time it writes to line
`LINE`.

This is very helpful when trying to find the NMODL code responsible for
printing a (faulty) line of code.


.gitmodules
```diff
@@ -22,3 +22,9 @@
 [submodule "ext/catch2"]
 	path = ext/catch2
 	url = https://github.com/catchorg/Catch2.git
+[submodule "ext/backward-cpp"]
+	path = ext/backward-cpp
+	url = https://github.com/bombela/backward-cpp.git
+[submodule "ext/backward"]
+	path = ext/backward
+	url = https://github.com/bombela/backward-cpp.git
```
pramodk, line 30: twice ?

src/printer/CMakeLists.txt
```diff
@@ -3,4 +3,4 @@
 # =============================================================================
 add_library(printer OBJECT code_printer.cpp json_printer.cpp nmodl_printer.cpp)
 set_property(TARGET printer PROPERTY POSITION_INDEPENDENT_CODE ON)
-target_link_libraries(printer PRIVATE util)
+target_link_libraries(printer PRIVATE util Backward::Interface)
```
pramodk, line 6: I don't remember precisely: does this link (or assume) the presence of libraries like libunwind? (Thinking of portability of wheels and also portability on windows)

1uc: I think it'll detect those things and reduces performance gracefully. I guess CI will tell us.

.gitmodules
1uc: Yes, it took two attempts =) I've removed one.

src/printer/code_printer.cpp
```diff
@@ -110,5 +114,19 @@ void CodePrinter::pop_block(const std::string_view& suffix, std::size_t num_newl
...
+void CodePrinter::blame() {
+    if (current_line == blame_line) {
+        *result << std::flush;
+
+        std::cout << "\n\n== Blame =======================================================\n";
+
+        backward::StackTrace st;
+        st.load_here(32);
+        backward::Printer p;
+        p.print(st, std::cout);
```
ohm314, line 126: I'm not convinced we want to bake this in unconditionally... I would rather add a build option, and only if enabled we add backward and compile in the blame option. This seems to me like a feature that is mostly useful to developers and less to users (yes, you could argue that this way you can ask the user to just send you the "blame" when encountering an issue, but you can just as well easily reproduce this locally).

ohm314 (approved):
looks good now. thanks for the useful feature!

pramodk (approved):

PR #1131, opened 2023-12-20
Include compiled mod files.

1uc:
Add a script for generating the reference values of compiled mod files. These are included with the sources to enable the following:

  * during review the reviewer can see what was generated.
  * these can serve as golden tests, flagging changes in unrelated mod files.

PR #1172, opened 2024-02-23
Remove brittle unit-test.

1uc:
The removed test only checks if certain substrings are present. The golden testing automatically adds these tests for any usecase. Therefore, this test is not needed anymore.

The usecase for implementing non-specific current, will add a version of a passive current; and test the values against the analytical reference.

Ions are already covered by the `ionic` usecase.

Array variables are covered by the `cnexp_array` usecase.

Parameters coved by the `parameters` usecase.

Simple ODEs are covered by `cnexp_*` usecases.


1uc:
While developing NRN support in NMODL, these tests are very annoying. I can never find the difference; and it's completely redundant with #1131.

pramodk (approved):
Sounds reasonable!

PR #1175, opened 2024-02-26
Document cable equations.

1uc:
Starts write up of the underlying equations.


1uc:
I think this tells me enough to understand up-to conductances. Therefore, I suggest reviewing and merging if though it's very incomplete.

docs/contents/cable_equations.rst
```diff
@@ -0,0 +1,87 @@
+Cable Equation
+==============
+
+In order to derive the `cable equations` we model a neuron as an electrical
+circuit. We first pick points along the neuron at which we model the voltage.
+We'll call them nodes. At every branch point we place a node. The precise
+choice is described elsewhere, see Figure 1.
+
+.. figure:: ../images/cable-eqn_nodes.svg
```
nrnhines, line 9: A perhaps irrelevant detail. Rather than draw compartments as cylinders, one can draw as truncated cones so that the join diameters match.  The area of the compartment is the surface area of the truncated cone, and the resistor is calculated using the changing diameter from compartment center to adjacent compartment center.

nrnhines (approved):

docs/contents/cable_equations.rst
nrnhines: Or, rather, point to the relative version of 
https://nrn.readthedocs.io/en/8.2.3/python/modelspec/programmatic/topology/geometry.html#geometry-geometry

1uc: Thank you for clarifying this point. I'll update the figure:

I remember getting stuck on this a when I first tried to derive the equations. What I needed was a simple picture that just places some nodes. The crux was: is there a node at branch points? If the answer is yes, then things are simple. If the answer is no, I ended up in a bog and couldn't derived the governing equations.

If I understand correctly, a concern is that it's misleading about how sophisticated the geometry is in NRN. My hope is that I can arrive at a set of equations that's independent of the exact geometry.

What I'm working towards is text that enables me to explain:
* how to update the conductance from inside mechanisms,
* how to update the `rhs`,
* why we have `sav_*`.

PR #1176, opened 2024-02-27
Blame NMODL for lines it's generating.

1uc:
Adds

    nmodl MOD_FILE blame --line LINE

which will cause NMODL to print a backtrace every time it writes to line `LINE`.

This is very helpful when trying to find the NMODL code responsible for printing a (faulty) line of code.


JCGoran (approved):
Great addition 👍

PR #1177, opened 2024-02-27
Deactivate Gitlab CI.

1uc:
Gitlab is currently completely broken and unusable. Hence we deactivate it until it can be used.


pramodk:
@1uc: if it's matter of days (assuming its temporary gitlab glitch), is making it an optional from GitHub settings is an option as well?

1uc:
@pramodk it would still generate jobs on Gitlab. We don't know how they're gonna fix it. The "fix" might require us to cancel everything that didn't start. We can get it back with a simple `git revert`.

1uc:
We had to disable the "Required" check, because not running is the same as not passing. We ended up disabling all "Required" checks, because we should merge when CI is green and not merge when it's red.

If CI is spuriously red, it needs to be fixed and if that's absolutely impossible disabled or removed.

JCGoran (approved):
LGTM, we should just remember to re-enable it when the cluster is fixed.

PR #1182, opened 2024-02-27
Improve usecases 'func_proc{,_pnt}'.

1uc:
Check that if there's multiple instances of the mechanisms the functions and procedures use the values for the correct instance.


1uc:
Golden references changed: https://github.com/BlueBrain/nmodl-references/pull/3

pramodk (approved):
make sense, LGTM!

PR #1184, opened 2024-02-28
Pass inst when passing id.

1uc:
For functions that are meant to be called "in a loop", i.e. function
which accept `id`, we also pass `inst`.


1uc:
See reference got updated: https://github.com/BlueBrain/nmodl-references/pull/3

pramodk (approved):

PR #1192, opened 2024-02-29
Implement access to neuron global variables.

1uc:
What's meant is things like `celsius`.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -596,7 +596,7 @@ std::string CodegenNeuronCppVisitor::get_variable_name(const std::string& name,
     if (iter != info.neuron_global_variables.end()) {
         std::string ret;
         if (use_instance) {
-            ret = "*(inst->";
+            ret = "*(inst.";
         }
         ret.append(varname);
         if (use_instance) {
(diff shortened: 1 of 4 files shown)


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -596,7 +596,7 @@ std::string CodegenNeuronCppVisitor::get_variable_name(const std::string& name,
     if (iter != info.neuron_global_variables.end()) {
         std::string ret;
         if (use_instance) {
-            ret = "*(inst->";
+            ret = "*(inst.";
```
ohm314, line 599: was that an error before?
by the way, why not use the `->` operator and drop the `*(...)` instead?

test/usecases/parameter/simulate.py
```diff
@@ -1,12 +1,24 @@
...
+celsius_hoc = h.Vector().record(s(0.5)._ref_range_celsius_NeuronVariables)
+
+h.stdinit()
+h.tstop = 5.0 * ms
+h.run()
+
+celsius = np.array(celsius_hoc.as_numpy())
+assert celsius[-1] == 6.3, f"{celsius[-1]=}"
+
+
```
1uc, line 24: We might need to configure the formatter to run `black` of these files.

test/usecases/parameter/neuron_variables.mod
```diff
@@ -0,0 +1,16 @@
...
+PARAMETER {
+    celsius = 123.456
+}
+
+ASSIGNED {
+    range_celsius
+}
+
+BREAKPOINT {
+  range_celsius = celsius
```
1uc, line 15:
```suggestion
    range_celsius = celsius
```

src/codegen/codegen_neuron_cpp_visitor.cpp
ohm314: ah.. duh.. yes obviously....

ohm314 (approved):

PR #1193, opened 2024-03-01
Don't install fmt and backward.

1uc:


matz-e (approved):

1uc:
@pramodk  The question for review is: can NMODL generate code that uses fmt for format a string? Something like:

```
    static void nrn_alloc_funkymech(Prop* _prop) {
        int k = 42;
        std::string name = fmt::format("foo_{}, k);

```

alkino (approved):

pramodk:
> @pramodk The question for review is: can NMODL generate code that uses fmt for formatting a string? Something like:

No, it doesn't! (it's only used inside visitors like codegen)

PR #1207, opened 2024-03-07
Enable formatting Python files.

1uc:


.bbp-project.yaml
```diff
@@ -19,3 +19,9 @@ tools:
       match:
       - ext/.*
       - src/language/templates/*
+  Black:
```
matz-e, line 22: Future reference: `ruff` can do linting and black-style format checks.

matz-e (approved):

.bbp-project.yaml
1uc: The BBP formatter does: C++, CMake and Python, including installing the correct version of all of those automatically.

PR #1209, opened 2024-03-08
Fix indexed COMPARTMENT block.

1uc:
When encountering an indexed `COMPARTMENT` block:

    COMPARTMENT i, volume_expr { species }

All state variables are searched if they match they match the pattern
`"{species}[%d]"`. For each matching state variables, we obtain the
value of the array index and substitute the name of the array index with
its value in `volume_expr`. This is then stored in the array of
compartment factors.

If multiple species are listed, the above happens for each (array-valued) species.

### DIFF
--- a/src/visitors/kinetic_block_visitor.cpp
+++ b/src/visitors/kinetic_block_visitor.cpp
@@ -8,10 +8,13 @@
 #include "kinetic_block_visitor.hpp"
 
 #include "ast/all.hpp"
+#include "index_remover.hpp"
 #include "symtab/symbol.hpp"
 #include "utils/logger.hpp"
 #include "utils/string_utils.hpp"
 #include "visitor_utils.hpp"
+
+#include <regex>
 
 
 namespace nmodl {
@@ -155,7 +158,6 @@ void KineticBlockVisitor::set_compartment_factor(int var_index, const std::strin
 
 void KineticBlockVisitor::compute_compartment_factor(ast::Compartment& node,
                                                      const ast::Name& name) {
-    // COMPARTMENT volume { species }
     const auto& var_name = name.get_node_name();
     const auto it = state_var_index.find(var_name);
     if (it != state_var_index.cend()) {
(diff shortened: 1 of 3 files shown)


1uc:
We'll want to merge #1206 first.

test/unit/visitor/kinetic_block.cpp
```diff
@@ -244,6 +244,26 @@ SCENARIO("Convert KINETIC to DERIVATIVE using KineticBlock visitor", "[kinetic][
...
+            DERIVATIVE states {
+                x'[0] = ((-1*(f(v)*x[0]*x[1])))/(vol[0])
+                x'[1] = ((-1*(f(v)*x[0]*x[1])))/(vol[1])
+            })";
+        THEN("Convert to equivalent DERIVATIVE block") {
+            auto result = run_kinetic_block_visitor(input_nmodl_text);
+            CAPTURE(input_nmodl_text);
+            REQUIRE(result[0] == reindent_text(output_nmodl_text));
+        }
+    }
```
1uc, line 266: @nrnhines could you please review this test? The two fragments of a MOD file should be equivalent, i.e. the reaction equation is converted to that system of ODEs. The bug was that the term `.../(vol[0])` was missing.

nrnhines: looks good to me. It is the same ode form as produced by nocmodl for ```static int _ode_spec1(```

src/visitors/kinetic_block_visitor.cpp
```diff
@@ -141,24 +144,71 @@ void KineticBlockVisitor::visit_conserve(ast::Conserve& node) {
...
+    }
+}
+
+void KineticBlockVisitor::compute_indexed_compartment_factor(ast::Compartment& node,
+                                                             const ast::Name& name) {
+    auto array_var_name = name.get_node_name();
+    auto index_name = node.get_name()->get_node_name();
+
+    auto pattern = fmt::format("^{}\\[([0-9]*)\\]$", array_var_name);
+    std::cout << "pattern = " << pattern << std::endl;
```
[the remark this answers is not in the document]
1uc: rebased and removed the debug output.

matz-e (approved):

PR #1211, opened 2024-03-08
Extend time limit for azure ci.

1uc:
I've seen the same Azure pipeline fail with a timeout three times today. Either the runner we're getting are a bit slower or the new build-system is a little slower. We should bump to time-limit to avoid having to start over after 45 minutes.


JCGoran (approved):

JCGoran:
For clarity, the build system is actually faster per Python version, it's just that after #1155 we've started building all of the Python wheels (3.8-3.12) instead of just the 3.11 one.

PR #1212, opened 2024-03-11
Try to fix homebrew CI issue.

1uc:


JCGoran (approved):
Looks like it's a recurring issue with MacOS runners (or rather, anything using brew):
- https://github.com/actions/runner-images/issues/9471
- https://github.com/actions/runner-images/issues/8838
- https://github.com/actions/runner-images/issues/4020

To quote one of the comments:
> It's a Homebrew packages dependencies bug. Quite old and infamous to be honest. We can do nothing from our side to fix it - next image will contain new packages by design and it will fix this issue until the next dependencies update and again and again.

So if this works for now, LGTM!

PR #1214, opened 2024-03-12
Unroll kinetic blocks.

1uc:
Fixes part of a bug related to reaction equations inside loops.

### DIFF
--- a/test/unit/visitor/kinetic_block.cpp
+++ b/test/unit/visitor/kinetic_block.cpp
@@ -39,12 +39,6 @@ std::vector<std::string> run_kinetic_block_visitor(const std::string& text) {
     // construct symbol table from AST
     SymtabVisitor().visit_program(*ast);
 
-    // unroll loops and fold constants
-    ConstantFolderVisitor().visit_program(*ast);
-    LoopUnrollVisitor().visit_program(*ast);
-    ConstantFolderVisitor().visit_program(*ast);
-    SymtabVisitor().visit_program(*ast);
-
     // run KineticBlock visitor on AST
     KineticBlockVisitor().visit_program(*ast);
 
--- a/test/unit/visitor/kinetic_block.cpp
+++ b/test/unit/visitor/kinetic_block.cpp
@@ -42,6 +42,10 @@ std::vector<std::string> run_kinetic_block_visitor(const std::string& text) {
     // run KineticBlock visitor on AST
     KineticBlockVisitor().visit_program(*ast);
 
+    // Since KineticBlockVisitor internally performs const-folding and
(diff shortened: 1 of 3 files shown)


test/unit/visitor/kinetic_block.cpp
```diff
@@ -39,12 +39,6 @@ std::vector<std::string> run_kinetic_block_visitor(const std::string& text) {
     // construct symbol table from AST
     SymtabVisitor().visit_program(*ast);
 
-    // unroll loops and fold constants
-    ConstantFolderVisitor().visit_program(*ast);
-    LoopUnrollVisitor().visit_program(*ast);
-    ConstantFolderVisitor().visit_program(*ast);
-    SymtabVisitor().visit_program(*ast);
```
1uc, line 46: Question for review: This line threatens that we need to run SymtabVisitor after things like const folding or unrolling. However, the way the fix is implement I don't think it'll run `SymtabVisitor` again. Do we need to run it as part of `unroll_kinetic_blocks`?

pramodk: (I am looking at this after quite some time, so feel free to point out anything obvious that I might have forgotten)

`LoopUnroll` or `ConstantFolder` doesn't seem to add or delete any new symbols/variables into the AST. So I "think" `SymtabVisitor` is not necessary. But as it modifies the AST structure, it would be appropriate to run `CheckParentVisitor` as we do [in the tests](https://github.com/BlueBrain/nmodl/blob/c8818c8fb13cc32b92ded4477a00e9a3b636a00f/test/unit/visitor/loop_unroll.cpp#L42).

1uc: After loop unrolling the number of state variables will be different, it'll haves things like `ca[0]` which for purposes of the KINETIC block is very much a single identifier. Not sure if it's also true for the symbol table. Similarly, unrolling will eliminate the loop variable.

pramodk: I didn't realise both of the above points. If the number of state variables are changed then make sense to run the visitor?

matz-e (approved):

test/unit/visitor/kinetic_block.cpp
pramodk: I am relooking after seeing `tests are green` comment on the slack:

* index variables from `FROM` loop are not in the symbol table as they don't have explicit declarations
* even if state variables are going to change with unroll, as part of this changeset / pass, we are not changing anything in the AST (e.g. changing the `STATE` block).

pramodk (approved):
LGTM

Issue #1190, opened 2024-02-29
Review setting conductances when generating code for NEURON.


1uc:
When implementing the nonspecific current, we didn't implement setting the conductance correct. We've documented our understanding of the cable equations. The usecase for non-specific current passes, but if one looks carefully, and plots the solution, it differs between NRN + nocmodl and NRN+nmodl.

The task is to check that the conductance is set correctly when generating code for NEURON (as opposed to generating code for CoreNEURON).

PR #1218, opened 2024-03-14
Correctly set conductances when using NONSPECIFIC_CURRENT

JCGoran:
Attempt at fixing #1190

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -1238,12 +1238,23 @@ void CodegenNeuronCppVisitor::print_nrn_jacob() {
     printer->add_newline(2);
     printer->add_line("/** nrn_jacob function */");
 
-    printer->fmt_line(
+    printer->fmt_push_block(
         "static void {}(_nrn_model_sorted_token const& _sorted_token, NrnThread* "
-        "_nt, Memb_list* _ml_arg, int _type){{}}",
+        "_nt, Memb_list* _ml_arg, int _type)",
         method_name(naming::NRN_JACOB_METHOD));
 
+    printer->add_multi_line(
+        "_nrn_mechanism_cache_range _lmr{_sorted_token, *_nt, *_ml_arg, _type};");
 
+    printer->fmt_line("auto inst = make_instance_{}(_lmr);", info.mod_suffix);
+    printer->fmt_line("auto node_data = make_node_data_{}(*_nt, *_ml_arg);", info.mod_suffix);
+    printer->add_multi_line(R"CODE(
+    auto nodecount = _ml_arg->nodecount;
+    for (int id = 0; id < nodecount; id++) {
+        int node_id = node_data.nodeindices[id];
+        node_data.node_diagonal[node_id] += inst.g_unused[id];
+    })CODE");
+    printer->pop_block();
 }
 
 
@@ -1582,9 +1593,9 @@ void CodegenNeuronCppVisitor::print_nrn_cur() {
 
 
     printer->add_line("node_data.node_rhs[node_id] -= rhs;");
-
     // TODO probably a good idea to call `g` something else
-    printer->add_line("node_data.node_diagonal[id] += g;");
+    printer->add_line("inst.g_unused[id] = g;");
+
 
     printer->pop_block();
 
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -1241,7 +1241,7 @@ void CodegenNeuronCppVisitor::print_nrn_jacob() {
     printer->fmt_push_block(
         "static void {}(_nrn_model_sorted_token const& _sorted_token, NrnThread* "
         "_nt, Memb_list* _ml_arg, int _type)",
-        method_name(naming::NRN_JACOB_METHOD));
+        method_name(naming::NRN_JACOB_METHOD));  // begin function
 
     printer->add_multi_line(
         "_nrn_mechanism_cache_range _lmr{_sorted_token, *_nt, *_ml_arg, _type};");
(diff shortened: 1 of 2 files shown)


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1236,10 +1238,32 @@ void CodegenNeuronCppVisitor::print_nrn_jacob() {
...
+
+    printer->fmt_line("auto inst = make_instance_{}(_lmr);", info.mod_suffix);
+    printer->fmt_line("auto node_data = make_node_data_{}(*_nt, *_ml_arg);", info.mod_suffix);
+    printer->fmt_line("auto nodecount = _ml_arg->nodecount;");
+    printer->push_block("for (int id = 0; id < nodecount; id++)");  // begin for
+
+    for (const auto& variable: codegen_float_variables) {
+        if (variable->get_name() ==
+                std::string(nmodl::codegen::naming::CONDUCTANCE_UNUSED_VARIABLE) ||
+            variable->get_name() == std::string(nmodl::codegen::naming::CONDUCTANCE_VARIABLE)) {
```
JCGoran, line 1257: Note to reviewer: should we only use `CONDUCTANCE_UNUSED_VARIABLE`, or also `CONDUCTANCE_VARIABLE` (when is each one generated/used)?

JCGoran:
Associated PR for updating references: https://github.com/BlueBrain/nmodl-references/pull/7

JCGoran:
Also a question for the reviewer: should I add a test for this feature?

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: If we follow CoreNEURON (for now), we don't need the conductance at all. If you compare, it doesn't store `g` computed in one of the currents anywhere. For debugging purposes it can store the value in `g_unused`.

NEURON  is different, it stores it to retrieve it in the function `*_jacob_*`. If you want to also implement that part we would probably want to understand why there's this difference. One explanation could be that it's needed for variable timestepping. If we need to store it then the correct place to store would be in a range variable called `_g` (same as NEURON uses), probably by renaming `g_unused` to `_g` when compiling for NRN.

1uc: I think by looping over the `codegen_float_variables` you risk that a mod file defines a variable that happens to be called `"g"` (aka `nmodl::codegen::naming::CONDUCTANCE_VARIABLE`), but with the meaning that `"g"` is whatever they define it to be in that MOD file; and then you overwrite with something the MOD file didn't expect.

JCGoran:
For future reference, when using the `nonspecific_current` mod file, the call stack of NEURON+NOCMODL looks a bit like this (relevant function calls and assignment statements are pointed out):

```plaintext
nrn_fixed_step_thread (or nrn_finitialize if looking at the first run)
  setup_tree_matrix
    nrn_rhs
      vec_rhs[i] = 0. # zero out RHS, L405
      nrn_cur_[MECHANISM] # updates RHS, but not conductances
      dv = vec_v[pi] - vec_v[i]
      vec_rhs[i] -= vec_b[i] * dv # updates RHS (axial?)
      vec_rhs[pi] += vec_a[i] * dv # also updates RHS (axial?)
    nrn_lhs
      vec_d[i] = 0. # zero out diag
      nrn_jacob_[MECHANISM] # we set conductances here originally (g_unused)
        _vec_d[_ni[_iml]] += _g
      nrn_cap_jacob
        vec_d[ni[i]] += cfac * ml_cache.fpfield<cm_index>(i) # updates diagonal
      vec_d[i] -= vec_b[i]
      vec_d[_nt->_v_parent_index[i]] -= vec_a[i]
  nrn_solve
    triang
      p = vec_a[i] / vec_d[i]
      pi = _nt->_v_parent_index[i]
      vec_d[pi] -= p * vec_b[i]
      vec_rhs[pi] -= p * vec_rhs[i]
    bksub
      vec_rhs[i] /= vec_d[i] # diagonal part
      vec_rhs[i] -= vec_b[i] * vec_rhs[_nt->_v_parent_index[i]] # axial part
      vec_rhs[i] /= vec_d[i] # second part using the diagonal part
  nrn_update_voltage
    vec_v[i] += vec_rhs[i] # update voltage from RHS
    nrn_capacity_current
      ml_cache.fpfield<i_cap_index>(i) = cfac * ml_cache.fpfield<cm_index>(i) *
        vec_rhs[ni[i]] # something using RHS
  nrn_fixed_step_lastpart
    nrn_state_[MECHANISM] # does nothing in our case
```

If we wanted to inline the setting of the diagonal/conductances (that is, place them in `nrn_cur_[MECHANISM]`), we would need to alter NEURON in the following way:

- move zeroing out of diagonal from `nrn_lhs` to `nrn_rhs`
- move `nrn_cap_jacob` from `nrn_lhs` to `nrn_rhs`

Unfortunately, this would require modifying code generated by NOCMODL as well, so it seems that for now we need to keep using `nrn_jac_[MECHANISM]` for setting the diagonal properly.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1235,10 +1237,29 @@ void CodegenNeuronCppVisitor::print_nrn_jacob() {
...
+    printer->fmt_line("auto node_data = make_node_data_{}(*_nt, *_ml_arg);", info.mod_suffix);
+    printer->fmt_line("auto nodecount = _ml_arg->nodecount;");
+    printer->push_block("for (int id = 0; id < nodecount; id++)");  // begin for
+
+    if (breakpoint_exist()) {
+        printer->add_line("// set conductances properly");
+        printer->add_line("int node_id = node_data.nodeindices[id];");
+        printer->fmt_line("node_data.node_diagonal[node_id] += inst.{}[id];",
+                          info.vectorize ? naming::CONDUCTANCE_UNUSED_VARIABLE
+                                         : naming::CONDUCTANCE_VARIABLE);
```
JCGoran, line 1258: Roughly the same thing as we do here: https://github.com/BlueBrain/nmodl/blob/2edf9aca43bdb80fa827b98f03a0d315c1e34e7c/src/codegen/codegen_cpp_visitor.cpp#L964-L975

1uc (approved):

Issue #717, opened 2021-08-16
Warnings from recent Bison and Flex versions


olupton:
Building with bison 3.6.4 and flex 2.6.4 I see
```
[4/193] -- NMODL : GENERATING C (11) PARSER WITH BISON! --
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/external/nmodl/src/parser/c11.yy:52.1-35: warning: deprecated directive: ‘%define parser_class_name {CParser}’, use ‘%define api.parser.class {CParser}’ [-Wdeprec
ated]
   52 | %define parser_class_name {CParser}
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      | %define api.parser.class {CParser}
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/external/nmodl/src/parser/c11.yy: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
[5/193] -- NMODL : GENERATING DIFFERENTIAL EQUATION PARSER WITH BISON! --
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/external/nmodl/src/parser/diffeq.yy:57.1-40: warning: deprecated directive: ‘%define parser_class_name {DiffeqParser}’, use ‘%define api.parser.class {DiffeqParser}’ [-Wdeprecated]
   57 | %define parser_class_name {DiffeqParser}
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      | %define api.parser.class {DiffeqParser}
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/external/nmodl/src/parser/diffeq.yy: warning: 5 shift/reduce conflicts [-Wconflicts-sr]
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/external/nmodl/src/parser/diffeq.yy: warning: fix-its can be applied.  Rerun with option '--update'. [-Wother]
[6/193] -- NMODL : GENERATING UNIT PARSER WITH BISON! --
/gpfs/bbp.cscs.ch/home/olupton/CoreNeuron/external/nmodl/src/parser/unit.yy:49.1-38: warning: deprecated directive: ‘%define parser_class_name {UnitParser}’, use ‘%define api.parser.class {UnitParser}’ [-Wdeprecated]
   49 | %define parser_class_name {UnitParser}
[...]

PR #1221, opened 2024-03-18
Replace deprecated bison commands

JCGoran:
TODO:
- [ ] according to [the docs](https://lists.gnu.org/archive/html/info-gnu/2019-01/msg00016.html), updating `name-prefix` to `api.prefix` is a bit non-trivial, and a naive fix results in compilation errors

### DIFF
--- a/src/parser/c11.yy
+++ b/src/parser/c11.yy
@@ -49,7 +49,7 @@
 %define api.namespace {nmodl::parser}
 
 /** set the parser's class identifier */
-%define parser_class_name {CParser}
+%define api.parser.class {CParser}
 
 /** keep track of the current position within the input */
 %locations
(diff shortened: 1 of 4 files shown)


alkino:
I never changed this one because Ubuntu still have old bison, IIRC

JCGoran:
According to https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md the Ubuntu runner should have Bison 3.5.1, so we should be good as far as versions are concerned.

alkino:
The problem is: https://packages.ubuntu.com/focal/bison

pramodk (approved):
> I never changed this one because Ubuntu still have old bison, IIRC

@alkino : I remember an old discussion where I was included to support previous versions. Looking at your comment here: https://github.com/BlueBrain/nmodl/issues/717#issuecomment-945723651, I guess now it's reasonable to support Ubuntu >= 20. With NEURON, users get NMODL via `pip install` anyway.

PR #1228, opened 2024-03-20
Add minor fixes to generate_references on MacOS

JCGoran:
For some reason, the previous version treated the quotation marks as literal, so it kept littering the `references` dir with backup files which had `''` appended to it. This one expands the variables properly (thanks bash for arrays!). Also, `realpath` may not exist on MacOS (of course...) so I've replaced it with `cd "$(dirname "$0")"; pwd -P` which should do the same thing cross-platform.

### DIFF
--- a/test/usecases/generate_references.sh
+++ b/test/usecases/generate_references.sh
@@ -8,21 +8,20 @@ if [[ $# -eq 3 ]]
 then
   output_dir="$3"
 else
-  script_dir="$(dirname "$(realpath "$0")")"
+  script_dir="$(cd "$(dirname "$0")"; pwd -P)"
   output_dir="${script_dir}/references/$(basename "$2")"
 fi
 
 function sanitize() {
   for f in "${1}"/*.cpp
   do
-    if [[ "$(uname)" == 'Darwin' ]]
-    then
-        sed_cmd="sed -i''"
+    if [[ "$(uname)" == 'Darwin' ]]; then
+        sed_cmd=("sed" "-i" "")
     else
-        sed_cmd="sed -i"
+        sed_cmd=("sed" "-i")
     fi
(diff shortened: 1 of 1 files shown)


test/usecases/generate_references.sh
```diff
@@ -8,21 +8,20 @@ if [[ $# -eq 3 ]]
 then
   output_dir="$3"
 else
-  script_dir="$(dirname "$(realpath "$0")")"
+  script_dir="$(cd "$(dirname "$0")"; pwd -P)"
```
1uc, line 11: Sure, though developers can simply install the required `coreutils`.

1uc (approved):

test/usecases/generate_references.sh
```diff
@@ -8,21 +8,20 @@ if [[ $# -eq 3 ]]
...
+    if [[ "$(uname)" == 'Darwin' ]]; then
+        sed_cmd=("sed" "-i" "")
     else
-        sed_cmd="sed -i"
+        sed_cmd=("sed" "-i")
     fi
-    ${sed_cmd} "s/Created         : .*$/Created         : DATE/" "$f"
-    ${sed_cmd} "s/NMODL Compiler  : .*$/NMODL Compiler  : VERSION/" "$f"
+    "${sed_cmd[@]}" "s/Created         : .*$/Created         : DATE/" "$f"
+    "${sed_cmd[@]}" "s/NMODL Compiler  : .*$/NMODL Compiler  : VERSION/" "$f"
```
alkino, line 24: What is the difference between the previous version and this new?

This is a problem with quoting?

JCGoran: > This is a problem with quoting?

Yes. I tried other variations, but none worked. This shouldn't be an issue since we run the script with Bash anyway.

PR #1231, opened 2024-04-02
Update usecase tests so they compare outputs with NOCMODL

JCGoran:
This PR modifies the testing of usecases as follows:

- instead of calling `[arch]/special`, we directly call Python, which allows us to actually parse command-line args (is there some downside/side-effect of doing this?)
- if there is an extra command line argument, _some_ (not all!) of the tests they will create an output file (either `nocmodl.txt` or `nmodl.txt`, depending on which one we use for the codegen) with the relevant quantities of interest (voltage as function of time, etc.)
- if the files differ, the `run_test.sh` script exits with an error code

All of the above should allow us to catch any differences between NEURON + NOCMODL and NEURON + NMODL (with NEURON codegen) a bit more easily.

Note that the nonspecific current test is not modified until #1218 is merged as the test fails.

### DIFF
--- a/test/usecases/cnexp_array/simulate.py
+++ b/test/usecases/cnexp_array/simulate.py
@@ -1,6 +1,8 @@
-import numpy as np
+import sys
 
-from neuron import h, gui
+import numpy as np
+from neuron import gui
+from neuron import h
 from neuron.units import ms
 
 nseg = 1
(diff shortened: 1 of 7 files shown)


1uc:
This type of diffing has caused considerable pain in the past, in is in many ways the opposite of what I'd like to build.

If there's a particular test success when is should fail, then the test should be made more strict. Once we implement enough functionality we'll gain access to ModelDB CI which does these types of checks.

PR #1232, opened 2024-04-04
Localize kinetic reaction rates.

1uc:
If assignments and reaction equations are interspersed the ODE might use incorrect reaction rates. Example:
```
rate = 42.0
~ a <-> b (1.0*rate, 2.0*rate)
rate = 43.0
~ c <-> d (3.0*rate, 4.0*rate)
```
was equivalent to:
```
rate = 42.0
rate = 43.0
~ a <-> b (1.0*rate, 2.0*rate)
~ c <-> d (3.0*rate, 4.0*rate)
```
This PR solves the issue by localizing the rates, i.e. it'll produce:
```
LOCAL kf0_, kb0_, kf1_, kb1_
rate = 42.0
kf0_ = 1.0*rate
kb0_ = 2.0*rate
~ a <-> b (kf0_, kb0_)
rate = 43.0
kf1_ = 3.0*rate
kb1_ = 4.0*rate
~ c <-> d (kf1_, kb1_)
```
Which can then be safely shuffled into:
```
LOCAL kf0_, kb0_, kf1_, kb1_
rate = 42.0
kf0_ = 1.0*rate
kb0_ = 2.0*rate

rate = 43.0
kf1_ = 3.0*rate
kb1_ = 4.0*rate

~ a <-> b (kf0_, kb0_)
~ c <-> d (kf1_, kb1_)
```
and from there transformed as usual.

### DIFF
--- a/src/visitors/kinetic_block_visitor.cpp
+++ b/src/visitors/kinetic_block_visitor.cpp
@@ -479,19 +479,42 @@ void KineticBlockVisitor::visit_kinetic_block(ast::KineticBlock& node) {
 }
 
 class LocalRateNames {
-    const std::string kf_stem = "kf";
-    const std::string kb_stem = "kb";
-    const std::string source_stem = "source";
+    static constexpr auto kf_stem = "kf";
+    static constexpr auto kb_stem = "kb";
+    static constexpr auto source_stem = "source";
 
     std::shared_ptr<ast::Name> generate_local_name(const std::string& stem) {
-        auto name = std::make_shared<ast::Name>(
-            std::make_shared<ast::String>(fmt::format("{}{}_", stem, n_equations)));
-        local_names.push_back(name);
+        std::string unmangled_name = fmt::format("{}{}_", stem, n_equations);
+        std::string mangled_name = unmangled_name;
+
+        size_t mangle_attempt = 0;
+        while (symtab->lookup_in_scope(mangled_name) ||
+               std::find(local_names.begin(), local_names.end(), mangled_name) !=
+                   local_names.end()) {
+            mangled_name = fmt::format("{}{:04d}", unmangled_name, mangle_attempt++);
+
+            if (mangle_attempt >= 10000) {
+                throw std::runtime_error("Failed to find unique local name.");
+            }
+        }
+
+        auto name = std::make_shared<ast::Name>(std::make_shared<ast::String>(mangled_name));
+        local_names.push_back(mangled_name);
 
         return name;
     }
 
   public:
+    LocalRateNames() = default;
+    LocalRateNames(const LocalRateNames&) = default;
+    LocalRateNames(LocalRateNames&&) = default;
+
+    LocalRateNames(symtab::SymbolTable const* symtab)
+        : symtab(symtab) {}
+
+    LocalRateNames& operator=(const LocalRateNames&) = default;
+    LocalRateNames& operator=(LocalRateNames&&) = default;
+
     std::shared_ptr<ast::Name> generate_forward_rate_name() {
         auto kf = generate_local_name(kf_stem);
         ++n_equations;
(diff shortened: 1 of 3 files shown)


src/visitors/kinetic_block_visitor.cpp
```diff
@@ -478,6 +478,162 @@ void KineticBlockVisitor::visit_kinetic_block(ast::KineticBlock& node) {
...
+class LocalRateNames {
+    const std::string kf_stem = "kf";
+    const std::string kb_stem = "kb";
+    const std::string source_stem = "source";
+
+    std::shared_ptr<ast::Name> generate_local_name(const std::string& stem) {
+        auto name = std::make_shared<ast::Name>(
+            std::make_shared<ast::String>(fmt::format("{}{}_", stem, n_equations)));
+        local_names.push_back(name);
+
```
pramodk, line 490: just a general comment: one precaution we have to usually take is to see if the variable doesn't conflict with the existing variable. E.g. what is the chance that `kf_0` is an existing user-defined variable? 

One could check if the new variable already exists in the symbol table. IIRC, for a similar reason, methods like [suffix_random_string ](https://github.com/BlueBrain/nmodl/blob/735b11dba9a494a5139be2ad5b82ca15e12859db/src/visitors/visitor_utils.hpp#L45) were added. There is also `class SingletonRandomString`.

src/visitors/kinetic_block_visitor.cpp
```diff
@@ -478,6 +478,162 @@ void KineticBlockVisitor::visit_kinetic_block(ast::KineticBlock& node) {
...
+        if (it != stmts.cend()) {
+            auto local_block = std::dynamic_pointer_cast<ast::LocalListStatement>(*it);
+            for (const auto& local: locals) {
+                local_block->emplace_back_local_var(local);
+            }
+        } else {
+            auto local_stmt = std::make_shared<ast::LocalListStatement>(locals);
+            stmt_block->insert_statement(stmts.cbegin(), local_stmt);
+        }
```
pramodk, line 574: could you check if we can use [add_local_variable()](https://github.com/BlueBrain/nmodl/blob/735b11dba9a494a5139be2ad5b82ca15e12859db/src/visitors/visitor_utils.cpp#L111) from visitor_utils?

pramodk (commented):
Two minor suggestions. It's done nicely! With the tests, it quite clear! 👌

src/visitors/kinetic_block_visitor.cpp
1uc: Correct! Note, it's not `kf_0` but `kf0_` we need to dodge (with the hope that users have better taste than naming a variable with a trailing `_`).

The problem is readability I'd like to avoid `kf0_hweutwhr` and similar. It would be nicer if we could get access to `_kf0`; but it tells me that we can't have names that start with `_`. Which is good, if it would let me state that I'm not a user, and I can name my variables whatever I want.

pramodk: > with the hope that users have better taste than naming a variable with a trailing _

Understand! just that to me it's appears a bit stronger assumption if one assumes the user won't declare a variable with the name `X` :).

Just a suggestion: if you would like to keep readability aspect, see if it's easy just to do a check with `symtab->lookup_in_scope(name)`.

src/visitors/kinetic_block_visitor.cpp
1uc: done.

src/visitors/kinetic_block_visitor.cpp
1uc: We mangle as `kf0_{:04d}` and count upwards, checking both the symtab of the current statement block.

test/unit/visitor/kinetic_block.cpp
```diff
@@ -413,21 +453,54 @@ SCENARIO("Convert KINETIC to DERIVATIVE using KineticBlock visitor", "[kinetic][
...
+            }
+            KINETIC states {
+                LOCAL kf0_0000, kb0_0000
+                ~ x <-> y (kf0_, b)
+            })";
+        std::string output_nmodl_text = R"(
+            DERIVATIVE states {
+                LOCAL kf0_0000, kb0_0000, kf0_0001, kb0_
+                kf0_0001 = kf0_
+                kb0_ = b
```
pramodk, line 494: 👌

src/visitors/kinetic_block_visitor.cpp
```diff
@@ -478,6 +478,172 @@ void KineticBlockVisitor::visit_kinetic_block(ast::KineticBlock& node) {
...
+        std::string mangled_name = unmangled_name;
+
+        size_t mangle_attempt = 0;
+        while (symtab->lookup_in_scope(mangled_name)) {
+            mangled_name = fmt::format("{}{:04d}", unmangled_name, mangle_attempt++);
+
+            if (mangle_attempt >= 10000) {
+                throw std::runtime_error("Failed to find unique local name.");
+            }
+        }
```
pramodk, line 497: seems more than sufficient!

pramodk (approved):

PR #1233, opened 2024-04-05
Fix array variables.

1uc:
These are the code generation changes required for:

    https://github.com/neuronsimulator/nrn/pull/2779

The solution is to fill `nrn_prop_param_size` with the number of doubles, not number of variables.


pramodk (approved):
LGTM. Let's see about the CI status, Azure failure looks spurious.

PR #1235, opened 2024-04-08
Add test for nonspecific current

JCGoran:
Since NEURON's ODE solver is not extremely accurate in the intermediate region where the jump is present, we only compare the final values (without the fix for the conductances, the result of `v` is a bunch of `nan`s).

This requires updating the references again: https://github.com/BlueBrain/nmodl-references/pull/9


1uc:
I would have preferred to not replace the test, but to have both: the existing one to check accuracy of `v(t)` and the new one to check stability. To have both it might be better to make the constant `0.005` a parameter that we can set from the Python script. Then one can use the same MOD file for both cases.

JCGoran:
> To have both it might be better to make the constant 0.005 a parameter that we can set from the Python script.

Since the current workflow is `nrnivmodl -> python simulate.py`, I don't see a way to set the parameters in the mod file in the Python script itself (that is, unless we add another script beforehand which modifies the AST of the mod file).

JCGoran:
Added new test (`simulate2.py` with mod file `leonhard2`).

1uc:
You can have global or range variables in the MOD file:
```

UNITS {
        (mA) = (milliamp)
}

NEURON {
        SUFFIX leonhard2
        NONSPECIFIC_CURRENT il
        RANGE c
}

ASSIGNED {
        il (mA/cm2)
}

BREAKPOINT {
        il = c * (v - 1.5)
}
```
(or similar). And then set from Python:
```
s.leonhard_c = 0.05
# s.c_leonhard = 0.05
```
(or some variant thereof.)

JCGoran:
> You can have global or range variables in the MOD file:
> ...

Huh, that's neat, updated!

1uc:
It's now a bit hard to understand how the two tests are very different. One is about checking the accuracy of `v(t)`. The other is about checking the stability of the numerical method, e.g. that the method is fully implicit.

We check the one my using small time-steps and checking all the value along the way. While, to me, checking stability is about taking a (stupidly) large timestep and checking it doesn't explode. I understand that we can check both by simply cranking `c` to be a higher value, under the assumption that `dt` doesn't automatically shrink. However, I can't judge if `0.5` is large enough compared to the sane value `0.005`.

The worry is that future maintenance would remove the second `compare(0.5)` because it's "redundant".

1uc (approved):
Thank you, this is a lot clearer to me.

PR #1237, opened 2024-04-12
Functioning hh.mod

JCGoran:
Introduces the necessary changes to get `hh.mod` to compile and run properly.

TODO:

- [x] add test(s?) (as an analytical solution is not available, not really sure how to test this without directly diffing w.r.t. NOCMODL, cf. #1231)
- [x] update references (see https://github.com/BlueBrain/nmodl-references/pull/13)

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -219,6 +219,16 @@ void CodegenNeuronCppVisitor::print_function_or_procedure(const ast::Block& node
         printer->fmt_line("int ret_{} = 0;", name);
     }
 
+    if (node.is_procedure_block() || node.is_function_block()) {
+        const auto& parameters = node.get_parameters();
+        auto result = std::find_if(parameters.begin(), parameters.end(), [](auto var) {
+            return var.get()->get_node_name() == "v";
+        });
+        if (result == parameters.end()) {
+            printer->fmt_line("auto v = inst.{}[id];", naming::VOLTAGE_UNUSED_VARIABLE);
+        }
+    }
+
     print_statement_block(*node.get_statement_block(), false, false);
     printer->fmt_line("return ret_{};", name);
     printer->pop_block();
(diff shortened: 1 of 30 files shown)


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -219,6 +219,16 @@ void CodegenNeuronCppVisitor::print_function_or_procedure(const ast::Block& node
         printer->fmt_line("int ret_{} = 0;", name);
     }
 
+    if (node.is_procedure_block() || node.is_function_block()) {
```
JCGoran, line 222: This is necessary because of the strange way NEURON treats global variables; in `hh.mod`, we have the following problem:
- the PROCEDURE `states` takes no args
- the PROCEDURE `rates` takes one arg, but which is actually _not_ an arg but the global voltage parameter `v`

On top of this, in the generated code, `v` is actually `v_unused`, hence the code here (`v_unused` is actually `naming::VOLTAGE_UNUSED_VARIABLE`).

1uc: If I look at `nocmodl` generated code I get:
```
static int  rates ( _internalthreadargsprotocomma_ double _lv ) {
  // ..
  _lalpha = .1 * vtrap ( _threadargscomma_ - ( _lv + 40.0 ) , 10.0 ) ;
  //..
}

static void {_hoc,_npy}_rates(void) {
  // ..
  rates ( _threadargscomma_ *getarg(1) );
}

static void initmodel(_internalthreadargsproto_) {
  // ... 
  rates ( _threadargscomma_ v ) ;
}
```
which to me reads like the argument of `rates` is a regular variable, that happens to be called `v` (and shadows the node value `v`).

Similarly for CoreNEURON:
```
inline int rates_hodhux(int id, int pnodecount, hodhux_Instance* inst, double* data, const Datum* indexes, ThreadDatum* thread, NrnThread* nt, double v, double arg_v);
```
(which is then never used).

nrnhines: > the argument of `rates` is a regular variable, that happens to be called `v` (and shadows the node value `v`).

That is correct, and the reason in nocmodl that the arg is translated as _lv using the pattern _larg.

JCGoran: So what's the recommended fix here exactly?

1uc (approved):
We've have: https://github.com/BlueBrain/nmodl/issues/1255 to fix the outstanding issue.

PR #1238, opened 2024-04-15
Implement simplified expsyn.

1uc:
Instead of setting the conductance as:

    g = g + g0 * exp(-(t - t0)/tau)

we set it to

    g += g0

where `g0` is the argument passed to NET_RECEIVE, i.e. the weight.

This allows for an analytic solution, that can be tested.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -15,9 +15,11 @@
 
 #include "ast/all.hpp"
 #include "codegen/codegen_utils.hpp"
+#include "codegen_naming.hpp"
 #include "config/config.h"
 #include "utils/string_utils.hpp"
 #include "visitors/rename_visitor.hpp"
+#include "visitors/var_usage_visitor.hpp"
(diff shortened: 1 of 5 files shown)


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1775,6 +1799,70 @@ void CodegenNeuronCppVisitor::print_net_event_call(const ast::FunctionCall& node
...
+    args.emplace_back("", "double", "", "_lflag");
+
+    printer->fmt_push_block("static void nrn_net_receive_{}({})",
+                            info.mod_suffix,
+                            get_parameter_str(args));
+
+    rename_net_receive_arguments(*node, *node);
+
+    printer->add_line("_nrn_mechanism_cache_instance _ml_obj{_pnt->prop};");
+    printer->add_line("auto * _nt = (NrnThread*) _pnt->_vnt;");
```
JCGoran, line 1852: Minor nitpick: shouldn't we use `static_cast` here (just for the sake of explicitness)?

JCGoran (approved):

PR #1241, opened 2024-04-16
Launch Python embedded interpreter only if using sympy

alkino:


1uc (approved):

1uc:
We need to check that it can still compile a MOD file containing a KINETIC block. The suspicion is that a KINETIC block will ask for sympy even if the subcommand `sympy` wasn't requested.

The other question is if we can do away with the initialization code entirely. I think there's a singleton that represents "access to Python". Maybe we can initialize the singleton properly when we first use it (and no sooner).

src/main.cpp
```diff
@@ -580,7 +580,7 @@ int main(int argc, const char* argv[]) {
         }
     }
 
-    if (sympy_opt) {
+    if (app.got_subcommand(sympy_opt)) {
```
1uc, line 583: This feels bug-prone. If we need to finalize, we should finalize if it was initialized. But we should ask the wrapper if it was initialized, and not rely on `app.got_subcommand(sympy_opt)`.

1uc:
I tried to find places where the embedded Python is used, but I only found the Sympy related stuff. Then I check if Sympy is used within a different visitor; but didn't find anything.

Therefore, this looks correct to me.

1uc (approved):

PR #1242, opened 2024-04-17
Check python compile-time vs runtime

alkino:
the ABI otherwise is not the same and can lead to segfaults

### DIFF
--- a/src/pybind/pyembed.cpp
+++ b/src/pybind/pyembed.cpp
@@ -13,6 +13,9 @@
 #include "pybind/pyembed.hpp"
 #include "utils/logger.hpp"
 
+#define STRINGIFY(x) #x
+#define TOSTRING(x)  STRINGIFY(x)
+
 namespace fs = std::filesystem;
 
 namespace nmodl {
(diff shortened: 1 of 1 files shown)


src/pybind/pyembed.cpp
```diff
@@ -45,6 +48,23 @@ void EmbeddedPythonLoader::load_libraries() {
...
+
+    // This code is imported from PyBind11 because this is primarly in details for internal usage
+    // License of PyBind11 is BSD-style
+    {
+        const char* compiled_ver = TOSTRING(PY_MAJOR_VERSION) "." TOSTRING(PY_MINOR_VERSION);
+        const char* (*fun)(void) = (const char* (*) (void) ) dlsym(pylib_handle, "Py_GetVersion");
+        const char* runtime_ver = fun();
+        std::size_t len = std::strlen(compiled_ver);
+        if (std::strncmp(runtime_ver, compiled_ver, len) != 0 ||
+            (runtime_ver[len] >= '0' && runtime_ver[len] <= '9')) {
```
1uc, line 60: I don't understand this:
```
            (runtime_ver[len] >= '0' && runtime_ver[len] <= '9')) {
```
are you afraid that `runtime_ver` is longer than `compiled_ver`? If yes, shouldn't you check for `runtime_ver[len] != "\0"`?

src/pybind/pyembed.cpp
```diff
@@ -45,6 +48,23 @@ void EmbeddedPythonLoader::load_libraries() {
         logger->critical(errstr);
         throw std::runtime_error("Failed to dlopen");
     }
+
+    // This code is imported from PyBind11 because this is primarly in details for internal usage
+    // License of PyBind11 is BSD-style
+    {
+        const char* compiled_ver = TOSTRING(PY_MAJOR_VERSION) "." TOSTRING(PY_MINOR_VERSION);
```
1uc, line 55: We have `fmt` and could maybe do:
```
std::string compiled_ver = fmt::format("{}.{}", PY_MAJOR_VERSION, PY_MINOR_VERSION);
```

src/pybind/pyembed.cpp
alkino: `3.1` vs `3.11`

The `runtime_ver` is something like `"3.0a5+ (py3k:63103M, May 12 2008, 00:53:55) \n[GCC 4.2.3]"`

This is quite useless, but I really copy the code of `pybind` here.

1uc: Good explanation, thank you!

1uc (approved):

PR #1243, opened 2024-04-17
Fix dependency issues when included in NRN.

1uc:
In NRN we depend on the target `nmodl` to ensure that when we run `nrnivmodl -coreneuron` the binary `nmodl` exists. We need to also ensure that everything `nmodl` needs to works is present.

This fixes a dependency bug on certain copied files by making the target for copying the files a dependency of `nmodl`.


pramodk (approved):
LGTM

alkino (approved):

PR #1245, opened 2024-04-23
Bug fix for NEURON wheel creation

pramodk:
- when neuron is built with nmodl then nmodl related files should be installed in `<prefix>/` and not in `<prefix>/nmodl/.data`
  - `<prefix>/nmodl/.data` is used when we build standalone wheels
- add one mod file using sparse solver - nmodl will automatically use `sympy` solver

This was the root cause for neuronsimulator/nrn/issues/2494!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ add_custom_target(
 # =============================================================================
 # Adjust install prefix for wheel
 # =============================================================================
-if(NOT LINK_AGAINST_PYTHON)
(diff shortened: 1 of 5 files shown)


nrnhines (approved):
LGTM

alkino (approved):

pramodk:
The CI is green on NEURON side: https://dev.azure.com/neuronsimulator/nrn/_build/results?buildId=11802&view=results

PR #1247, opened 2024-05-02
Add TABLE statement

JCGoran:
Needs #1260.


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -754,7 +805,30 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
     }
 
     if (info.table_count > 0) {
-        throw std::runtime_error("Not implemented, global table count.");
+        // basically the same code as coreNEURON uses
```
JCGoran, line 808: As mentioned, this taken verbatim from coreNEURON. I could move this to `CodegenCppVisitor` if needed, since the code is identical (if there's no potential issues that could arise with other visitors).

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1775,6 +1879,220 @@ void CodegenNeuronCppVisitor::print_net_event_call(const ast::FunctionCall& node
     throw std::runtime_error("Not implemented.");
 }
 
+static const TableStatement* get_table_statement(const ast::Block& node) {
```
JCGoran, line 1882: Taken from coreNEURON codegen.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1775,6 +1879,220 @@ void CodegenNeuronCppVisitor::print_net_event_call(const ast::FunctionCall& node
...
+    if (table_statements.size() != 1) {
+        auto message = fmt::format("One table statement expected in {} found {}",
+                                   node.get_node_name(),
+                                   table_statements.size());
+        throw std::runtime_error(message);
+    }
+    return dynamic_cast<const TableStatement*>(table_statements.front().get());
+}
+
+void CodegenNeuronCppVisitor::print_table_replacement_function(const ast::Block& node) {
```
JCGoran, line 1894: Taken from coreNEURON codegen.

test/usecases/hodgkin_huxley/hodhux.mod
```diff
@@ -0,0 +1,114 @@
...
+FUNCTION vtrap(x,y) {  :Traps for 0 in denominator of rate eqns.
+        if (fabs(x/y) < 1e-6) {
+                vtrap = y*(1 - x/y/2)
+        }else{
+                vtrap = x/(exp(x/y) - 1)
+        }
+}
+ 
+UNITSON
+
```
pramodk, line 113: Based on slack conversation, I guess we can use simple MOD file & test: https://github.com/neuronsimulator/nrn/blob/master/docs/nmodl/python_scripts/function_table.py

src/codegen/codegen_neuron_cpp_visitor.cpp
pramodk: @JCGoran : IMO, if there is a common code then having it in the top level/base visitor would be helpful than having two copies.

pramodk (commented):
I understand this is a draft but I quickly skimmed through it.

src/codegen/codegen_neuron_cpp_visitor.cpp
JCGoran: Most of the common code was merged in #1251

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
+from neuron import h, gui
+
+s = h.Section(name="soma")
+s.L = 10
+s.diam = 10
+s.insert("tbl")
+
+h.k_tbl = -0.1
+h.d_tbl = -40
```
1uc, line 10: Not covered by the test is the feature that the table updates if `k` or `d` change.

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,33 @@
...
+std = (vvec.c(), gvec.c(), tvec.c())
+
+h.usetable_tbl = 1
+h.run()
+
+t_table = np.array(tvec.as_numpy())
+v_table = np.array(vvec.as_numpy())
+
+# the table should be accurate to 1%
+assert np.allclose(v, v_table, rtol=1e-2)
```
1uc, line 33: Is the behaviour for `v` outside of the tabulated range being checked?

1uc:
Further comments on: https://github.com/BlueBrain/nmodl-references/pull/20

1uc:
One should probably check if multiple tables are permitted and work. The reason is that we do use both `sig` and `sigmoid` to mangle names. Therefore, the question is can one have two different proceedures that both compute (a) `sig` (with different values) and both tabulate their `sig`.

test/usecases/table/simulate.py
JCGoran: I think we need a FUNCTION or something to properly test this, no? Since a PROCEDURE will always return an error code or 0.

test/usecases/table/table.mod
```diff
@@ -0,0 +1,31 @@
...
+}
+
+BREAKPOINT {
+    sigmoid1(v)
+    g = gmax * sig
+    i = g*(v - e)
+}
+
+PROCEDURE sigmoid1(v) {
+    TABLE sig DEPEND k, d FROM -100 TO 100 WITH 100
```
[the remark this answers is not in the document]
JCGoran: Fixed by using different numbers.

test/usecases/table/simulate.py
JCGoran: Thanks, added it now. Note that I don't vary the parameters very much since the interpolation becomes inaccurate, but it seems to actually respond properly to the change in values.

test/usecases/table/simulate.py
JCGoran: Added test for it.

JCGoran:
> One should probably check if multiple tables are permitted and work. The reason is that we do use both `sig` and `sigmoid` to mangle names. Therefore, the question is can one have two different proceedures that both compute (a) `sig` (with different values) and both tabulate their `sig`.

It seems this will not error out at the parsing stage, but will at the compilation stage:

```plaintext
NEURON {
    SUFFIX minimal
    RANGE v1, v2
    GLOBAL c1, c2
}

PARAMETER {
    c1 = 1
    c2 = 2
}

ASSIGNED {
    v1
    v2
}

PROCEDURE p1(arg) {
    TABLE v1, v2 DEPEND c1, c2 FROM -1 TO 1 WITH 100
    v1 = c1 * arg + c2
    v2 = c2 * arg + c1
}

PROCEDURE p2(arg) {
    TABLE v1, v2 DEPEND c1, c2 FROM -5 TO 5 WITH 2000
    v1 = c1 * arg
    v2 = c2 * arg
}
```

due to redefinitions of `v1` and `v2` tables:

```cpp
 -> Compiling arm64/ex.cpp
arm64/ex.cpp:220:17: error: redefinition of '_t_v1'
 static double *_t_v1;
                ^
arm64/ex.cpp:218:17: note: previous definition is here
 static double *_t_v1;
                ^
arm64/ex.cpp:221:17: error: redefinition of '_t_v2'
 static double *_t_v2;
                ^
arm64/ex.cpp:219:17: note: previous definition is here
 static double *_t_v2;
                ^
2 errors generated.
```
so I think it's safe to assume NOCMODL does not support duplicate TABLE variables. I'll see if I can add this check in NMODL as part of one of the visitors.

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+    section = setup_sim()
+
+    x = np.linspace(-4, 6, 300)
+
+    proc = section(0.5).tbl.example_procedure
+
+    def modified_args(arg):
+        """
+        Pass arg to the procedure and return new values
+        """
```
1uc, line 67: The comment is either pointing out the immediately obvious, or not drawing my attention to the surprise.

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+        proc(arg)
+        return section(0.5).tbl.v1, section(0.5).tbl.v2
+
+    h.c1_tbl = 1
+    h.c2_tbl = 2
+
+    h.usetable_tbl = 0
+
```
1uc, line 75:
```suggestion
```

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+
+def test_procedure():
+    section = setup_sim()
+
+    x = np.linspace(-4, 6, 300)
+
+    proc = section(0.5).tbl.example_procedure
+
+    def modified_args(arg):
```
1uc, line 64: Feels like a misnomer, because it's not modifying any args afaict. Maybe `run_proc` or `call_proc` or something might be a little clearer?

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        proc(arg)
+        return section(0.5).tbl.v1, section(0.5).tbl.v2
+
+    h.c1_tbl = 1
+    h.c2_tbl = 2
+
+    h.usetable_tbl = 0
+
+    v1_no_table, v2_no_table = np.transpose(np.array([modified_args(i) for i in x]))
```
1uc, line 76: Often Numpy functions will attempt to convert their input to numpy arrays. A quick test seemed to suggest that the explicit `np.array` might not be needed here.

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+    return section
+
+
+def check_table(y_no_table, y_table, rtol1=1e-2, rtol2=1e-8):
+    assert np.allclose(y_no_table, y_table, rtol=rtol1), f"{y_no_table} != {y_table}"
+    # if the table is not used, we should get identical results, but we don't,
+    # hence the assert below
+    assert not np.allclose(
+        y_no_table, y_table, rtol=rtol2
+    ), f"{y_no_table} != {y_table}"
```
1uc, line 19:
```suggestion
    ), f"Broken test logic: {y_no_table} == {y_table}"
```

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+        np.array([modified_args(i) for i in x])
+    )
+
+    h.usetable_tbl = 1
+    v1_params_table, v2_params_table = np.transpose(
+        np.array([modified_args(i) for i in x])
+    )
+
+    check_table(v1_params_table, v1_params_no_table, rtol1=1e-3)
+    check_table(v2_params_table, v2_params_no_table, rtol1=1e-3)
```
1uc, line 108: If you feel like playing golf, the preceeding 15 lines, are repeated from above. You could rework as:
```
   check_proc_table(1, 2)
   check_proc_table(0.1, 0.3)
```
Then you notice that if you then keep `v1` and `v2` in a single array, you can play golf again; and probably get it to also work for functions too. The only thing that differs is how you call the function/procedure. Hence you could inject that as a callback:
```
def compare_table_and_no_table(c1, c2, modified_args):
    h.c1_tbl = c1
    h.c2_tbl = c2

    h.usetable_tbl = 0
    values_no_table = np.array([modified_args(i) for i in x])

    h.usetable_tbl = 1
    values_with_table =(
        np.array([modified_args(i) for i in x])
    )

    check_table(values_with_table, values_no_table, rtol1=1e-3)
```
Then one round of renaming:
* `compare_table_and_no_table` would be `check_table`.
* `check_table` would unfortunately have to turn into `is_{similar, close}`. You might be able to inline it.

This way you can collapse four copy-paste repetitions into a single function, called four times.

test/usecases/table/simulate.py
```diff
@@ -0,0 +1,170 @@
...
+    # run with a table (same params as above)
+    h.usetable_tbl = 1
+    h.run()
+
+    v_params_table = np.array(vvec.as_numpy())
+
+    return t, v_exact, v_table, v_params, v_params_table
+
+
+def check_solution(v, v_table, rtol=1e-2):
```
1uc, line 152: You essentially already have this, it's called `check_table`. This name is better.

test/usecases/table/table.mod
```diff
@@ -0,0 +1,46 @@
...
+    g = gmax * sig
+    i = g*(v - e)
+}
+
+PROCEDURE sigmoid1(v) {
+    TABLE sig DEPEND k, d FROM -127 TO 128 WITH 155
+    sig = 1/(1 + exp(k*(v - d)))
+}
+
+FUNCTION example_function(arg) {
```
1uc, line 37: If you want to match the previous theme: `quadratic` and `sinusoidal` might work as function/procedure names. Not sure there's much advantage to it.

src/codegen/codegen_cpp_visitor.hpp
```diff
@@ -1409,7 +1417,9 @@ void CodegenCppVisitor::print_vector_elements(const std::vector<T>& elements,
  * different in case of table statement.
  */
 template <typename T>
-void CodegenCppVisitor::print_function_declaration(const T& node, const std::string& name) {
+void CodegenCppVisitor::print_function_declaration(const T& node,
+                                                   const std::string& name,
+                                                   bool hidden) {
```
1uc, line 1422: This feels like a point where we'd eventually end up with a lot of arguments. One for: `inline`, `constexpr`, `static`, `extern "C"` and probably a couple of others.

In other places we've noticed that function like `apply(values, f, true, false, true)` (or whatever it was called) aren't extremely easy to read. It's similar here `false` or `true` aren't closely linked to the notion of adding `static`.

Neither is a particular issue; but since it would fix both issues in one go, it might be worth changing the argument `hidden` to something line `modifiers` (or whatever the correct term is) and convert it into an enum.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -147,6 +152,41 @@ void CodegenNeuronCppVisitor::print_point_process_function_definitions() {
...
+                          table_function_prefix(),
+                          method_name(name),
+                          get_parameter_str(internal_params));
+    }
+
+    // definition of `_check_table_thread` function
+    // signature must be same as the `nrn_thread_table_check_t` type
+    printer->fmt_line(
+        "static void {}(Memb_list* _ml, size_t id, Datum* _ppvar, Datum* _thread, "
+        "NrnThread* _nt, int _type, _nrn_model_sorted_token const& _sorted_token)",
```
1uc, line 174: There are utilities in place to help with this:
* ParamVector
* get_parameter_str
and I feel the list is sufficiently long to use them. I'm aware there's other places where we don't either.

src/codegen/codegen_cpp_visitor.hpp
JCGoran: Okay, now I added the option of passing a `std::unordered_set` as the third arg, whose elements are `CppObjectSpecifier`s (basically an enum containing a bunch of common C++ object/function specifiers. Name is subject to change). I contemplated using bit flags, but figured the implementation would be more convoluted, and this part isn't exactly performance- or memory-critical, so I chose to go with a simple set container instead.

test/usecases/table/simulate.py
1uc: The two functions are semantically identical: check if the two arrays of values are the same up to a relative tolerance of `rtol`, while also ensuring that they're not identical.

1uc (approved):

PR #1248, opened 2024-05-03
Fix missing variables when calling functions or procedures

JCGoran:
The variables `_ml`, `_thread`, and `_ppvar` are sometimes needed when calling functions or procedures.

### DIFF
--- a/test/usecases/CMakeLists.txt
+++ b/test/usecases/CMakeLists.txt
@@ -6,7 +6,8 @@ set(NMODL_USECASE_DIRS
     parameter
     func_proc
     func_proc_pnt
-    nonspecific_current)
+    nonspecific_current
+    cabpump)
 
 file(GLOB NMODL_GOLDEN_REFERENCES "${CMAKE_CURRENT_SOURCE_DIR}/references/*")
 if(NMODL_GOLDEN_REFERENCES STREQUAL "")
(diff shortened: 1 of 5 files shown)


test/usecases/CMakeLists.txt
```diff
@@ -6,7 +6,8 @@ set(NMODL_USECASE_DIRS
     parameter
     func_proc
     func_proc_pnt
-    nonspecific_current)
+    nonspecific_current
+    cabpump)
```
pramodk, line 10: the PR title/description and changes LGTM. But I am wondering if we can use/modify the existing `func_proc*` tests to cover FUNCTION/PROCEDURE related aspects. 

(i.e. I am wondering what is specific in this `cabpump` test that can't be covered by the existing test. Also thinking that the mod file looks long/complicated for its purpose?)

JCGoran: I am all for simplifying it, I just took the test from https://github.com/BlueBrain/nmodl/blob/master/test/integration/mod/cabpump.mod and removed the TABLE statement. I'll see if I can reproduce it with a smaller mod file first.

JCGoran: Turns out the issue occurs when the INITIAL block has a function call, and a minimal change to the `func_proc` test exposes the issue.

1uc (approved):

PR #1253, opened 2024-05-08
Tests for voltage dependency function/procedures.

1uc:
Adds tests the demonstrates how the voltage `v` behaves when passed to functions or procedures.


JCGoran (approved):
Additional tests are always welcome ✅

Issue #1255, opened 2024-05-10
Pick one: `v_used` or not.


1uc:
Currently, we sometimes use `v_unused` and sometimes we dereference the pointer to the node value. This needs to be made consistent.

PR #1256, opened 2024-05-10
Remove usage of `v_unused`

JCGoran:
Attempt at fixing #1255.


1uc:
This might be helpful: https://github.com/neuronsimulator/nrn/pull/2768 otherwise you'll not have a way to get the Node from a Prop.

PR #1260, opened 2024-05-14
Fix missing variables when calling a FUNCTION or PROCEDURE in BREAKPOINT block

JCGoran:


test/usecases/nonspecific_current/leonhard.mod
```diff
@@ -17,5 +17,17 @@ PARAMETER {
...
 }
+
+PROCEDURE func() {
+}
+
+PROCEDURE func_with_v(v) {
+}
+
+PROCEDURE func_with_other(q) {
+}
```
1uc, line 33: This isn't related to NONSPECIFIC_CURRENT. We have a other usecases about functions/procedures `func_proc` (+pnt).

Maybe reuse one of those or start a new MOD file. Also please make it a bit stronger, be actually checking functionality. Currently, it only checks that everything compiles.

1uc (changes requested):
Please also consider comments made here:
https://github.com/BlueBrain/nmodl-references/pull/18

test/usecases/nonspecific_current/leonhard.mod
JCGoran: `nrn_cur` seems to be generated only when NONSPECIFIC_CURRENT is present, so I figured why not add the test directly there.

Regarding the functionality checking, I'm not sure how exactly this should be tested as the only thing missing are the variables, which prevent compilation from succeeding. I guess your concern is that something is not being passed correctly (by value instead of reference, etc.)?

1uc: The empty function bodies feel a lot like something a bug can work its way around quite easily. Then there's the aspect that it runs a simulation and asserts values. But those values are not affected in any way be the code under test.

If you only want to check that it compiles, you could replace `simulate.py` with:
```
# This usecase only checks that the generated code compiles.
print("success: ...")
```

test/usecases/func_in_breakpoint/func_in_breakpoint.mod
```diff
@@ -0,0 +1,27 @@
+NEURON {
+    SUFFIX func_in_breakpoint
+    NONSPECIFIC_CURRENT il
+}
+
+ASSIGNED {
+        il (mA/cm2)
```
1uc, line 7:
```suggestion
    il
```

test/usecases/nonspecific_current/leonhard.mod
JCGoran: Noted, `simulate.py` is now just that.

1uc (approved):

Issue #1258, opened 2024-05-14
Crash after SymPy fails to parse variable named `is` in mod file


kbvw:
When running NMODL on a mod file (`kv4_ch.mod`, found in https://github.com/Hjorthmedh/BasalGangliaData/tree/main/data/neurons/mechanisms) containing a variable named `is`, SymPy throws a `SyntaxError` that later results in NMODL crashing with seemingly unrelated errors.

```
(.venv_nmodl) vanwalst@bbpv1 ~/p/snudda> nmodl BasalGangliaData/data/neurons/mechanisms/kv4_ch.mod
INFO : Using nmodl-nightly Package (Developer Version)
[NMODL] [warning] :: SympySolverVisitor :: solve_non_lin_system python exception: Sympify of expression 'could not parse 'is'' failed, because of exception being raised:
SyntaxError: invalid syntax (<string>, line 1)
[NMODL] [warning] :: SympySolverVisitor :: solve_non_lin_system python exception: Sympify of expression 'could not parse 'is'' failed, because of exception being raised:
SyntaxError: invalid syntax (<string>, line 1)
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[NMODL] [error] :: NeuronSolveVisitor :: solver method 'sparse' not supported
[...]

PR #1263, opened 2024-05-14
Fix bug with state named 'is'.

1uc:
Sympy doesn't tolerate variable names called `is`.

### DIFF
--- a/python/nmodl/ode.py
+++ b/python/nmodl/ode.py
@@ -30,20 +30,15 @@ if not ((major >= 1) and (minor >= 2)):
 # because sympy will try to use its own internal one; or error out for invalid variables.
 # Rename it before and after to a unique name.
 forbidden_var = [
-    # Python keywords
+    # Selected Python keywords
     "is",
     "as",
     "count",
-    "def",
     "del",
     "elif",
-    "finally",
-    "import",
     "in",
     "lambda",
     "pass",
-    "raise",
-    "yield",
     # SymPy functions
     "beta",
     "gamma",
--- a/python/nmodl/ode.py
+++ b/python/nmodl/ode.py
@@ -26,10 +26,27 @@ known_functions["abs"] = "fabs"
 if not ((major >= 1) and (minor >= 2)):
     raise ImportError(f"Requires SympPy version >= 1.2, found {major}.{minor}")
 
-# Some functions are protected inside sympy, if user has declared such a function, it will fail
-# because sympy will try to use its own internal one.
-# Rename it before and after to a single name
+# Some identifiers are protected inside sympy, if user has declared such a function, it will fail
+# because sympy will try to use its own internal one; or error out for invalid variables.
+# Rename it before and after to a unique name.
 forbidden_var = [
+    # Python keywords
+    "is",
+    "be",
+    "as",
+    "count",
+    "def",
+    "del",
+    "elif",
+    "finally",
+    "import",
+    "in",
+    "lambda",
+    "pass",
+    "raise",
+    "yield",
+
+    # SymPy functions
     "beta",
     "gamma",
     "uppergamma",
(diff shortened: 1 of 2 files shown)


1uc:
Fixes #1258.

python/nmodl/ode.py
```diff
@@ -26,10 +26,27 @@
...
     raise ImportError(f"Requires SympPy version >= 1.2, found {major}.{minor}")
 
-# Some functions are protected inside sympy, if user has declared such a function, it will fail
-# because sympy will try to use its own internal one.
-# Rename it before and after to a single name
+# Some identifiers are protected inside sympy, if user has declared such a function, it will fail
+# because sympy will try to use its own internal one; or error out for invalid variables.
+# Rename it before and after to a unique name.
 forbidden_var = [
+    # Python keywords
```
JCGoran, line 33: You can use https://docs.python.org/3/library/keyword.html#keyword.kwlist instead of hardcoding these.

1uc: i really don't want to.

JCGoran:
Will still break if the following keywords are declared as NMODL variables:

```plaintext
'nonlocal', 'except', 'else', 'from', 'not', 'with', 'for', 'or', 'try', 'and', 'class', 'None', 'True', 'async', 'continue', 'break', 'False', 'if', 'return', 'assert', 'await', 'global'
```

1uc:
I understand that it's rather partial solution. Some are tricky: should `if` be a legal variable name? I don't know and I want to punt. Same for `global`. I'd be quite happy if it continues to crash for all of those.

The problem is we only have strings. We don't know if a particular substring is a variable or if it's a keyword. We know if a particular string is used as a variable name; but we don't know if a particular occurrence of the string is a variable or a keyword.

test/integration/mod/variable_names.mod
```diff
@@ -0,0 +1,65 @@
...
+    ~ lambda <-> pass (1, 1)
+    ~ raise <-> yield (1, 1)
+    ~ alpha <-> beta (1, 1)
+    ~ gamma <-> delta (1, 1)
+    ~ epsilon <-> zeta (1, 1)
+    ~ eta <-> theta (1, 1)
+    ~ iota <-> kappa (1, 1)
+    ~ lambda <-> mu (1, 1)
+    ~ nu <-> xi (1, 1)
+    ~ omicron <-> pi (1, 1)
```
JCGoran, line 62: With `pi` commented out above, this doesn't compile under NOCMODL, is that intended (and do we care)?

JCGoran:
Re: legality of variable naming, shouldn't NOCMODL be the reference for what is valid and what isn't (however strange the rules may be)? In any case, the variables which NOCMODL doesn't seem to like (out of the ones I listed above) are apparently `else not for or and if return assert`, while the other ones compile fine, so maybe we can just add those in the list of variables to rename and call it a day for now.

test/integration/mod/variable_names.mod
1uc: No, that's a bug in NMODL and a mistake in the MOD file. `pi` is `3.14` and is not a valid name of a variable.

1uc:
Makes sense. The proper place to fix all of this is when NMODL create the string it sends to `SymPy`. That string should not contain any troublesome identifiers, i.e. all variable should probably be mangled first; and demangled after SymPy. There we'd have a better understanding of each substring.

1uc:
I'm surprised `class` and `try` would work.

JCGoran:
> I'm surprised class and try would work.

NOCMODL seems to `define` those, which is apparently legal (though it raises a warning on clang): https://godbolt.org/z/sa9qddYTY

1uc:
It feels very strange to allow C++ keywords as a variable names. It feels even more off to "fix" it by patching the string sent to SymPy, meaning the C++ keywords are returned from SymPy again.

I get your point of `nocmodl` being the reference. Since it's all not very clean, I've reduced the number of keywords we sanitize in this unsatisfactory way to a small number of what I believe a plausible variable names. I'd be open to reducing what works even futher to just `is` and maybe `in`, to make the bug Koen is seeing go away.

JCGoran:
Since keeping a list of both C++ and Python keywords is a bit cumbersome, we could add prefix/suffix to all variables internally or something, but we can postpone the details of that discussion for later. For now, as long as this PR fixes the issue, I'm fine with it as a temporary workaround.

JCGoran (approved):

pramodk (approved):
LGTM

pramodk (approved):

PR #1271, opened 2024-05-21
Only build min and max Python version wheels on PRs

JCGoran:
To speed-up the CI on PRs, we only build wheels for the min (currently 3.8) and max (currently 3.12) versions of Python.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -201,37 +201,32 @@ stages:
       vmImage: 'ubuntu-20.04'
     steps:
     - task: UsePythonVersion@0
+
     - checkout: self
       submodules: True
       condition: succeeded()
+
     - script: |
         if [[ "$(RELEASEWHEELBUILD)" != "True" ]]; then
-          export TAG="-nightly"
+          echo "##vso[task.setvariable variable=TAG;]-nightly"
         else
-          export TAG=""
+          echo "##vso[task.setvariable variable=TAG;]"
         fi
+      displayName: "Set wheel tag"
+
+    - bash: echo "##vso[task.setvariable variable=CIBW_BUILD;]cp38* cp312*"
+      condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
+      displayName: "Set build identifiers"
+
+    - script: |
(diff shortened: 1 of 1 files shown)


1uc:
It wasn't the last piece of CI to go green anymore.

azure-pipelines.yml
```diff
@@ -216,8 +216,23 @@ stages:
...
+          export TAG="-nightly"
+        else
+          export TAG=""
+        fi
+        python3 -m pip install --upgrade pip
+        python3 -m pip install cibuildwheel==2.16.5 tomli tomli-w
+        # change the name accordingly
+        python3 packaging/change_name.py pyproject.toml "NMODL${TAG}"
+        export SETUPTOOLS_SCM_PRETEND_VERSION="$(git describe --tags | cut -d '-' -f 1,2 | tr - .)"
+        CIBW_BUILD="cp38* cp312*" CIBW_ENVIRONMENT_PASS_LINUX='SETUPTOOLS_SCM_PRETEND_VERSION' python3 -m cibuildwheel --output-dir wheelhouse
```
1uc, line 232: Can one reduce the duplication by setting `CIBW_BUILD` to one value for nightly and another value for regular PR CI runs?

1uc (approved):

azure-pipelines.yml
JCGoran: I think I removed most of the duplication now, with two things to keep in mind (i.e. an explanation why it works):
- any azure variable not set, but used later (via `$(VARIABLE)`) will expand to an empty string
- if `CIBW_BUILD` is set to the empty string, `cibuildwheel` will ignore it and build all configurations (which is what we want for non-PRs)

1uc (approved):
Is there a particular advantage of using `###vso[task....]` syntax over the standard bash command `export`?

JCGoran:
I just read a bit on https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#code-try-0 and `vso[...]` was shown as the way to set variables. `export` is nowhere to be found there, so I don't know if it works when you have multiple steps in a single job.

PR #1272, opened 2024-05-22
Move POINT_PROCESS and TQITEM code.

1uc:
This code is shared between NEURON and CoreNEURON. Therefore, it's moved the common base class.

### DIFF
--- a/src/codegen/codegen_cpp_visitor.hpp
+++ b/src/codegen/codegen_cpp_visitor.hpp
@@ -842,6 +842,15 @@ class CodegenCppVisitor: public visitor::ConstAstVisitor {
     /*                             Code-specific helper routines                            */
     /****************************************************************************************/
 
+    /**
+     * Add the variable tqitem during `get_int_variables`.
+     */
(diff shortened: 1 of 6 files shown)


src/codegen/codegen_cpp_visitor.hpp
```diff
@@ -842,6 +842,15 @@ class CodegenCppVisitor: public visitor::ConstAstVisitor {
...
 
+    /**
+     * Add the variable tqitem during `get_int_variables`.
+     */
+    virtual void add_variable_tqitem(std::vector<IndexVariableInfo>& variables) = 0;
+
+    /**
+     * Add the variable point_process during `get_int_variables`.
+     */
+    virtual void add_variable_point_process(std::vector<IndexVariableInfo>& variables) = 0;
```
JCGoran, line 853: Question: do we have any other code generators that inherit directly from `CodegenCppVisitor` (other than the NEURON and coreNEURON ones)? If we don't, LGTM.

1uc: this is the "correct" place to lift it to.

JCGoran (approved):

PR #1273, opened 2024-05-22
Fix ARTIFICIAL_CELL net_send.

1uc:


test/usecases/net_send/simulate.py
```diff
@@ -3,23 +3,51 @@
...
+    y = np.array(y_hoc.as_numpy())
+    art_y = np.array(art_y_hoc.as_numpy())
 
-assert np.all(np.abs(y - y_exact) == 0), f"{y} != {y_exact}, delta: {y - y_exact}"
+    return t, y, art_y
+
+
+def check_solution(t, y, art_y):
+    y_exact = np.array(t >= 2.0, np.float64)
+    art_y_exact = np.array(t >= 1.5, np.float64)
```
JCGoran, line 33: What do the two numbers here (2.0 and 1.5) represent? Are they related to some of the values used in the mod file?

1uc: I've added a brief description to the top. The two numbers are the time at which the event should be received. At that point `y` toggles "on".

JCGoran: Ahh, so the time is basically the first argument to `net_send` in the mod file?

JCGoran (approved):
LGTM overall, only minor comments.

test/usecases/net_send/simulate.py
1uc: Correct, though technically, its the delay from the current time.

PR #1278, opened 2024-05-24
Fix recursive FUNCTION calls

JCGoran:
The following mod file does not compile under both NEURON and coreNEURON codegen:

```plaintext
NEURON {
    SUFFIX recursion
}
FUNCTION f(n) {
    f = f(n)
}
```
Some more examples are added in `recursion.mod`, which should cover all of the cases of interest (note that only some of them are tested for correctness since the convoluted ones don't have a terminating condition, but they should all at least compile).

### DIFF
--- a/test/usecases/recursion/recursion.mod
+++ b/test/usecases/recursion/recursion.mod
@@ -2,21 +2,6 @@ NEURON {
     SUFFIX recursion
 }
 
-: single recursive call
-FUNCTION f_simple(n) {
-    f_simple = f_simple(n)
-}
-
-: multiple recursive calls
-FUNCTION fc(n) {
-    fc = fc(fc(fc(fc(n))))
-}
-
-: multiple recursive calls with different args
-FUNCTION fd(n) {
-    fd = fd(fc(fd(n) + fc(fd(n))))
-}
-
 FUNCTION myfactorial(n) {
     if (n == 0 || n == 1) {
         myfactorial = 1
@@ -24,11 +9,3 @@ FUNCTION myfactorial(n) {
         myfactorial = n * myfactorial(n - 1)
     }
 } 
-
-FUNCTION myfib(n) {
-    if (n == 0 || n == 1){
-        myfib = n
-    } else {
-        myfib = myfib(n - 1) + myfib(n - 2)
-    }
-}
--- a/test/usecases/recursion/recursion.mod
+++ b/test/usecases/recursion/recursion.mod
@@ -1,0 +1,11 @@
+NEURON {
+        SUFFIX recursion
+}
+
+FUNCTION myfactorial(n) {
+    if (n == 0 || n == 1) {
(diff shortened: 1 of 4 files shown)


test/usecases/recursion/recursion.mod
```diff
@@ -0,0 +1,34 @@
...
+}
+
+: single recursive call
+FUNCTION f_simple(n) {
+    f_simple = f_simple(n)
+}
+
+: multiple recursive calls
+FUNCTION fc(n) {
+    fc = fc(fc(fc(fc(n))))
```
1uc, line 12: This looks like infinite recursion. Why do you want to assert that this is valid?

test/usecases/recursion/recursion.mod
```diff
@@ -0,0 +1,34 @@
...
+}
+
+: multiple recursive calls with different args
+FUNCTION fd(n) {
+    fd = fd(fc(fd(n) + fc(fd(n))))
```
1uc, line 17: Every call to `fd` calls `fd` 3 times; and `fc` (in itself infinitely recursing). Hence, same question what does this test?

test/usecases/recursion/recursion.mod
```diff
@@ -0,0 +1,34 @@
+NEURON {
+    SUFFIX recursion
+}
+
+: single recursive call
+FUNCTION f_simple(n) {
+    f_simple = f_simple(n)
```
1uc, line 7: Same here.

1uc:
Nice, what's the purpose of testing both `fib` and `factorial`? How do they differ and enhance eachother?

test/usecases/recursion/recursion.mod
JCGoran: As mentioned, this is here to make sure that it actually compiles, there are no run-time guarantees (hence it's not present in `simulate.py`).

test/usecases/recursion/recursion.mod
JCGoran: Different arguments mainly, it's again there to make sure that NMODL compiles the mod file successfully (whether or not it's a good idea to actually run this function is a different issue). I put several tests just to see if I can break the codegen somehow.

test/usecases/recursion/recursion.mod
1uc: Why do we want it to compile? I'd argue that we should be allowed to at least make an attempt to reject such MOD files (yes, yes halting-problem bla bla). This leaves a trail suggesting that this is essential functionality. Which is plausible, because there might be some MOD file that has an `fc` and like here never calls it.

Second issue I have: If you can't test the behaviour of the code how do you know that whatever bug you're trying to squash got squashed? (NMODL could simply replace it with a `""` and things would compile.)

1uc (approved):

PR #1280, opened 2024-05-29
Fix PROCEDUREs using TABLEs containing the same var

JCGoran:
Also adds a test to it.

### DIFF
--- a/src/visitors/semantic_analysis_visitor.cpp
+++ b/src/visitors/semantic_analysis_visitor.cpp
@@ -37,6 +37,26 @@ bool SemanticAnalysisVisitor::check(const ast::Program& node) {
     }
     /// -->
 
+    // check that we do not have any duplicate TABLE statement variables in PROCEDUREs
+    const auto& procedure_nodes = collect_nodes(node, {ast::AstNodeType::PROCEDURE_BLOCK});
+    std::unordered_set<std::string> procedure_vars{};
+    for (const auto& proc_node: procedure_nodes) {
+        const auto& table_nodes = collect_nodes(*proc_node, {ast::AstNodeType::TABLE_STATEMENT});
+        for (const auto& table_node: table_nodes) {
+            const auto& table_vars =
+                std::dynamic_pointer_cast<const ast::TableStatement>(table_node)->get_table_vars();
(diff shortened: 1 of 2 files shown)


test/unit/visitor/semantic_analysis.cpp
```diff
@@ -65,6 +65,21 @@ SCENARIO("TABLE stmt", "[visitor][semantic_analysis]") {
...
+                TABLE var FROM 0 TO 1 WITH 10
+                var = arg
+            }
+            PROCEDURE p2(arg) {
+                TABLE var FROM 0 TO 1 WITH 10
+                var = 2 * arg
+            }
+        )";
+        THEN("fail") {
+            REQUIRE(run_semantic_analysis_visitor(nmodl_text));
```
1uc, line 80:
```suggestion
            # This is asserted because `nocmodl` generated code fails to
            # compile in these cases, not because it's must not work.
            REQUIRE(run_semantic_analysis_visitor(nmodl_text));
```

src/visitors/semantic_analysis_visitor.cpp
```diff
@@ -37,6 +37,26 @@ bool SemanticAnalysisVisitor::check(const ast::Program& node) {
...
+    // check that we do not have any duplicate TABLE statement variables in PROCEDUREs
+    const auto& procedure_nodes = collect_nodes(node, {ast::AstNodeType::PROCEDURE_BLOCK});
+    std::unordered_set<std::string> procedure_vars{};
+    for (const auto& proc_node: procedure_nodes) {
+        const auto& table_nodes = collect_nodes(*proc_node, {ast::AstNodeType::TABLE_STATEMENT});
+        for (const auto& table_node: table_nodes) {
+            const auto& table_vars =
+                std::dynamic_pointer_cast<const ast::TableStatement>(table_node)->get_table_vars();
+            for (const auto& table_var: table_vars) {
+                const auto& result = procedure_vars.insert(table_var->get_node_name());
```
1uc, line 49:
```suggestion
                auto [var_name, inserted] = procedure_vars.insert(table_var->get_node_name());
```
I think we might be allowed to use a structured binding.

1uc (approved):

PR #1285, opened 2024-05-29
Implement GLOBALs.

1uc:


test/usecases/generate_references.py
```diff
@@ -0,0 +1,68 @@
...
+
+    nmodl = sys.argv[1]
+    usecase_dir = sys.argv[2]
+    output_basedir = sys.argv[3]
+    script_dir = os.path.dirname(sys.argv[0])
+
+    config = load_config(script_dir=script_dir, usecase_dir=usecase_dir)
+    for c in config:
+        output_dir = os.path.join(output_basedir, c["output_reldir"])
+        generate_references(nmodl, usecase_dir, output_dir, c["nmodl_flags"])
```
pramodk, line 68: didn't look thoroughly for this file but looks ok to me!

test/usecases/global/references.json
```diff
@@ -0,0 +1,3 @@
+[
+    { "output_reldir": "neuron", "nmodl_flags": ["--neuron"]}
```
pramodk, line 2: hoping we will need to care about only a finite number of combinations! (i.e. will remove the need of various combinations)

test/usecases/global/simulate.py
```diff
@@ -0,0 +1,79 @@
...
+assert np.all(y0[i2] == 33.3)
+assert np.all(y0[i3] == z0)
+
+# The values on thread 1:
+assert y1[i0] == g_w_init
+# The value of `g_w` on thread 1 is not specified, after setting
+# `g_w` from Python.
+# assert np.all(y0[i1] == g_w1)
+assert np.all(y1[i2] == 34.3)
+assert np.all(y1[i3] == z1)
```
pramodk, line 79: must say these use case examples are very good for understanding neuron's existing working! 👌

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -556,6 +556,32 @@ std::string CodegenNeuronCppVisitor::int_variable_name(const IndexVariableInfo&
...
 
 
+std::string CodegenNeuronCppVisitor::thread_variable_name(const ThreadVariableInfo& var_info,
+                                                          bool use_instance) const {
+    auto i_var = var_info.offset;
+    auto var_name = var_info.symbol->get_name();
+    std::string simd_width = "1";
+
+    // std::string inst_id = fmt::format("(id % {})", simd_width);
+    // std::string inst_id = "0";
```
pramodk, line 566: adding a comment just as a reminder

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1090,9 +1147,50 @@ void CodegenNeuronCppVisitor::print_mechanism_register() {
...
+    printer->pop_block();
+    printer->pop_block();
+
+    printer->push_block("static void thread_mem_cleanup(Datum* _thread) ");
+    printer->fmt_line("double * _thread_data_ptr = _thread[{}].get<double*>();", thread_data_index);
+    printer->push_block(fmt::format("if(_thread_data_ptr == {})", static_thread_data));
+    printer->fmt_line("{} = 0;", inuse);
+    printer->pop_block();
+    printer->push_block("else");
+    printer->add_line("delete[] _thread_data_ptr;");
```
pramodk, line 1188: this is just a question rather than a review comment: from the old times I was wondering if we could make some of such clean-up routines redundant (i.e. replace them with smart ptrs or something).

Again, just saying if you see opportunities to change/remove old behavior, feel free to do so!

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1225,6 +1323,37 @@ void CodegenNeuronCppVisitor::print_make_node_data() const {
...
+    if (codegen_thread_variables.empty()) {
+        return;
+    }
+
+    printer->add_newline(2);
+    printer->fmt_push_block("struct {} ", thread_variables_struct());
+    printer->add_line("double * thread_data;");
+    printer->add_newline();
+
+    std::string simd_width = "1";
```
pramodk, line 1336: IIUC, currently this is always 1 but in the future something we can parameterise based on the execution environment / target.

pramodk (approved):
LGTM.

There are some comments just for clarification!

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Not immediately obvious because it needs to also be able to point to the static copy of the thread variable. It is however, another occurrence of a owning container that can act as a few.

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: ty, gone

PR #1286, opened 2024-05-30
Make `generate_references.py` configurable.

1uc:
Adds the ability to customize with which flags `nmodl` is run to generate the references.

The default is `""` (for CoreNEURON) and `"--neuron"` (for NEURON). These choices can be overridden by adding a file `references.json` to the usecase directory.

Since the logic is slightly more involved, the Bash script is reimplemented as a Python script.


1uc:
Can be seen in use here: https://github.com/BlueBrain/nmodl/pull/1285

test/usecases/generate_references.py
```diff
@@ -0,0 +1,68 @@
...
+
+    nmodl = sys.argv[1]
+    usecase_dir = sys.argv[2]
+    output_basedir = sys.argv[3]
+    script_dir = os.path.dirname(sys.argv[0])
+
+    config = load_config(script_dir=script_dir, usecase_dir=usecase_dir)
+    for c in config:
+        output_dir = os.path.join(output_basedir, c["output_reldir"])
+        generate_references(nmodl, usecase_dir, output_dir, c["nmodl_flags"])
```
pramodk, line 68: looks reasonable to me!

pramodk (approved):

PR #1288, opened 2024-05-31
Add CMake export infrastructure

matz-e:
Currently will export a target `nmodl::nmodl` for the NMODL binary, and a `nmodl_BINARY` variable for the filename.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -24,6 +24,7 @@ include_directories(${PYBIND11_INCLUDE_DIR} ${PYTHON_INCLUDE_DIRS})
 # Add executables
 # =============================================================================
 add_executable(nmodl main.cpp)
+set_target_properties(nmodl PROPERTIES PUBLIC_HEADER nmodl.hpp)
 target_link_libraries(
   nmodl
   CLI11::CLI11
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -38,7 +39,7 @@ add_dependencies(nmodl nmodl_copy_python_files nmodl_copy_solver_files)
 cpp_cc_configure_sanitizers(TARGET nmodl)
 
 # =============================================================================
-# Add dependency with nmodl pytnon module (for consumer projects)
+# Add dependency with nmodl python module (for consumer projects)
```
1uc, line 42:
```suggestion
# Add dependency with nmodl Python module (for consumer projects)
```

src/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ include_directories(${PYBIND11_INCLUDE_DIR} ${PYTHON_INCLUDE_DIRS})
 # Add executables
 # =============================================================================
 add_executable(nmodl main.cpp)
+set_target_properties(nmodl PROPERTIES PUBLIC_HEADER nmodl.hpp)
```
1uc, line 27: I see it used to get installed and I like one change at a time, so no need for changes. However, did you happen to find out why we need this? To my knowledge NMODL is an application (not library) and a Python module. Hence, it's not obvious to me why we need this.

1uc (commented):
Have you tested integration into NRN?

src/CMakeLists.txt
matz-e: Indeed, it seems to be only used by the unit tests. I can try to remove it, together with adding some "legacy" glue.

1uc: No legacy glue please.

matz-e:
Tested in the NRN repo, seems to run fine.

1uc (approved):

PR #1289, opened 2024-06-03
Use `ParamVector` everywhere

JCGoran:
Also refactored some functions so they don't use `char*` but a `std::string` instead.

### DIFF
--- a/src/codegen/codegen_coreneuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_coreneuron_cpp_visitor.cpp
@@ -721,7 +721,7 @@ CodegenCoreneuronCppVisitor::ParamVector CodegenCoreneuronCppVisitor::internal_m
 
 
 const std::string CodegenCoreneuronCppVisitor::external_method_arguments() noexcept {
-    return "id, pnodecount, data, indexes, thread, nt, ml, v";
+    return get_arg_str(external_method_parameters());
 }
 
 
--- a/src/codegen/codegen_coreneuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_coreneuron_cpp_visitor.cpp
@@ -39,6 +39,8 @@ using visitor::RenameVisitor;
 using visitor::SymtabVisitor;
 using visitor::VarUsageVisitor;
 
+using ParamVector = std::vector<std::tuple<std::string, std::string, std::string, std::string>>;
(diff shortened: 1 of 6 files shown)


src/codegen/codegen_cpp_visitor.cpp
```diff
@@ -32,6 +32,20 @@ bool CodegenCppVisitor::ion_variable_struct_required() const {
...
+    std::string str;
+    bool is_first = true;
+    for (const auto& param: params) {
+        if (is_first) {
+            is_first = false;
+        } else {
+            str += ", ";
+        }
+        str += std::get<3>(param);
+    }
```
1uc, line 45: You can use `fmt::join`.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -31,6 +31,7 @@ using visitor::RenameVisitor;
 using visitor::VarUsageVisitor;
 
 using symtab::syminfo::NmodlType;
+using ParamVector = std::vector<std::tuple<std::string, std::string, std::string, std::string>>;
```
1uc, line 34: Please remove this. `ParamVector` should be correctly defined in a header.

src/codegen/codegen_coreneuron_cpp_visitor.cpp
```diff
@@ -39,6 +39,8 @@ using visitor::RenameVisitor;
 using visitor::SymtabVisitor;
 using visitor::VarUsageVisitor;
 
+using ParamVector = std::vector<std::tuple<std::string, std::string, std::string, std::string>>;
```
1uc, line 42: Please remove this.

src/codegen/codegen_neuron_cpp_visitor.cpp
JCGoran: For some reason, when I remove this, or even just put it in the header file only, the compilation fails (clang 15.0.0 on MacOS), even though it should be visible due to this snippet:
https://github.com/BlueBrain/nmodl/blob/5366532ee2ca2f4961e635672f18043c69e23443/src/codegen/codegen_cpp_visitor.hpp#L266

Anything obvious that I'm missing here?

alkino (approved):

PR #1291, opened 2024-06-03
Fix missing null-terminator.

1uc:


1uc:
Would have been detectable with ASAN enabled version of NRN.

JCGoran (commented):
This fixes a memory leak? If so, what's a MWE that would demonstrate the problem?

1uc:
> This fixes a memory leak? If so, what's a MWE that would demonstrate the problem?

No, it fixes an out-of-bounds access. It's covered many times already; but without an NRN with ASAN enabled (or similar) there's no guarantee that it results in a failure (e.g. SEGFAULT).

JCGoran (approved):
LGTM then!

PR #1294, opened 2024-06-03
Cleanup 'usecases'.

1uc:
The new structure is to have one directory per keyword. Tests can be organized in separate test files, named `test_*.py` (or `simulate.py` if that's a better name).

Numerous tests have been polished to make them easier to read, using better names and functions to extract common functionality.


JCGoran (approved):
LGTM, nice QoL improvement.

Issue #1290, opened 2024-06-03
Fix `ctest --parallel`


1uc:
Running `ctest --parallel` reliably fails with:
```
Filters: "Scenario: Check parents in valid NMODL constructs"
Randomness seeded to: 515386643

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
testparser is a Catch2 v3.5.3 host application.
Run with -? for options

-------------------------------------------------------------------------------
Scenario: Check parents in valid NMODL constructs
-------------------------------------------------------------------------------
/home/lucg/git/bbp/nmodl/test/unit/parser/parser.cpp:200
...............................................................................

/home/lucg/git/bbp/nmodl/test/unit/parser/parser.cpp:207: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  NMODL Parser Error : can not open file : "Unit.inc" [Location : 2.25-34]

-------------------------------------------------------------------------------
Scenario: Check parents in valid NMODL constructs
-------------------------------------------------------------------------------
/home/lucg/git/bbp/nmodl/test/unit/parser/parser.cpp:200
...............................................................................

/home/lucg/git/bbp/nmodl/test/unit/parser/parser.cpp:207: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  NMODL Parser Error : can not open file : "Unit.inc" [Location : 2.25-34]

===============================================================================
test cases:  1 |  0 passed | 1 failed
assertions: 78 | 76 passed | 2 failed
```

JCGoran:
Since #1295 was merged, can this issue be closed?

PR #1295, opened 2024-06-03
Remove generation of INCLUDE file in the tests

pramodk:
- include mod files generation cause race condition with parallel tests, see #1290
- the included files are trivial examples and they don't need to be generated from the code
  - could be copied via configure_file
  - cmake could generate those at a configure time

Considering the use case and it's triviality, I thought 2nd option is better. That also allows removal of some extra code/logic.


alkino (approved):

1uc (approved):
It ran several times without issues.

PR #1301, opened 2024-06-05
Make usecases runnable under pytest

JCGoran:
This is based on #1294 so I don't need to solve future merge conflicts.

To outline how it works:

- all Python files need to have the prefix `test` (hence `simulate.py` becomes `test_sim.py` or whatever) so pytest can pick them up
- all Python objects we want to test using pytest need to also have the prefix `test`
- the `usecases` dir needs an `__init__.py` (because pytest makes it basically impossible to use something like fixtures) since that's where we define a common function for compiling mod files, `compile_modfile`
- due to Python's import mechanics, each usecase itself needs an `__init__.py` (can be empty, but the file must exist) so we can actually import the `compile_modfile` function
- we parametrize two cases, one with NOCMODL (default), and one with NMODL (for which we need to specify a path to the executable, hence the arg is either `None` or some `Path` instance), and run those in turn (as a discussion point, should the argument be just `nmodl`/`nocmodl`, and then it looks up the location dynamically? Then we can parametrize the function with just `['nocmodl', 'nmodl']` rather than the more verbose `[None, shutil.which("nmodl")]` which requires remembering to import the `shutil` module)
- if the file is compiled successfully, we use `load_mechanisms` to actually make it visible for NEURON (otherwise `insert` fails)


1uc:
As discussed offline, the simplicity of simply having a mod file and an associated script is valuable because it mixes well with many important debugging tools. It also allows the usecases to double as examples. Finally, we often test that things are in the correct global state, before running "init". Therefore, we need to run tests in separate processes.

PR #1305, opened 2024-06-06
Make `nmodl` independent of `nmodl`.

1uc:
The executable from the Python library.

### DIFF
--- a/src/pybind/ode_py.hpp.inc
+++ b/src/pybind/ode_py.hpp.inc
@@ -1,0 +1,19 @@
+#include <string>
+
+// This file is generated from `ode.py`.
+//
+// During code-generation NMODL needs to call SymPy, e.g. to compute
+// derivatives symbolically. The code to do this can be found in `ode.py`.
+//
+// To avoid a dependency of the `nmodl` binary on the Python library `nmodl`.
+// We embed `ode.py` like this.
+//
+// However, because we want to be able to test `ode.py` via pytest we can't
+// move it here.
+
+namespace nmodl::pybind_wrappers {
+const std::string ode_py = R"jiowi(
+@NMODL_ODE_PY@
+)jiowi";
+
+}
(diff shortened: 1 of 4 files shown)


src/pybind/ode_py.hpp.inc
```diff
@@ -0,0 +1,19 @@
...
+// derivatives symbolically. The code to do this can be found in `ode.py`.
+//
+// To avoid a dependency of the `nmodl` binary on the Python library `nmodl`.
+// We embed `ode.py` like this.
+//
+// However, because we want to be able to test `ode.py` via pytest we can't
+// move it here.
+
+namespace nmodl::pybind_wrappers {
+const std::string ode_py = R"jiowi(
```
matz-e, line 15: What's `jiowi`?

matz-e (approved):

src/pybind/ode_py.hpp.inc
```diff
@@ -0,0 +1,19 @@
...
+@NMODL_ODE_PY@
```
pramodk, line 16: now its clear what you were referring to other day! 👍

src/pybind/wrapper.cpp
```diff
@@ -198,11 +185,6 @@ void destroy_ads_executor_func(AnalyticDiffExecutor* exec) {
 
 void initialize_interpreter_func() {
     pybind11::initialize_interpreter(true);
-    const auto python_path_cstr = std::getenv("PYTHONPATH");
-    if (python_path_cstr) {
-        pybind11::module::import("sys").attr("path").cast<pybind11::list>().insert(
-            0, python_path_cstr);
-    }
```
pramodk, line 205: Just for my own understanding: if a user has sympy installed in some external location, let's say as a module `py-sympy`, having that in `PYTHONPATH`, everything will still work? (or do you still see any issues?)

src/pybind/ode_py.hpp.inc
1uc: Some arbitraty string that's unlikely to appear inside `ode.py`.

PR #1306, opened 2024-06-07
Modernize NMODL -> Python layer.

1uc:
The changes are:
* Don't forcefully set the PYTHONPATH.
* Replace false object oriented code with a function (each).
* Replace a global variable, with a function to initialize.
* Use a reference for a pointer that's should never be a `nullptr`.


pramodk (approved):
I didn't check every single changed line from `src/pybind/wrapper.cpp` but overall these are certainly nice improvements and make it clear/concise!

1uc:
Double checked the issue about `PYTHONPATH` and at least locally it's not required. Plus it should get tests via the Spack-powered CI (since everything is then added via PYTHONPATH).

PR #1309, opened 2024-06-11
Consistently use `_lmc` and references.

1uc:
Frequently `_lmr` is passed by pointer, which requires printing:
```
  foo(&_lmr, ...);
```
the `&` is annoying, because it prevents us from using `get_arg_str`. It
also sometimes leads us to do:
```
  auto * _ml = &_lmr;
```
which is bad, because it's not a `Memb_list`.

In a second step we rename `_lmr` to `_lmc` which is short for `mechanism_cache` which is better since the `_range` part can also be `_instance`.


1uc:
In `nocmodl` it's local. Feel free to change it if it bothers you. Prefixing with `_` help avoid clashes, in NMODL, it's a bit murky since, we rarely print variables just plain, usually it's `inst.{}`.

JCGoran (approved):

PR #1311, opened 2024-06-11
Fix calls from NET_RECEIVE.

1uc:


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -2129,6 +2129,10 @@ void CodegenNeuronCppVisitor::print_net_receive() {
     printer->add_line("auto * _ppvar = _nrn_mechanism_access_dparam(_pnt->prop);");
 
     printer->fmt_line("auto inst = make_instance_{}(_lmc);", info.mod_suffix);
+    printer->fmt_line("// nocmodl has a nullptr dereference for thread variables.");
+    printer->fmt_line("// NMODL will fail to compile at a later point, because of");
+    printer->fmt_line("// missing '_thread_vars'.");
```
JCGoran, line 2134: This can be a comment in NMODL, not the generated NMODL code.

1uc: The point is to avoid a panic attack when reading the generated code.

pramodk: No strong opinion on this. From users point of view,  until now we treated generated code as a "black box".  If we start adding more explanatory notes in the generated code, for potential debugging or understanding things, it's OK/better as well! (for devs as well)

pramodk (approved):
LGTM

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: It's only for ourselves. I can absolutely see myself panicking at my earlier stupidity if I see:
```
double * _thread = nullptr;
foo(..., _thread, ...);
```

PR #1312, opened 2024-06-12
Lift shared code for KINETIC.

1uc:


test/usecases/kinetic/X2Y.mod
```diff
@@ -0,0 +1,30 @@
...
+}
+
+INITIAL {
+    X = 0
+    Y = 1
+}
+
+KINETIC state {
+    ~ X <-> Y (0.4, 0.5)
+    i = (f_flux-b_flux)
```
JCGoran, line 29: I don't see `b/f_flux` anywhere in the generated code (or something close to those var names) so I'm curious, how does this work? Is this an internal NEURON var?

1uc: In the line above there's material flowing from the left to the right (and vice-versa), the amount is available via `{f,b}_flux`.

JCGoran: Ah, found some docs online: https://www.neuron.yale.edu/neuron/static/docs/refman/kinetic.html
Unrelated, I feel like this should be part of the RTD docs because the [search](https://nrn.readthedocs.io/en/latest/search.html?q=f_flux&check_keywords=yes&area=default) reveals nothing useful...

JCGoran (approved):

PR #1313, opened 2024-06-12
Fix integration tests.

1uc:
There must not be a clash of FUNCTION and RANGE names.


JCGoran (approved):
For context, `nrnivmodl` fails with:

```plaintext
Error: alfa used as both variable and function in file glia_sparse.mod
```
while `nmodl --neuron` generates the file without issues, which it shouldn't.

1uc:
There's (now) a ticket for the issue: #1316

PR #1315, opened 2024-06-12
Implement top-level LOCAL variables.

1uc:
These are global variables, and when THREAD_SAFE act like thread variables, without being exposed to HOC/Python.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -859,11 +859,6 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
     }
 
     if (!info.thread_variables.empty()) {
-        printer->fmt_line("int thread_data_in_use{};", value_initialize);
-        printer->fmt_line("{} thread_data[{}] /* TODO init thread_data */;",
-                          float_type,
-                          info.thread_var_data_size);
-
         size_t prefix_sum = 0;
         for (size_t i = 0; i < info.thread_variables.size(); ++i) {
             const auto& var = info.thread_variables[i];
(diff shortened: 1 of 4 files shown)


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -859,22 +859,35 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
...
+            codegen_thread_variables.push_back({var, n_thread_vars + i, prefix_sum});
+
+            prefix_sum += var->get_length();
+        }
+    }
+
+    if (!codegen_thread_variables.empty()) {
+        auto thread_data_size = info.thread_var_data_size + info.top_local_thread_size;
+        printer->fmt_line("int thread_data_in_use{};", value_initialize);
+        printer->fmt_line("{} thread_data[{}];", float_type, thread_data_size);
```
pramodk, line 885: I was just reminding myself about our [slack discussion](https://bluebrainproject.slack.com/archives/C0237TJ7T1U/p1716895624761059?thread_ts=1716887069.202769&cid=C0237TJ7T1U). IIUC, our approach going forward would be to treat such usages as user trying to use these variables as "scratch pad memory". Internally NMODL will decide how they should be implemented for different backends. For example, CPU-SIMD backend should consider vector width, GPU backend should consider shared memory with warp/thread-blocks etc etc. Right?

pramodk (approved):
LGTM.
(I added a comment just to have future notes (as slack history will disappear one day))

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Yes, I can take the time to document this properly.

Issue #1279, opened 2024-05-27
Expand `blame` stack trace


JCGoran:
This mod file:

```plaintext
NEURON {
    SUFFIX recursion
}

FUNCTION f(n) {
    f = f(n)
}
```

provides the following stack trace when running `nmodl --neuron [MODFILE] blame --line 206`:

```cpp
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 546, in print_statement_block
       545:              !statement->is_mutex_unlock() && !statement->is_protect_statement()) {
    >  546:              printer->add_indent();
       547:          }
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 623, in visit_string
       622:      }
    >  623:      printer->add_text(name);
       624:  }
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 776, in visit_binary_expression
       775:          lhs->accept(*this);
    >  776:          printer->add_text(" " + op + " ");
       777:          rhs->accept(*this);
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 466, in print_function_call
       465:      const auto& arguments = node.get_arguments();
    >  466:      printer->add_text(function_name, '(');
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.hpp", line 1396, in print_vector_elements<std::shared_ptr<nmodl::
ast::Expression> >
      1395:      for (auto iter = elements.begin(); iter != elements.end(); iter++) {
    > 1396:          printer->add_text(prefix);
      1397:          (*iter)->accept(*this);
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 623, in visit_string
       622:      }
    >  623:      printer->add_text(name);
       624:  }
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 482, in print_function_call
       481:      print_vector_elements(arguments, ", ");
    >  482:      printer->add_text(')');
       483:  }
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 550, in print_statement_block
       549:          if (need_semicolon(*statement)) {
    >  550:              printer->add_text(';');
       551:          }
    Source: "/nmodl/src/codegen/codegen_cpp_visitor.cpp", line 553, in print_statement_block
       552:          if (!statement->is_mutex_lock() && !statement->is_mutex_unlock()) {
    >  553:              printer->add_newline();
[...]

PR #1320, opened 2024-06-12
Detailed blame.

1uc:


JCGoran:
I seem to be getting additional backtrace output sometimes (tested on BB5 with file from #1279 and command `nmodl --neuron [modfile] blame --line 208 --detailed`):

```cpp
28:    Source: "", line 0, in
27:    Source: "", line 0, in _start
26:    Source: "", line 0, in
25:    Source: "/gpfs/bbp.cscs.ch/data/project/proj16/jelic/nmodl/src/main.cpp", line 589, in main
```

Can the first 3 in the above be removed?

1uc:
We could, though the more magic we add the more likely it is to break on us; and it's just a debugging assistant.

Maybe I'm too complacent, but it feels like there's often junk at the top of a backtrace. In fact after writing the whole reply I noticed, that I have 4 lines of spam. They're definitely part of a backtrace I guess it helps identify the beginning of a backtrace :)

If we want to make it more compact there's better ways: The backtraces for each of the commands that lead to a string being printed to the requested line, often have a lot in common: a) they share most of the outer traces; b) `fmt_line` is gonna create 3 backtraces, one for printing the indent, one of interest, and another for printing the `"\n"` (annoyingly the difference gets chopped off, because it's too hard to find the interesting line otherwise).

JCGoran (approved):
LGTM

PR #1322, opened 2024-06-17
Pin numpy max version to <2

JCGoran:
NMODL counterpart of https://github.com/neuronsimulator/nrn/pull/2916. Since NEURON first installs its own Python requirements, and only afterwards the requirements of its external dependencies, we need to merge this one first, then update and merge https://github.com/neuronsimulator/nrn/pull/2900, and finally merge https://github.com/neuronsimulator/nrn/pull/2916.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dependencies = [
 scripts = {nmodl = "nmodl._binwrapper:main"}
 requires-python = ">=3.8"
 
-optional-dependencies.test = ["pytest>=3.3.0", "pytest-cov", "numpy", "scipy"]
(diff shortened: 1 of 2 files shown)


1uc:
We can test this by updating the submodule in https://github.com/neuronsimulator/nrn/pull/2916 to this commit. If that passes we know we can merge this, update the submodule on the Python side again and merge that commit. Then close https://github.com/neuronsimulator/nrn/pull/2900.

JCGoran:
Since NEURON didn't require patching, and it seems to work fine with NumPy 2, I'm closing this as I don't see any issues with using it.

PR #1325, opened 2024-06-17
Fix implicit argument handling

JCGoran:
`at_time`, an internal NEURON function, requires an instance of `NrnThread*`, which is implicit in the mod file, but was previously called with a non-existing argument name (parent function had `NrnThread* _nt`, while the call to `at_time` had passed `nt`).

Note that I am slightly annoyed at the possible fixes here: in NEURON codegen, all of the arguments passed to functions apparently start with an underscore, while for coreNEURON codegen all of them start without an underscore.
This suggests a couple of possible fixes:

- make NEURON generate `nt` instead of `_nt` (this PR)
- make coreNEURON generate `_nt` instead of `nt` (requires a bit more work since the unit tests check the codegen directly)
- refactor `ImplicitArgumentVisitor` (or make another one, or add a function in the codegen) which correctly sets the right variable (either with or without an underscore) depending on the the code backend used for the codegen
- ??? (your proposal here)


test/usecases/implicit_argument/cacur.mod
```diff
@@ -0,0 +1,27 @@
+NEURON {
+	SUFFIX cacur
+	USEION ca WRITE ica
```
1uc, line 3: Why mixed with ions?

test/usecases/implicit_argument/simulate.py
```diff
@@ -0,0 +1,44 @@
...
+    v = np.array(v_hoc.as_numpy())
+    t = np.array(t_hoc.as_numpy())
+
+    return t, v
+
+
+def check_solution(t, v):
+    # solution is an affine function until t = 1 (ms), afterwards is a constant
+    # (1000 - 65)
+    solution = 1000 * t - 65
```
1uc, line 29: You can make it terser if you want by using `np.min(t, 1)`. The comment merely states literally what's written in Python, it can go (so it wont become outdated).

test/usecases/implicit_argument/simulate.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
+from neuron import gui, h
+from neuron.units import ms
+
+
+def test_cacur():
```
1uc, line 6: Not a test. Hence, it needs a different name.

test/usecases/implicit_argument/cacur.mod
JCGoran: It's this file here: https://github.com/neuronsimulator/nrn/blob/master/share/examples/nrniv/nmodl/cacur.mod

test/usecases/implicit_argument/simulate.py
JCGoran: `np.minimum`, but yes I agree.

test/usecases/implicit_argument/cacur.mod
1uc: Good to know. We have Model DB CI to check a variety of scientific examples. In the "usecases" we'd like to break down each language construct into the minimum needed to test it. Here `at_time` is likely not related to ions. Please strip everything unrelated. Then simplify such that we can use the MOD file to check one thing at a time.

Do you know why it's called `implicit_argument`? We've (mostly) switch calling the directory by the name of the keyword it's testing, i.e. `at_time`. (If multiple keywords are required in a test we pick one.)

1uc:
Please check how widely `_nt` is used in VERBATIM code. Then we can decide if switching to `nt` is a good idea.

test/usecases/implicit_argument/cacur.mod
JCGoran: >  In the "usecases" we'd like to break down each language construct into the minimum needed to test it.

In this case, I don't think explicit testing is required, what's important is that it compiles since it's an internal NEURON function (so I assume the functionality itself is tested somewhere in the NEURON test suite).

> Do you know why it's called implicit_argument?

Yes, it's because it's a consequence of the `ImplicitArgumentVisitor`:
https://github.com/BlueBrain/nmodl/blob/5ff221d045c92389b4cab5ee159d46770aa12665/src/main.cpp#L536-L538

JCGoran:
@1uc let's just say the answer is "non-zero". Can we change coreNEURON codegen instead then (i.e. use option 2 from the description)?

1uc:
You might want to check neurodamus-models before changing the tests, BBP has written MOD files that consist mostly of VERBATIM blocks. I assume they work with CoreNEURON (and NEURON). Therefore, there's a good chance that they use both `_nt` and `nt` depending on which simulator is targeted.

JCGoran:
Using both `grep -RiIl 'verbatim' | xargs grep '\<_nt\>'` and directly `grep -RiI '\<nt\>'` reveals no matches in `neurodamus-models`; any other repos that could potentially be problematic? Otherwise, if there's too many things that could potentially break, I'll go back to the drawing board.

1uc (approved):

1uc (approved):

PR #1326, opened 2024-06-18
Fix (NON)LINEAR block with explicit SOLVEFOR

JCGoran:
The (NON)LINEAR block in NMODL was not accounting for the SOLVEFOR variables, which means that something like the following would fail (when used with `sympy --analytic`):

```plaintext
STATE {
    x
    y
    z
}
LINEAR lin SOLVEFOR x, y {
    ~ 3 * x = v - y
    ~ x = z * y - 5
}
```

because the SOLVEFOR variables were not accounted for, only the STATE ones, which caused the solver to think this was a nonlinear system due to the presence of the `z * y` term in the second equation (i.e. the solver would try to solve the system for _all_ STATE variables instead of just the ones specified in SOLVEFOR).

### DIFF
--- a/src/visitors/sympy_solver_visitor.cpp
+++ b/src/visitors/sympy_solver_visitor.cpp
@@ -284,8 +284,6 @@ void SympySolverVisitor::construct_eigen_solver_block(
 
 
 void SympySolverVisitor::solve_linear_system(const std::vector<std::string>& pre_solve_statements) {
-    // construct ordered vector of state vars used in linear system
-    init_state_vars_vector();
     // call sympy linear solver
     bool small_system = (eq_system.size() <= SMALL_LINEAR_SYSTEM_MAX_STATES);
     auto solver = pywrap::EmbeddedPythonLoader::get_instance().api().solve_linear_system;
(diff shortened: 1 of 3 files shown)


src/visitors/sympy_solver_visitor.cpp
```diff
@@ -566,6 +564,18 @@ void SympySolverVisitor::visit_linear_block(ast::LinearBlock& node) {
...
+        init_state_vars_vector();
+
+        // in case we have a SOLVEFOR in the block, we need to set `state_vars` to those instead
+        const auto& solvefor_vars = node.get_solvefor();
+        if (!solvefor_vars.empty()) {
+            state_vars.clear();
+            for (const auto& solvefor_var: solvefor_vars) {
+                state_vars.push_back(solvefor_var->get_node_name());
+            }
+        }
```
1uc, line 577: The new lines seem to be well described by "initialize state variables" is there a reason it's not part of the function `init_state_vars_vector()`? It would have the advantage that reading the function that takes care of initializing `state_vars` is clearly aware of `SOLVEFOR`. Hence, one then doesn't need to search for the place where it gets fixed up; and then checking that it always gets fixed up correctly.

JCGoran: Refactored.

JCGoran:
It seems that, according to the docs, you can also use SOLVEFOR in KINETIC, which I believe requires modifying some other visitor (which one?):
https://github.com/neuronsimulator/nrn/blob/386212697752a52a083b7611a07ab4030d9cb3f0/src/nmodl/simultan.cpp#L372-L389

1uc (approved):

PR #1327, opened 2024-06-18
Make non-thread-safe warnings more specific

JCGoran:
The warning `MOD file uses non-thread safe constructs of NMODL` does not really tell me which ones are not thread-safe, so this PR makes it more specific and now outputs a warning with about the ones which are not safe. More of a UX improvement than anything else.

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -1359,11 +1359,6 @@ void CodegenCppVisitor::setup(const Program& node) {
     }
     info.rsuffix = info.point_process ? "" : "_" + info.mod_suffix;
 
-    if (!info.vectorize) {
-        logger->warn(
-            "CodegenCoreneuronCppVisitor : MOD file uses non-thread safe constructs of NMODL");
-    }
-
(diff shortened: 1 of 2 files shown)


1uc (commented):
Seems error prone and `vectorize` is apparently obsolete, we're working towards removing it.

pramodk:
@JCGoran: just to expand a bit Luc’s comment:  In the context of neuron, we will need to support code generation for all cases i.e. everything that nocmodl supports today. As part of the PR like https://github.com/neuronsimulator/nrn/pull/2906, we are removing/refactoring parts related to VECTORIZE. Hence this wouldn’t be needed.

1uc:
I'm sorry, I confused myself. We're getting rid of `vectorize` by renaming it `thread-safe` and sometimes genuinely removing it. The problem is that `vectorize` refers to both putting data into vectors (which we always do nowadays); and being thread-safe (which MOD file may or may not be). These instances would likely just become `info.thread_safe`.

Personally, I'd do it differently. Keep the catch all as a warning (because users might want to know that something "is up") and make the new log messages info or debug (because the details are only important while debugging).

PR #1329, opened 2024-07-02
Include solvers to avoid include path dependency.

matz-e:
Similar to #1305, include the solver headers in variables in a C++
header, and include them in the code generated for the MOD files. This
should allow to substitute different NMODL binaries in NEURON builds
without confusion as to which solver headers are going to be included in
the MOD file compilation.

Tested in neuronsimulator/nrn#2953

CI_BRANCHES:NEURON_BRANCH=nmodl-include-header-cleanup


src/solver/CMakeLists.txt
```diff
@@ -1,22 +1,15 @@
...
-# =============================================================================
-
-install(DIRECTORY ${CMAKE_BINARY_DIR}/include/ DESTINATION ${NMODL_INSTALL_DIR_SUFFIX}include)
+# Read headers, remove everything up to and including "#pragma once", then remove header includes.
+file(READ ${CMAKE_CURRENT_SOURCE_DIR}/crout/crout.hpp NMODL_CROUT_HPP_RAW)
+string(REGEX REPLACE ".*#pragma once[ \t\r\n]*" "" NMODL_CROUT_HPP "${NMODL_CROUT_HPP_RAW}")
+file(READ ${CMAKE_CURRENT_SOURCE_DIR}/newton/newton.hpp NMODL_NEWTON_HPP_RAW)
+string(REGEX REPLACE ".*#pragma once[ \t\r\n]*" "" NMODL_NEWTON_HPP_TMP "${NMODL_NEWTON_HPP_RAW}")
+string(REGEX REPLACE "#include <crout/crout.hpp>[ \t\r\n]*" "" NMODL_NEWTON_HPP
+                     "${NMODL_NEWTON_HPP_TMP}")
```
pramodk, line 7: Like this PR! The only thing I am a bit "uncomfortable" about is this string processing in CMake and I wonder if there could be a better / more robust way?

Let's see also what @alkino says!

alkino: in CMake there is none

src/solver/solver.hpp.inc
```diff
@@ -0,0 +1,20 @@
...
+//
+// To avoid a dependency of the `nmodl` binary on a specific header include directory and
+// to avoid mod file compilation using wrong headers, we include them like this.
+//
+// However, because we want to be able to test the headers separately we can't
+// move them here.
+
+namespace nmodl::solvers {
+const std::string crout_hpp = R"jiowi(
+@NMODL_CROUT_HPP@
```
alkino, line 13: jiowi?

src/solver/CMakeLists.txt
matz-e: I decided to leave the headers as standard C++ - if I wanted to forego the CMake text manipulation, I'd have to at least remove the `#pragma once`  - which seems a step backwards.

src/solver/solver.hpp.inc
matz-e: @1uc used that for `ode.py` :)

Just a unique identifier garbage.

src/solver/CMakeLists.txt
matz-e: Yes, in the GitLab CI we test a full installation of Neurodamus with CoreNeuron and NMODL.

This is the same guard we have against typos in the code generation, too, no?

pramodk: Ok. I just thinking of gitlab going away and if this would be tested locally or GitHub.

 I took a look at the current CI configs and it's true that we have been only testing code generation and not compiling/linking.

But, very soon `test/usecases` will have solver related mod files and then this will be tested as part of github CI. I would say this is good to go!

pramodk (approved):

CMakeLists.txt
```diff
@@ -111,8 +111,6 @@ include(cmake/hpc-coding-conventions/cpp/cmake/3rdparty.cmake)
 if(NOT TARGET CLI11::CLI11)
   cpp_cc_git_submodule(cli11 BUILD PACKAGE CLI11 REQUIRED)
 endif()
-# For the moment do not try and use an external Eigen.
-cpp_cc_git_submodule(eigen)
```
pramodk, line 115: @matz-e : reference in `.gitmodules` needed?

matz-e: Moved down and will only get triggered if tests are requested - we still need Eigen to test the solvers.

PR #1334, opened 2024-07-09
Implement `wrote_conc`.

1uc:


docs/contents/ions.rst
```diff
@@ -107,6 +107,47 @@ This optimization is implemented in NMODL. It can be activated on the CLI via
...
+Ion Styles
+~~~~~~~~~~
+The style is a 12-bit wide bitfield and stored in the ``dparam`` array of the
+``Prop`` as an ``int``.  It's stored at location ``0`` (variable name:
+``iontype_index_dparam``) in the ``dparam`` array.
+
+From MOD files we need to notify NEURON if we're reading/writing concentration
+or reversal potential. This is done by calling ``nrn_promote(., conc, rev)``.
+The magic value for ``rev`` and ``conc`` is ``1`` for reading and ``3`` for
+writing.
```
alkino, line 148: All of this is a bit unclear for me and I don't know if we could add some small snippet here and there.
If this is irrelevant only ignore this comment.

1uc: Thank you, I read it again and fixed a mistake and spotted an omission. The difficulty with this text is that in a way this is maybe not the place to document this at all. For example the exact meaning of each of the 12 bits is information internal to NEURON and when generating code we don't need to know, but we do kinda need to know that it's an int and used as a bit-field. Similarly, we're repeating the magic values.

What we do need to know is four things:
1. We must tell NEURON that we'll write to concentrations. Via `nrn_writes_conc` when registering the mech.
2. We must tell NEURON more accurately which concentration we're writing, so it can warn the user if multiple MOD files set the same concentration. Via `nrn_check_conc_write` when allocating the prop.
3. We must inform NEURON (yet again) that we're reading/writing a concentration and/or the reversal potential. Via `nrn_promote` also while allocating the prop.
4. We must tell NEURON (in the INITIAL block only) what we've written the concentration. So it can decide if it needs to recompute the reversal potential. `nrn_wrote_conc`.

In addition there's a few pieces of information that are needed to implement the above requirements. We need to know where the `style` is stored in the ion `Prop` (i.e. at `0`, yes it's a hard-coded constant). I can't help but feel it would be nice to know that exterior is `0` and interior is `1`. Similarly reading is `1` and writing is `3`.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -565,11 +565,25 @@ void CodegenNeuronCppVisitor::print_namespace_stop() {
...
+    auto style_name = fmt::format("_style_{}", ion_name);
+    auto style_stmt = fmt::format("int {} = *(_ppvar[{}].get<int*>())", style_name, dparam_index);
+    statements.push_back(ShadowUseStatement{style_stmt, "", ""});
+
+
+    auto wrote_conc_stmt = fmt::format("nrn_wrote_conc(_{}_sym, {}, {}, {}, {})",
+                                       ion_name,
+                                       get_variable_name("ion_" + ion_name + "_erev"),
+                                       get_variable_name("ion_" + ion_name + "i"),
+                                       get_variable_name("ion_" + ion_name + "o"),
```
pramodk, line 584: A suggestion: for the `ion_name + "_erev"`, `ion_name + "i"`, etc parts: I thought we have methods in `struct Ion` to get these names. But I see that there is `ion.is_intra_cell_conc()` but no `ion.intra_cell_conc()`. Maybe we should add those in order to avoid these things getting hardcoded at multiple places?

I know there are already inconsistencies / hardcoded things. But just want to mention whenever I am seeing this aspect.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -565,11 +565,25 @@ void CodegenNeuronCppVisitor::print_namespace_stop() {
     printer->pop_block();
 }
 
+void CodegenNeuronCppVisitor::append_conc_write_statements(
+    std::vector<ShadowUseStatement>& statements,
+    const Ion& ion,
+    const std::string& /* concentration */) {
```
pramodk, line 571: still needs to be kept?

1uc: What needs to be kept? The unused argument `concentration`? Yes, CoreNEURON generates complately different code and it needs it. At least IIRC.

pramodk: yes, that's what I was referring to! my bad, I missed it's neuron-specific visitor!

src/utils/string_utils.hpp
```diff
@@ -146,6 +146,15 @@ static inline bool ends_with(const std::string& haystack, const std::string& nee
     return std::equal(haystack_begin, haystack.end(), needle.begin(), needle.end());
 };
 
+/**
+ * Check if `haystack` starts with `needle`.
+ *
+ * Every string ends with the empty string.
```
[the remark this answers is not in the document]
pramodk: As a developer, code fragment helped me better understand what you meant! :)

PR #1338, opened 2024-07-12
Remove `inline` from function definitions.

1uc:
Since the compiler decides to inline regardless of the keyword `inline`, it's not helpful for performance reasons.

The issue is that if we accidentally define multiple, incompatible copies of the same symbol, when linking we'll got get an error, instead the linker will pick one of the definitions at random.

Since the compiler, might decide to inline, we might not immediately notice the bug, e.g. through integration tests.

### DIFF
--- a/src/codegen/codegen_coreneuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_coreneuron_cpp_visitor.cpp
@@ -737,7 +737,7 @@ void CodegenCoreneuronCppVisitor::append_conc_write_statements(
 
 void CodegenCoreneuronCppVisitor::print_first_pointer_var_index_getter() {
     printer->add_newline(2);
-    printer->push_block("static inline int first_pointer_var_index()");
+    printer->push_block("static int first_pointer_var_index()");
     printer->fmt_line("return {};", info.first_pointer_var_index);
     printer->pop_block();
 }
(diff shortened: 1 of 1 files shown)


pramodk:
> Since the compiler decides to inline regardless of the keyword inline, it's not helpful for performance reasons.

Aware that 1) inline is just hint and not a guarantee 2) compilers anyway can decide to inline functions based on some heuristics 

But just want to add historical context and experience of why « inline » was added:

from my experience, 1) compiler to compiler, when and how functions are inlined is a bit different 2) presence of inline still affects that decision 3) we added inline first time (2016/17/18) when PGI/NVHPC was not inlining functions and causing issues in OpenACC porting. Otherwise, there were link errors and we needed to define those functions as « acc seq ».

But that’s not the issue in 2024! 

I don’t know about the issue you mentioned but just adding relevant information. (i.e. removing inline is fine if you see it’s needed/helpful)

mgeplf:
@pramodk > 3) we added inline first time (2016/17/18) when PGI/NVHPC was not inlining functions and causing issues in OpenACC porting. Otherwise, there were link errors and we needed to define those functions as « acc seq ».

For the link errors, were they being specified as `static` like the ones in this PR?  I'm interested in what sort of link errors were happening.

pramodk:
> For the link errors, were they being specified as static like the ones in this PR? I'm interested in what sort of link errors were happening.

👍 
The link errors I am referring to are related to the OpenACC programming model: when a function is not inlined and used in a GPU offloaded region then that function needs to have a GPU version (i.e. like `__device__` kernel). Absence of this end up in a link error for GPU build/execution. Addition of `#acc routine seq` then explicitly tells the compiler to create a device kernel.

PR #1346, opened 2024-07-16
Fix RANGE PARAMETER defaults.

1uc:
PARAMETERs that are RANGE variables need to register their default
values, because NEURON needs to be able to initialize the parameters
with their default values.

This commits adds the registration code, and sets defaults in
`nrn_alloc` from the same array.

### DIFF
--- a/test/usecases/parameter/test_default.py
+++ b/test/usecases/parameter/test_default.py
@@ -1,0 +1,55 @@
+from neuron import h
+
+
+def check_defaults(obj, parameters):
+    defaults = {
+        "b": 0.1,
+        "y": 2.1,
+    }
+
+    for p in parameters:
+        actual = getattr(obj, f"{p}_default_parameter")
+        expected = defaults.get(p, 0.0)
+        assert actual == expected
+
+
+def test_default_values():
+    nseg = 10000
+
+    s = h.Section()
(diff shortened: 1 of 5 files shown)


test/usecases/parameter/test_default.py
```diff
@@ -0,0 +1,55 @@
...
+    # correctly.
+
+    nseg = 5
+
+    s = h.Section()
+    s.nseg = nseg
+    s.insert("default_parameter")
+
+    # Caused a SEGFAULT:
+    h.load_file("parcom.hoc")
```
pramodk, line 50: just to clarify: loading neuron's `parcom.hoc` / ParallelComputeTool caused a segfault but we don't know which aspect?

pramodk (approved):

test/usecases/parameter/test_default.py
1uc: We do, you need to instantiate a `NrnProperty`. Which will lead to these lines:
https://github.com/neuronsimulator/nrn/blob/f8f9cc649a755545ba2966354ae35e26b2c1811a/src/nrniv/ndatclas.cpp#L69-L72

which results in a SEGFAULT, if nothing was registered in `memb_func[mechtype].parm_default`.

PR #1348, opened 2024-07-16
Remove usage of `unistd` from Flex

JCGoran:
Mainly useful for Windows support.


pramodk (commented):
Sounds good!

Just a minor thing - better to use `%option nounistd` as we use `%option` syntax in lexer file for many other options?

alkino (approved):

JCGoran:
Through a bunch of trial and error, it seems that using only `%option nounistd` still generates headers with `unistd.h` included (the CPP files do not have it), while using both `%option nounistd` and calling the CLI with `--nounistd` removes those from the headers as well, so I'm not sure which one of the 2 is more appropriate.

pramodk:
> Through a bunch of trial and error, it seems that using only %option nounistd still generates headers with unistd.h included (the CPP files do not have it), while using both %option nounistd and calling the CLI with --nounistd removes those from the headers as well, so I'm not sure which one of the 2 is more appropriate.

That's correct. But, when we add ` %option nounistd`, it adds `  #define YY_NO_UNISTD_H 1` at the top of generated cpp file. So if I am not mistaken, that should be sufficient 

```cpp
 #ifndef YY_NO_UNISTD_H
 /* Special case for "unistd.h", since it is non-ANSI. We include it way
  * down here because we want the user's section 1 to have been scanned first.
  * The user has a chance to override it with an option.
  */
 /* %if-c-only */
 /* %endif */
 /* %if-c++-only */
 #include <unistd.h>
 /* %endif */
 #endif
...
```

In verbatim.ll we don't have added ` #define YY_NO_UNISTD_H`. I guess we can replace all those by just `%option nounistd` and then it should be sufficient (i.e. flex doesn't need `--nounistd `).

Of course I am not testing this on windows! Correct me if I am missing something obvious!

pramodk (approved):

PR #1349, opened 2024-07-16
Use Eigen on CPUs.

1uc:


src/solver/newton/newton.hpp
```diff
@@ -83,8 +86,10 @@ EIGEN_DEVICE_FUNC int newton_solver(Eigen::Matrix<double, N, 1>& X,
         // Check if J is singular
         if (nmodl::crout::Crout<double>(N, J.data(), pivot.data(), rowmax.data()) < 0)
             return -1;
-        Eigen::Matrix<double, N, 1> X_solve;
         nmodl::crout::solveCrout<double>(N, J.data(), F.data(), X_solve.data(), pivot.data());
+#else
+        X_solve = J.partialPivLu().solve(F);
+#endif
```
pramodk, line 92: Could you add a brief description/comment if this is done for performance reasons or something else?

I don't remember this well and hence asking for clarity: do we expect small differences in the results between the two? If there are two code paths, just think from CPU vs GPU debugging or results comparison perspectives. (obviously, not thinking fp differences as they exist on cpu vs gpu. But more from solver/accuracy perspective).

1uc: Yes, without using NMODL is a 25% regression, compared to NOCMODL on the Masoli circuit we've been running recently.

That's a tricky question. The variation we implement ourselves is also "partial pivoting". There's two ways of doing LU decompositions: Dolittle and Crout. They result in almost the same matrices. Mostly I think they just iterate over the matrix differently. For a well-conditioned matrix, I doubt it makes any difference.

PR #1351, opened 2024-07-16
Use fmt instead of string concatenation

alkino:


1uc (approved):

pramodk (commented):
didn't think of it! 😅  thanks @alkino !

PR #1352, opened 2024-07-16
Accept negative number for limits

alkino:


1uc:
Overnight, `neuron-nightly` updated and has the corresponding fix in `nocmodl` needed to get the "usecases" to compile (because we compare against nocmodl).

1uc (approved):

PR #1355, opened 2024-07-17
Smaller changes for MSVC compatibility.

matz-e:
These fixes come from #1354, and should not be controversial.

### DIFF
--- a/src/parser/diffeq_context.cpp
+++ b/src/parser/diffeq_context.cpp
@@ -151,7 +151,7 @@ std::string DiffEqContext::get_non_cnexp_solution() const {
     std::string result;
     if (!deriv_invalid) {
         result = get_cvode_linear_diffeq();
-    } else if (deriv_invalid and eqn_invalid) {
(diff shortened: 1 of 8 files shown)


src/parser/diffeq_context.cpp
```diff
@@ -151,7 +151,7 @@ std::string DiffEqContext::get_non_cnexp_solution() const {
     std::string result;
     if (!deriv_invalid) {
         result = get_cvode_linear_diffeq();
-    } else if (deriv_invalid and eqn_invalid) {
+    } else if (deriv_invalid && eqn_invalid) {
```
pramodk, line 154: I thought I never used `and` instead of `&&`! Anyway, I guess this was not causing issue on windows?

pramodk (approved):
LGTM

src/parser/diffeq_context.cpp
matz-e: It was 😅

PR #1356, opened 2024-07-17
Sort inputs to std::set_intersection.

matz-e:
Found by MSVC. In the reference:

> Constructs a sorted range beginning at d_first consisting of elements that are found in **both sorted ranges** [first1, last1) and [first2, last2).

(emphasis mine)

See https://en.cppreference.com/w/cpp/algorithm/set_intersection


1uc:
What went red, aka are we missing test coverage?

matz-e:
> What went red, aka are we missing test coverage?

It triggered a debug assert in MSVC when parsing the unit file. Apparently, Microsoft eagerly checks that the inputs are sorted.

1uc:
Feels like an observable problem (with a little bit of luck), meaning we should add the test required to cover the class of issues.

mgeplf:
> Feels like an observable problem (with a little bit of luck), meaning we should add the test required to cover the class of issues.

The unsorted usage is UB, so I'm not sure you'd be able to guarantee it's observable.
I think @matz-e PR here is a good plan to catch these things in the future: https://github.com/BlueBrain/nmodl/pull/1358

matz-e:
Above PR fails like this now: https://github.com/BlueBrain/nmodl/actions/runs/9991287283/job/27613892488?pr=1358

PR #1357, opened 2024-07-17
Provide both end() iterators to std::mismatch to prevent seeking beyond the end of std::string

matz-e:
Became apparent with MSVC, which in this case crashes hard comparing `it.first = "yotta"` and `denomitator_name = "sec2"`.

### DIFF
--- a/src/units/units.cpp
+++ b/src/units/units.cpp
@@ -132,7 +132,7 @@ void UnitTable::calc_nominator_dims(const std::shared_ptr<Unit>& unit, std::stri
         while (changed_nominator_name) {
             changed_nominator_name = 0;
             for (const auto& it: prefixes) {
-                auto res = std::mismatch(it.first.begin(), it.first.end(), nominator_name.begin());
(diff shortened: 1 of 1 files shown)


1uc:
How did you find this? With the immediate following are we missing test coverage?

matz-e:
> How did you find this? With the immediate following are we missing test coverage?

MSVC's debug build has very aggressive asserts in the STL. Building with MSVC will provide adequate coverage? Maybe GCC/Clang have other guards for this.

1uc:
It reads like it's a proper bug. On that should result in an ASAN failure. If true, why do we need MSVC to find the bug? Are we not running ASAN during CI? Or are we missing a test that exposes the problem? Again, it would be helpful to know what exactly failed.

matz-e:
I think there are flags in the glibc that would trigger similar asserts.

1uc:
Very good!

1uc (approved):

PR #1358, opened 2024-07-18
Let glibcxx be more assertive, fix `std::set_intersection` usage

matz-e:
Will ensure that macros check the inputs to stl algorithms etc. Sample output:

https://github.com/BlueBrain/nmodl/actions/runs/9991287283/job/27613892488

Originally found by MSVC. In the C++ reference:

> Constructs a sorted range beginning at d_first consisting of elements that are found in **both sorted ranges** [first1, last1) and [first2, last2).

(emphasis mine)

See https://en.cppreference.com/w/cpp/algorithm/set_intersection

Closes #1356.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -148,7 +148,7 @@ stages:
       submodules: True
     - script: |
         brew install flex bison cmake python@3
-        python3 -m pip install --upgrade pip "setuptools<=70.0.3"
+        python3 -m pip install --upgrade pip "setuptools<=70.3.0"
         python3 -m pip install --user -r requirements.txt
       displayName: 'Install Dependencies'
     - script: |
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -148,7 +148,7 @@ stages:
       submodules: True
     - script: |
         brew install flex bison cmake python@3
-        python3 -m pip install --upgrade pip setuptools
+        python3 -m pip install --upgrade pip "setuptools<=70.3.0"
```
1uc, line 151: Where does this requirement come from? (I think I've seen it in the NEURON world.)

matz-e: Otherwise, I get CI failure when installing neuron

1uc: Alright, if that's what it takes to make it run :)

1uc (approved):

PR #1360, opened 2024-07-22
Fix `net_send` and `net_move` for ARTIFICIAL_CELLs.

1uc:
For `net_send` and `net_move` we need to call `artcell_*`. This enables using a
self queue. Since this is an optimization it's hard to test functionally. Hence, we
only perform a crude check that `artcell_*` exist in the generated C++ file.

### DIFF
--- a/test/unit/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/test/unit/codegen/codegen_neuron_cpp_visitor.cpp
@@ -209,3 +209,39 @@ SCENARIO("Read `eca`.", "[codegen]") {
         }
     }
 }
+
+SCENARIO("ARTIFICIAL_CELL with `net_send`") {
+    GIVEN("a mod file") {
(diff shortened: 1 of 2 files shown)


test/unit/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -209,3 +209,39 @@ SCENARIO("Read `eca`.", "[codegen]") {
...
+            }
+            NET_RECEIVE(w) {
+                net_send(t+1, 1)
+            }
+        )";
+        std::string cpp = transpile(nmodl);
+
+        THEN("it contains") {
+            REQUIRE_THAT(cpp, ContainsSubstring("artcell_net_send"));
+        }
```
pramodk, line 227: For cases like this, this seems ok/acceptable!  :)

pramodk (approved):

PR #1362, opened 2024-07-22
Deduplicate integer magic code.

1uc:
The code to compute the "position" of float/integer variables was duplicated
several times, this extracts the shared code in to separate functions.


pramodk (approved):
♻️ !

PR #1366, opened 2024-07-22
Use SymPy solver when using `derivimplicit` method with NEURON codegen

JCGoran:
One runs into a bunch of compile errors with https://github.com/BlueBrain/nmodlbench/blob/master/benchmark/channels/lib/modlib/cacumm.mod otherwise.

Also added tests.


test/usecases/solve/test_scalar.py
```diff
@@ -0,0 +1,33 @@
...
+    x = np.array(x_hoc.as_numpy())
+    t = np.array(t_hoc.as_numpy())
+
+    x_exact = 42.0 * np.exp(-t)
+    np.testing.assert_allclose(x, x_exact, rtol=rtol)
+
+
+if __name__ == "__main__":
+    simulate("cnexp", rtol=1e-12)
+    simulate("derivimplicit", rtol=1e-3, dt=1e-4)
```
1uc, line 33: How large is the error with larger timesteps? Does it differ between NOCMODL and NMODL generated code?

test/usecases/solve/test_scalar.py
```diff
@@ -0,0 +1,33 @@
...
+    if dt is not None:
+        h.dt = dt * ms
+    h.tstop = 5.0 * ms
+    h.run()
+
+    x = np.array(x_hoc.as_numpy())
+    t = np.array(t_hoc.as_numpy())
+
+    x_exact = 42.0 * np.exp(-t)
+    np.testing.assert_allclose(x, x_exact, rtol=rtol)
```
1uc, line 28: Very intersting. This might solve a very annoying aspect of the usecases: `assert ..., f"{} - {} = {}"`

test/usecases/solve/test_scalar.py
```diff
@@ -0,0 +1,33 @@
+from typing import Optional
+
+import numpy as np
+from neuron import h, gui
+from neuron.units import ms
+
+
+def simulate(method: str, rtol: float, dt: Optional[float] = None):
```
1uc, line 8: I don't mind the type hints. If we want to use them, we need to enforce them. Otherwise inevitably they'll become outdated and highly misleading. I remember this happening to me in other BBP code and it's simply not worth it.

A quick test suggests what `mypy` can be `pip install`ed and easily made to spot issues in our code (if we ignore imports).

Therefore, if you want to keep the type hints, please create an issue with the intention of working on it at some point in the near future.

1uc (commented):
Thank you this makes sense.

test/usecases/solve/test_scalar.py
JCGoran: Good question! From my experiments, both NMODL and NOCMODL produce identical relative differences w.r.t. the exact result, with the following max relative error:
- `0.06340655` for default value of `h.dt`
- `0.00250146` for `h.dt = 1e-3 * ms`
- `0.00025001` for `h.dt = 1e-4 * ms`
- `2.49963513e-06` for `h.dt = 1e-6 * ms`

and so on.

test/usecases/solve/test_scalar.py
JCGoran: I've removed the type hints for now, will open a separate issue to maybe add those in the future.

test/usecases/solve/test_scalar.py
1uc: Perfect, exactly the convergence rate we want to see.

1uc (approved):

PR #1367, opened 2024-07-23
Implement RANDOM for NEURON.

1uc:


test/usecases/random/test_random_variable.py
```diff
@@ -0,0 +1,35 @@
...
+        np.array([s(x).random_variable.negexp() for _ in range(n_samples)])
+        for x in [0.34, 0.74]
+    ]
+    expected_hashes = [
+        "3b66d7f83dc81ea485929aa8a4f347a3befee7170971dfb9c35a1a1d40ed0407",
+        "896c59d129cc0a248e57c548c8bf4be7ae4d39fbd92113e37f59c332be61a2b7",
+    ]
+
+    assert hash_array(samples[0]) == expected_hashes[0]
+    assert hash_array(samples[1]) == expected_hashes[1]
```
JCGoran, line 31: There's also `numpy.testing.assert_string_equal` which is a bit more readable on failure.

JCGoran (approved):
LGTM, especially since it fixes a bunch of models in the channel benchmark.

test/usecases/random/test_random_variable.py
1uc: Not 100% convinced:
```
In [3]: np.testing.assert_string_equal("foo", "bar")
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 np.testing.assert_string_equal("foo", "bar")

File ~/git/bbp/nmodl/venv/lib/python3.12/site-packages/numpy/testing/_private/utils.py:1190, in assert_string_equal(actual, desired)
   1188 msg = f"Differences in strings:\n{''.join(diff_list).rstrip()}"
   1189 if actual != desired:
-> 1190     raise AssertionError(msg)

AssertionError: Differences in strings:
- foo+ bar
```
Plus they're really just checksums that either match or don't. The value is uninteresting.

PR #1378, opened 2024-07-30
Remove `print_backend_namespace_{start,stop}`.

1uc:
The only backend specialized backend: acc, doesn't use this either.


JCGoran (approved):
Looking through the git history, I suppose we don't use/need this anymore.

1uc:
Thanks for checking, I couldn't find anything either.

PR #1380, opened 2024-07-30
Refactor print_namespace_{start,stop}.

1uc:
Since the only thing that differs is the name of the namespace, we're better off adding a virtual method for the namespace name, and devirtualize the printing of `namespace {} {`.


JCGoran (approved):
LGTM, nice simplification.

PR #1383, opened 2024-08-02
Throw an error when using `derivimplicit` with `PROCEDURE` block

JCGoran:
The following modfile does not work with NOCMODL:

```plaintext
: mwe.mod
NEURON	{
	SUFFIX example
}

BREAKPOINT	{
    SOLVE myfunc METHOD derivimplicit
}

PROCEDURE myfunc() {}
```

With the error:

```plaintext
Error: Method derivimplicit can't be used with Block myfunc at line 10 in file mwe.mod
```

On the other hand, `nmodl --neuron mwe.mod` just segfaults with no useful error; with debugging symbols, I get a slightly more verbose backtrace:

```plaintext
Assertion failed: (node_to_solve->get_node_type() == ast::AstNodeType::DERIVATIVE_BLOCK), function create_solution_expression, file solve_block_visitor.cpp, line 59
```

This PR fixes it so it actually throws a `runtime_error` with the traceback:

```plaintext
terminating due to uncaught exception of type std::runtime_error: Method derivimplicit cannot be used for ProcedureBlock block myfunc
```

### DIFF
--- a/src/visitors/solve_block_visitor.cpp
+++ b/src/visitors/solve_block_visitor.cpp
@@ -56,7 +56,12 @@ ast::SolutionExpression* SolveBlockVisitor::create_solution_expression(
     if (solve_method == codegen::naming::DERIVIMPLICIT_METHOD &&
         !has_sympy_solution(*node_to_solve)) {
         /// typically derivimplicit is used for derivative block only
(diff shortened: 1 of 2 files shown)


1uc:
This could either be tested via Catch2 or we can setup a directory `tests/invalid_mod_files` to collect numerous invalid modfiles, that we run similarly to the `usecases`.

1uc (approved):

PR #1384, opened 2024-08-02
Toggle `emit_cvode` for codegen

JCGoran:
Note: there is no actual codegen in the PR, this one is only meant for toggling the `emit_cvode` flag.

Notable changes:

- ~~added `emit_cvode()` public method to `CodegenCppVisitor` since `codegen` is a protected member so we can't access it directly (open to suggestions for how to handle this in another way)~~ public API remains unchanged


src/codegen/codegen_helper_visitor.cpp
```diff
@@ -152,6 +153,66 @@ void CodegenHelperVisitor::find_ion_variables(const ast::Program& node) {
...
+                                   ->get_node_name() &&
+                           method->get_node_name() == codegen::naming::AFTER_CVODE_METHOD;
+                });
+        });
+
+    // only case when we emit CVODE code is if we have exactly one block, and
+    // that block is either a KINETIC/DERIVATIVE with any method, or a
+    // PROCEDURE with `after_cvode` method
+    if (solve_nodes.size() == 1 && (kinetic_or_derivative_nodes.size() || using_after_cvode)) {
+        logger->debug("Emitting code for CVODE");
```
alkino, line 212:
```suggestion
        logger->debug("Will emit code for CVODE");
```

Sorry, this is a nitpick but if we debug it one day, this is good to know that this pass is not the emitting itself.

alkino (approved):

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -152,6 +153,66 @@ void CodegenHelperVisitor::find_ion_variables(const ast::Program& node) {
...
+                [&solve_node](const auto& procedure_node) {
+                    const auto& method =
+                        std::dynamic_pointer_cast<const ast::SolveBlock>(solve_node)->get_method();
+                    if (!method) {
+                        return false;
+                    }
+                    return procedure_node->get_node_name() ==
+                               std::dynamic_pointer_cast<const ast::SolveBlock>(solve_node)
+                                   ->get_block_name()
+                                   ->get_node_name() &&
```
1uc, line 203: Long enough to deserve a name & can be pulled out of the `any_of`.

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -152,6 +153,66 @@ void CodegenHelperVisitor::find_ion_variables(const ast::Program& node) {
...
+    // all (global) kinetic/derivative nodes
+    const auto& kinetic_or_derivative_nodes =
+        collect_nodes(node, {AstNodeType::KINETIC_BLOCK, AstNodeType::DERIVATIVE_BLOCK});
+
+    // all (global) procedure nodes
+    const auto& procedure_nodes = collect_nodes(node, {AstNodeType::PROCEDURE_BLOCK});
+
+    // there can only be one BREAKPOINT block in the entire program
+    const auto& breakpoint_node = std::dynamic_pointer_cast<const ast::BreakpointBlock>(
+        breakpoint_nodes[0]);
```
1uc, line 183: This should move upwards to the other breakpoint code. The comment can be expressed nicely as an `assert`.

src/codegen/codegen_cpp_visitor.hpp
```diff
@@ -1500,6 +1500,12 @@ class CodegenCppVisitor: public visitor::ConstAstVisitor {
      */
     void print_table_check_function(const ast::Block&);
 
+    /**
+     * Whether or not we need to emit code for CVODE (only makes a difference
+     * for NEURON codegen).
+     */
+    bool emit_cvode() const;
```
1uc, line 1507: Should this be part of the CppVisitor if we only need it for NEURON? Do we need it for anything other than testing? If no, is there a reasonably simple change prevents adding a new getter?

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -152,6 +153,66 @@ void CodegenHelperVisitor::find_ion_variables(const ast::Program& node) {
...
+                    const auto& method =
+                        std::dynamic_pointer_cast<const ast::SolveBlock>(solve_node)->get_method();
+                    if (!method) {
+                        return false;
+                    }
+                    return procedure_node->get_node_name() ==
+                               std::dynamic_pointer_cast<const ast::SolveBlock>(solve_node)
+                                   ->get_block_name()
+                                   ->get_node_name() &&
+                           method->get_node_name() == codegen::naming::AFTER_CVODE_METHOD;
```
1uc, line 204: Since they're mentioned in the comment: how are `cvode_t` and `cvode_t_v` detected?

src/codegen/codegen_helper_visitor.cpp
```diff
@@ -152,6 +153,66 @@ void CodegenHelperVisitor::find_ion_variables(const ast::Program& node) {
...
+
+    // check whether any of the SOLVE blocks are solving any PROCEDURE with `after_cvode` method
+    const auto using_after_cvode = std::any_of(
+        solve_nodes.begin(), solve_nodes.end(), [&procedure_nodes](const auto& solve_node) {
+            return std::any_of(
+                procedure_nodes.begin(),
+                procedure_nodes.end(),
+                [&solve_node](const auto& procedure_node) {
+                    const auto& method =
+                        std::dynamic_pointer_cast<const ast::SolveBlock>(solve_node)->get_method();
```
1uc, line 196: The pointer_cast can be done once.

1uc (commented):
Nice, how did you find the set of rules under which we set `info.emit_cvode`?

src/codegen/codegen_cpp_visitor.hpp
JCGoran: > Do we need it for anything other than testing?

No, which is why I was slightly annoyed when I added it as a shortcut, but I'm not really sure how else to do it without a new getter (other than making the whole `CodegenInfo` struct `public`). Ideas are welcome!

JCGoran:
> Nice, how did you find the set of rules under which we set info.emit_cvode?

Michael mentioned it in the NEURON dev slack channel.

src/codegen/codegen_helper_visitor.cpp
JCGoran: They weren't before, my bad. Added them in `codegen_naming` now.

src/codegen/codegen_cpp_visitor.hpp
1uc: Since you only need the `info` object, not the visitor, can you create the `CodegenInfo` on it's own? Not as part of the visitor, just by itself?

If not, please review why not. Then we can create the appropriate issue :)

JCGoran: The `codegen_helper` tests provided some hints on how to achieve this without changing the API, so the new ones for CVODE-related codegen are now there (which IMHO makes more sense since the changes in the PR are actually to that visitor instead of the NEURON codegen one).

1uc (approved):
Nicely done.

src/codegen/codegen_cpp_visitor.hpp
1uc: You're correct, it makes more sense to test `info` directly.

PR #1387, opened 2024-08-07
Initialize, non-ARRAY, STATE variables.

1uc:


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -1493,6 +1493,22 @@ void CodegenNeuronCppVisitor::print_nrn_init(bool skip_init_check) {
...
+        auto lhs_name = get_variable_name(state_name);
+        auto rhs_name = get_variable_name(state_name + "0");
+
+        if (!state->is_array()) {
+            auto state_name = state->get_name();
+            printer->fmt_line("{} = {};", lhs_name, rhs_name);
+        } else {
+            auto n_elements = state->get_length();
+            printer->fmt_push_block("for(size_t _i = 0; _i < {}; ++_i)", n_elements);
+            printer->fmt_line("{}[_i] = {};", lhs_name, rhs_name);
```
nrnhines, line 1507: Optional.  I wonder if it would be more clear to enclose the one line body of the for loop in ```{}```. (I realize I failed to do that with nocmodl code generation).

nrnhines (approved):

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: When using `push_block` or `fmt_push_block` the `{` is implied. Similary `pop_block` will close the scope with a `}`. In the timeline for NMODL PRs there's often a thin item:

2024-08-09-135545_1669x261_scrot

If you click on "References ..." it'll bring you to the `nmodl-references` repository. For every MOD file in `test/usecases` it will contain the generated `.cpp` files. Generally, once for NEURON and once for CoreNEURON. However, it's configurable, but we rare do so.

The generated code for NEURON is:
```
            inst.X[id] = inst.global->X0;
            inst.Y[id] = inst.global->Y0;
            inst.Z[id] = inst.global->Z0;
            for(size_t _i = 0; _i < 3; ++_i) {
                (inst.A+id*3)[_i] = inst.global->A0;
            }
            for(size_t _i = 0; _i < 2; ++_i) {
                (inst.B+id*2)[_i] = inst.global->B0;
            }
```

PR #1393, opened 2024-08-09
Load 'morphology' symbol during registration

alkino:
This way if someone define again this symbol (in hoc file for example), neuron will use the right one during allocation of the module.

This is how it is currently done in nocmodl.

This behaviour has been spotted with nmodl benchmark.


1uc (commented):
Usually, we can use `get_variable_name("morphology_sym")` to get the string we need to print. However, that requires that it's a "known" global variable. Which because it's type is not `double` is a little tricky. 

There's also a risk of clashing with a GLOBAL variable called `morphology_sym`. Easily avoided by calling it `_morphology_sym` (since user defined variables cant' start with `_`). 

I think it solves a real problem we're facing. Hence, I'd be okay merging it.

1uc (approved):

PR #1399, opened 2024-08-13
CVODE codgen for NEURON

JCGoran:
The implementation works as follows:

- find any `DERIVATIVE` blocks, and copy them into `DERIVATIVE_ORIGINAL_FUNCTION` and `DERIVATIVE_ORIGINAL_JACOBIAN`. This is done since the `cnexp` (and some other visitors) solve the ODEs in-place, so the information about `f(x)` (and consequently its Jacobian) is lost. Why two new constructs instead of one? Because NMODL cannot insert two identical blocks (it doesn't matter if they have a different name) in the symbol table
- `DERIVATIVE_ORIGINAL_FUNCTION` basically replaces `x' = f(x)` with `Dx = f(x)`, while `DERIVATIVE_ORIGINAL_JACOBIAN` replaces `x' = f(x)` with `Dx = Dx / (1 - dt * J(f))`, where `J` is the Jacobian of the transformation (computed analytically using SymPy)

TODO:
- [ ] add tests for KINETIC
- [ ] fix issue with CONSERVE statement


JCGoran:
Closing in favor of #1493

PR #1400, opened 2024-08-14
Deduplicate renaming of state vars

### LINKED ISSUE #1395: Deduplicate printing STATE initial values.
STATE variable require are initialized from `0`-suffixed global variables, i.e. `X` from `X0`. NMODL contains the logic once for CoreNEURON and once for NEURON.

The task would be to extract the code that does:
```c++
printer->fmt_line("{} = {};", get_variable_name("X"), get_variable_name("X0"));
```
Then lift it to `CodegenCppVisitor` and use it in both `Codegen{Coreneuron,Neuron}Visitor`.

alkino:
Shoud close #1395

### DIFF
--- a/src/codegen/codegen_coreneuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_coreneuron_cpp_visitor.cpp
@@ -1793,21 +1793,7 @@ void CodegenCoreneuronCppVisitor::print_initial_block(const InitialBlock* node) 
         printer->add_line(statement);
     }
 
-    // initialize state variables (excluding ion state)
-    for (auto& var: info.state_vars) {
-        auto name = var->get_name();
-        if (!info.is_ionic_conc(name)) {
-            auto lhs = get_variable_name(name);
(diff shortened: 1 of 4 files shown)


1uc:
Here's the changes:
https://github.com/BlueBrain/nmodl-references/commit/4bb98734b97101d4e66559094d5d718c4d02ab37

Please double-check which if you prefer one variation over the other or if they're equal.

1uc (approved):
Nice, one more thing tidied up :)

alkino:
> Here's the changes: [BlueBrain/nmodl-references@4bb9873](https://github.com/BlueBrain/nmodl-references/commit/4bb98734b97101d4e66559094d5d718c4d02ab37)
> 
> Please double-check which if you prefer one variation over the other or if they're equal.

What do you think?
I choose this one to avoid creating a local variable `_i`...

1uc:
I'm coming around to the NOCMODL way of reserving all `_` prefixed names; but either is fine.

PR #1403, opened 2024-08-19
Newton error propagation based convergence check.

1uc:
Since we have access to the Jacobian, we can use Newton error propagation:
```
err = sum(J**2 * sd**2) ** 0.5
```
with `sd = eps * X` to estimate the impact of roundoff errors on `F`.

### DIFF
--- a/src/solver/newton/newton.hpp
+++ b/src/solver/newton/newton.hpp
@@ -72,7 +72,7 @@ EIGEN_DEVICE_FUNC int newton_solver(Eigen::Matrix<double, N, 1>& X,
                                     int max_iter = MAX_ITER) {
     // Vector to store result of function F(X):
     Eigen::Matrix<double, N, 1> F;
-    // Matrix to store jacobian of F(X):
+    // Matrix to store Jacobian of F(X):
     Eigen::Matrix<double, N, N> J;
     // Solver iteration count:
     int iter = -1;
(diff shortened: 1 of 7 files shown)


src/solver/newton/newton.hpp
```diff
@@ -34,8 +34,22 @@ namespace newton {
  * @{
  */
 
-static constexpr int MAX_ITER = 1e3;
-static constexpr double EPS = 1e-12;
+static constexpr int MAX_ITER = 50;
+static constexpr double EPS = 1e-13;
```
matz-e, line 38: Why the change here?

src/solver/newton/newton.hpp
```diff
@@ -110,7 +121,7 @@ EIGEN_DEVICE_FUNC int newton_solver_small_N(Eigen::Matrix<double, N, 1>& X,
     while (++iter < max_iter) {
         functor(X, F, J);
         double error = F.norm();
```
JCGoran, line 123:
```suggestion
```

Think we don't need this one anymore.

matz-e (approved):
Looks sensible.

src/solver/newton/newton.hpp
1uc: The one for MAX_ITER is a compromise between getting closer to NRN and allowing infinite guessing. If Newton converges it converges extremely fast, in every iteration the number of `0` doubles. The problem tends to be that it doesn't converge. However, if it doesn't more iterations are unlikely to help. Either one's stuck just above EPS, or it's paying ping pong, or it's diverging.

The EPS is needed to make the tests pass the check that the residual is under a certain threshold.

matz-e: Was just wondering why you lowered the EPS. Thanks for elaborating on the much lower MAX_ITER.

src/solver/newton/newton.hpp
1uc: Thank you, gone.

PR #1407, opened 2024-08-22
Rename FUNCTION/PROCEDURE arguments.

1uc:
The NMODL language supports hiding variable names, e.g.

    FUNCTION foo(cai) {
      printf("%e", cai);
    }

should print the first argument passed to `foo`. Even if the
the MOD files uses the `ca` ion and the interior concentration
in particular.

To avoid the issue safely, we rename all function arguments, by
prefixing them with `_l`.


1uc:
> In the comment you wrote that you prefix with `_arg_` but it seems to be `_l`

Correct, I've updated the description. This is related to VERBATIM code. I started with `_arg_` because there was something similar nearby. However, the NOCMODL convention would be `_l`. Hence, there is VERBATIM code that knows that the arguments name is `_l{var}`. If we use `_arg_{var}` then the VERBATIM massaging code that's used for CoreNEURON codegen, doesn't massage the `_l{var}` into the right shape and the compiler errors out because of undefined variables. If we use `_l{var}` we get lucky and things seem to work out nicely.

1uc:
The generated code is available also as a (fake) PR: https://github.com/BlueBrain/nmodl-references/pull/22/files

matz-e (approved):
Looks sensible.

PR #1409, opened 2024-08-23
Remove weak, brittle tests.

1uc:
Due to their volume these tests prevent all change, even sensible change.

The checks are mostly performed as string comparisons where the strings are in a very specific, volatile, internal representation. There's some code to sometimes convert the string representation into an abstract representation which is fed into SymPy. However, that doesn't save us from editing 1000s of lines of code when we make trivial changes that aren't/can't be judged as equivalent.

The things these tests prevent are:

  * Transposing J,
  * `X = X_old + dt * dX` -> `(X - X_old)/dt = dX`,
  * `(X - X_old)/dt = dX / vol` -> `vol*(X - X_old)/dt = dX`,
  * name changes,
  * whitespace fixes.

Finally, these tests do not provide any reassurance that the generated code is functionally correct. At best it just confirms that it outputs what the author expected it to output.


matz-e (approved):
Approval conditional on a better replacement.

PR #1414, opened 2024-08-28
Get voltage via `prop->node`.

1uc:
When calling functions directly, don't have access to the `NrnThread`
and therefore can't get node properties, like the voltage from there.

The solution is to create a link from the Prop to the Node. Then use
that link to figure out the node properties.

The trick to make the two cases uniform is the same as we use for
instance data, we create pointers to array of length one (by taking the
address of the element) and setting `_iml`/`id` to `0`.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -822,10 +822,6 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
         // TODO implement these when needed.
     }
 
-    if (!info.vectorize && !info.top_local_variables.empty()) {
-        throw std::runtime_error("Not implemented, global vectorize something.");
-    }
-
     if (!info.thread_variables.empty()) {
         size_t prefix_sum = 0;
         for (size_t i = 0; i < info.thread_variables.size(); ++i) {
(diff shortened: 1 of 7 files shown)


test/usecases/global/test_non_threadsafe.py
```diff
@@ -0,0 +1,33 @@
...
+
+    instance = s(0.5).non_threadsafe
+
+    # Check INITIAL values.
+    assert instance.get_parameter() == 41.0
+    assert instance.get_gbl() == 42.0
+    assert instance.get_top_local() == 43.0
+
+    # Check reassigning a value. Top LOCAL variables
+    # are not exposed to HOC/Python.
```
matz-e, line 24: Could that be tested `with pytest.raises`?

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -834,6 +848,15 @@ void CodegenNeuronCppVisitor::print_mechanism_global_var_structure(bool print_in
         }
     }
 
+    if (!info.vectorize && !info.top_local_variables.empty()) {
+        for (size_t i = 0; i < info.top_local_variables.size(); ++i) {
+            const auto& var = info.top_local_variables[i];
+            std::cout << get_name(var) << std::endl;
```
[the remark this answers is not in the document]
1uc: ty, gone.

test/usecases/global/test_non_threadsafe.py
1uc: Yes, good point. However, nicely systematically testing visibility can be a separate task.

1uc: https://github.com/BlueBrain/nmodl/issues/1421

(however, not with `pytest`, because I absolutely want to be able to run these without pytest.)

matz-e (approved):

PR #1419, opened 2024-09-02
Reapply "Newton error propagation based convergence check."

1uc:
Second attempt at implementing Newton error propagation
based convergence check.

Adds `EIGEN_DEVICE_FUNC` to `is_converged` for GPU builds.

### DIFF
--- a/src/solver/newton/newton.hpp
+++ b/src/solver/newton/newton.hpp
@@ -34,8 +34,22 @@ namespace newton {
  * @{
  */
 
-static constexpr int MAX_ITER = 1e3;
-static constexpr double EPS = 1e-12;
+static constexpr int MAX_ITER = 50;
+static constexpr double EPS = 1e-13;
+
+template <int N>
+bool is_converged(const Eigen::Matrix<double, N, 1>& X,
+                  const Eigen::Matrix<double, N, N>& J,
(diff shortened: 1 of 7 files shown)


1uc:
CI went green for NRN with this PR: https://github.com/neuronsimulator/nrn/pull/3068

src/solver/newton/newton.hpp
```diff
@@ -34,8 +34,33 @@ namespace newton {
  * @{
  */
 
-static constexpr int MAX_ITER = 1e3;
-static constexpr double EPS = 1e-12;
+static constexpr int MAX_ITER = 50;
```
pramodk, line 37: Don't have relevant expertise and hence curious about which aspects allow much lower max iterations.

pramodk (approved):
LGTM

src/solver/newton/newton.hpp
1uc: NEURON uses 20.

If Newton converges, the number of `0`s in the error doubles with each iteration, i.e. from 0.1 error it takes `3` iterations to hit `1e-14`. Unlike stochastic gradient descent, if it doesn't converge, there's no nothing to make it eventually converge.

PR #1424, opened 2024-09-10
Implement CONSTRUCTOR/DESTRUCTOR.

1uc:


test/usecases/constructor/test_ctor.py
```diff
@@ -0,0 +1,36 @@
...
+    check_ctor(12, "ctor")
+
+
+def test_art_ctor():
+    check_ctor(32, "art_ctor")
+
+
+if __name__ == "__main__":
+    test_ctor()
+    test_art_ctor()
```
pramodk, line 36: Note relevant but just to mention: remember coreneuron days when there was no easy way to do tests like this. Especially when coreneuron was a completely different project and only file mode was supported.

So much nicer now to have such unit tests!!

pramodk (approved):

test/usecases/constructor/test_ctor.py
1uc: Soon, we'll even have a reasonably complete basic set of tests. We could eventually think of adding something to enable running them (or some of them) with CoreNEURON too.

PR #1432, opened 2024-09-16
Fix `nrn_ghk`.

1uc:
The NEURON version of `nrn_ghk` has four arguments, unlike the
CoreNEURON variation which as five.


JCGoran (approved):
LGTM, nice find!

PR #1437, opened 2024-09-18
Support `NET_RECEIVE{ INITIAL { ... }}`.

1uc:


JCGoran:
I think `CMakeLists.txt` needs to be updated as well since the diff of the refs is empty: https://github.com/BlueBrain/nmodl-references/commit/0a1eb3aad93f5309ce7a07837a61fdeebe2a07cf

1uc:
It's registered, but the most recent *commit* didn't change the references. If it doesn't show what you want you can create a fake PR which then shows you "everything". Here:
https://github.com/BlueBrain/nmodl-references/pull/23/files

(These are never merged, they're only there to use Github as a field viewer.)

JCGoran (approved):
> It's registered, but the most recent commit didn't change the references

Okay, makes sense. In that case, it's good to go!

PR #1438, opened 2024-09-19
Add the missing `fornetcon_data` "integer" variable.

1uc:
The semantics for this integer variable exist, but not the variable itself. This also defines the variable.


matz-e (approved):

src/codegen/codegen_coreneuron_cpp_visitor.cpp
```diff
@@ -1426,11 +1426,7 @@ void CodegenCoreneuronCppVisitor::print_mechanism_register() {
                           net_recv_init_arg);
     }
     if (info.for_netcon_used) {
-        // index where information about FOR_NETCON is stored in the integer array
-        const auto index =
-            std::find_if(info.semantics.begin(), info.semantics.end(), [](const IndexSemantics& a) {
-                return a.name == naming::FOR_NETCON_SEMANTIC;
-            })->index;
+        const auto index = position_of_int_var(naming::FOR_NETCON_VARIABLE);
```
JCGoran, line 1429: Probably missing something obvious here, but why was `FOR_NETCON_SEMANTIC` replaced with `FOR_NETCON_VARIABLE` here?

1uc: The one searches `info.semantics` and the other searches `info.integer_variables`.

PR #1444, opened 2024-09-20
Allow `derivimplicit` to use finite differences.

1uc:
In NOCMODL `derivimplicit` uses finite differences to compute each
element of the Jacobian. In stead we try to use SymPy, however, if it
fails, e.g. because it encounters an opaque function, we allow it to use
a finite difference instead.

When SymPy can't compute a derivative analytically, it replaces it
with a `Derivative({func}, ({variable}, {n}))`, which represents the
`n`th derivative of `func` w.r.t. `variable`.

In this PR we crawl SymPy's AST to find `Derivative` objects, and
replace them with a finite difference. The step-with is the same
as used in NEURON for computing the elements of the Jacobian.

Additionally, there's some code to enable computing the `\Delta x`.

### DIFF
--- a/python/nmodl/ode.py
+++ b/python/nmodl/ode.py
@@ -358,10 +358,10 @@ def discretize_derivative(expr):
 
 
 def transform_expression(expr, transform):
-    if expr.args is tuple():
+    if not expr:
         return expr
 
-    args = list(transform_expression(transform(arg), transform) for arg in expr.args)
+    args = (transform_expression(transform(arg), transform) for arg in expr.args)
     return expr.func(*args)
 
 
@@ -372,21 +372,6 @@ def transform_matrix_elements(mat, transform):
             for i in range(mat.cols)
         ]
     )
-
-
-def finite_difference_variables(mat):
-    vars = []
-
-    def recurse(expr):
-        for arg in expr.args:
-            if isinstance(arg, sp.Derivative):
-                var = arg.args[1][0]
-                vars.append((var, finite_difference_step_variable(var)))
-
-    for expr in mat:
-        recurse(expr)
-
-    return vars
 
 
 def needs_finite_differences(mat):
--- a/python/nmodl/ode.py
+++ b/python/nmodl/ode.py
@@ -344,6 +344,55 @@ def solve_lin_system(
     return code, new_local_vars
 
 
+def finite_difference_step_variable(sym):
+    return f"{sym}_delta_"
+
+
+def discretize_derivative(expr):
+    if isinstance(expr, sp.Derivative):
+        x = expr.args[1][0]
+        dx = sp.symbols(finite_difference_step_variable(x))
+        return expr.as_finite_difference(dx)
+    else:
+        return expr
+
+
+def transform_expression(expr, transform):
+    if expr.args is tuple():
+        return expr
+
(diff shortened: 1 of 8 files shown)


python/nmodl/ode.py
```diff
@@ -344,6 +344,55 @@ def solve_lin_system(
...
+    if isinstance(expr, sp.Derivative):
+        x = expr.args[1][0]
+        dx = sp.symbols(finite_difference_step_variable(x))
+        return expr.as_finite_difference(dx)
+    else:
+        return expr
+
+
+def transform_expression(expr, transform):
+    if expr.args is tuple():
```
JCGoran, line 361:
```suggestion
    if not expr.args:
```
An empty tuple is falsy, and using `is` for stuff other than singletons (like `None`) can be error-prone.

python/nmodl/ode.py
```diff
@@ -344,6 +344,55 @@ def solve_lin_system(
...
+    def recurse(expr):
+        for arg in expr.args:
+            if isinstance(arg, sp.Derivative):
+                var = arg.args[1][0]
+                vars.append((var, finite_difference_step_variable(var)))
+
+    for expr in mat:
+        recurse(expr)
+
+    return vars
```
JCGoran, line 389: It's admittedly a bit difficult to parse this; how about this alternative (provided I didn't make any typos):

```suggestion
def finite_difference_variables(mat):
    return [
        ((var := arg.args[1][0]), finite_difference_step_variable(var))
        for expr in mat
        for arg in expr.args
        if isinstance(arg, sp.Derivative)
    ]
```

1uc: ah, we don't need this at all.

python/nmodl/ode.py
```diff
@@ -344,6 +344,55 @@ def solve_lin_system(
...
+        return expr.as_finite_difference(dx)
+    else:
+        return expr
+
+
+def transform_expression(expr, transform):
+    if expr.args is tuple():
+        return expr
+
+    args = list(transform_expression(transform(arg), transform) for arg in expr.args)
```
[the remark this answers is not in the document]
1uc: The list is form debugging, we don't need it at all. The generator works.

python/nmodl/ode.py
1uc: Spec promises it's an empty something. Way too many things are falsey, including `0` to make that mistake.

python/nmodl/ode.py
1uc: ... and both a buggy, neither actually descends recurses down the AST.

test/usecases/solve/test_finite_difference.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_finite_difference():
+    nseg = 1
+
+    s = h.Section()
+    s.insert("finite_difference")
+    s.nseg = nseg
+
+    x_hoc = h.Vector().record(getattr(s(0.5), f"_ref_x_finite_difference"))
```
JCGoran, line 13:
```suggestion
    x_hoc = h.Vector().record(s(0.5)._ref_x_finite_difference)
```

Either this, or just remove the `f` since no value is being replaced in the string.

JCGoran (approved):
LGTM, some minor comments.

PR #1445, opened 2024-09-20
Format MOD files.

1uc:
and list the usecase directories in order.


JCGoran:
Is there a formatting tool for these you are using, or is it all by hand?

1uc:
Vim is the best formatting tool!

matz-e (approved):
I'm sure you could run (neo)vim in headless mode to automatically format these (and check in the CI) :)

pramodk (approved):

PR #1457, opened 2024-09-23
Implement FUNCTION_TABLE.

1uc:


1uc:
Only the two most recent commits are "novel". We've created PRs for everything else.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -642,6 +685,10 @@ std::string CodegenNeuronCppVisitor::get_variable_name(const std::string& name,
         // not printing a NET_RECEIVE block.
     }
 
+    if (stringutils::starts_with(name, "_ptable_")) {
```
pramodk, line 688: Just for my own understading: I was staring at this a bit to understand if the "hardcoding" is relevant: Is this for VERBATIM block? Or something else?

pramodk (approved):
LGTM

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: Right, this is part is annoying. If you look at the test we can use FUNCTION_TABLE without a section. However, for historical reasons the global variables are part of the Instance struct. No section, no instance and hence no valid instance struct.

Therefore to avoid segfaults, we need to use the static variable `global_struct_instance()`; and can't simply rely on the usual way of translating a global variable into their C++ implementation.

I'm not aware of a different way of recognizing the global variables `_ptable_{}` for FUNCTION_TABLES.

PR #1460, opened 2024-09-24
Refactor to use `print_function_table_call`.

1uc:
Introduces a way to customize (CoreNEURON vs. NEURON) calling function tables.

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -166,6 +166,13 @@ bool CodegenCppVisitor::defined_method(const std::string& name) const {
     return function && function->has_any_property(properties);
 }
 
+bool CodegenCppVisitor::is_function_table_call(const std::string& name) const {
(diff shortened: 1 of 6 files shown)


src/codegen/codegen_cpp_visitor.cpp
```diff
@@ -166,6 +166,13 @@ bool CodegenCppVisitor::defined_method(const std::string& name) const {
     return function && function->has_any_property(properties);
 }
 
+bool CodegenCppVisitor::is_function_table_call(const std::string& name) const {
+    auto it = std::find_if(info.function_tables.begin(),
```
JCGoran, line 170: `std::any_of` would also work here (saving a grand total of 1 line 😄)

JCGoran (approved):

src/codegen/codegen_cpp_visitor.cpp
1uc: Thank you, there's plenty more in other parts. We definitely need something to tame all the `std::find`s.

PR #1462, opened 2024-09-24
Fix POINTER variables.

1uc:
Both the MOD file and NEURON independently number the pointer variables; and then use their respective order to access the `dparam` array. In order to make the two match we must ensure that all pointer variables are consecutive in `dparams` and pass their offset, i.e. the index of the first POINTER variable, to `{point,}_register_mech`.


pramodk (approved):
LGTM

PR #1463, opened 2024-09-24
Implement `nrn_pointing`.

1uc:
It seem `nrn_pointing` is a special builtin function that requires the
pointer, not the dereferenced value of a pointer.

### DIFF
--- a/src/codegen/codegen_cpp_visitor.cpp
+++ b/src/codegen/codegen_cpp_visitor.cpp
@@ -484,6 +484,11 @@ void CodegenCppVisitor::print_function_call(const FunctionCall& node) {
         function_name = method_name(name);
     }
 
+    if (is_nrn_pointing(name)) {
+        print_nrn_pointing(node);
+        return;
+    }
+
     if (is_net_send(name)) {
         print_net_send_call(node);
         return;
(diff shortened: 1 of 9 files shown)


1uc:
Requires #1462.

src/codegen/codegen_cpp_visitor.cpp
```diff
@@ -524,6 +529,18 @@ void CodegenCppVisitor::print_function_call(const FunctionCall& node) {
     printer->add_text(')');
 }
 
+void CodegenCppVisitor::print_nrn_pointing(const ast::FunctionCall& node) {
+    const auto& args = node.get_arguments();
+    if (args.size() != 1) {
+        throw std::runtime_error(
+            fmt::format("nrn_pointing excepts exactly one argument, got: {}", args.size()));
+    }
```
pramodk, line 537: just a nitpick: checks like this are typically done in [the semantic analysis visitor](https://github.com/BlueBrain/nmodl/blob/2f448528977bd8b60e552123fc7195bd35cecf2b/src/visitors/semantic_analysis_visitor.hpp#L39).

pramodk (approved):
LGTM

src/codegen/codegen_cpp_visitor.cpp
1uc: ty, changed.

PR #1464, opened 2024-09-24
Remove `hoc_nrnpointerindex`.

1uc:
The implementation of `hoc_nrnpointerindex` is incorrect; and wont be needed anymore after merging #1462.


pramodk (approved):
LGTM

PR #1467, opened 2024-09-24
Add `CvodeBlock` and `CvodeVisitor`

JCGoran:
Part 1 of #1399.
Adds the CVODE block since NMODL solves the DERIVATIVE block in-place, and we lose information about it. Also adds a visitors which actually replaces statements of the form `x' = f(x)` into `Dx = f(x)`.


src/visitors/derivative_original_visitor.hpp
```diff
@@ -0,0 +1,64 @@
...
+ * `DERIVATIVE_ORIGINAL_FUNCTION` block.
+ *
+ * If \ref SympySolverVisitor runs successfully, it replaces the original
+ * solution. This block is inserted before that to prevent losing access to
+ * information about the block.
+ */
+class DerivativeOriginalVisitor: public AstVisitor {
+  private:
+    /// The copy of the derivative block we are solving
+    ast::DerivativeBlock* der_block_function = nullptr;
```
1uc, line 39: Why is this a `_function`?

src/visitors/derivative_original_visitor.hpp
```diff
@@ -0,0 +1,64 @@
...
+    /// true while visiting differential equation
+    bool differential_equation = false;
+
+    /// global symbol table
+    symtab::SymbolTable* program_symtab = nullptr;
+
+    /// visiting derivative block
+    bool derivative_block = false;
+
+    ast::AstNodeType node_type = ast::AstNodeType::NODE;
```
1uc, line 50: Is this unused?

src/visitors/derivative_original_visitor.cpp
```diff
@@ -0,0 +1,103 @@
...
+    const auto& lhs = node.get_lhs();
+
+    /// we have to only solve ODEs under original derivative block where lhs is variable
+    if (!derivative_block || !differential_equation || !lhs->is_var_name()) {
+        return;
+    }
+
+    auto name = std::dynamic_pointer_cast<ast::VarName>(lhs)->get_name();
+
+    if (name->is_prime_name() || name->is_indexed_name()) {
```
1uc, line 56: This `if` seems redundant, you can use:
```
if(name->is_prime_name()) { ... }
else if(name->is_indexed_name()) { ... }
```

src/visitors/derivative_original_visitor.cpp
```diff
@@ -0,0 +1,103 @@
...
+            logger->debug("DerivativeOriginalVisitor :: replacing {} with {} on LHS of {}",
+                          to_nmodl(node.get_lhs()),
+                          varname,
+                          to_nmodl(node));
+            auto expr_statement = std::dynamic_pointer_cast<ast::ExpressionStatement>(
+                create_statement(statement));
+            const auto bin_expr = std::dynamic_pointer_cast<const ast::BinaryExpression>(
+                expr_statement->get_expression());
+            node.set_lhs(std::shared_ptr<ast::Expression>(bin_expr->get_lhs()->clone()));
+            // TODO add symbol?
```
1uc, line 83: Please review what's to be done, e.g. remove the comment or implement what's missing, or create a detailed description of what's to be done.

src/language/codegen.yaml
```diff
@@ -87,7 +87,30 @@
...
+                              - name:
+                                  brief: "Name of the derivative block"
+                                  type: Name
+                                  node_name: true
+                                  suffix: {value: " "}
+                              - statement_block:
+                                  brief: "Block with statements vector"
+                                  type: StatementBlock
+                                  getter: {override: true}
+                            brief: "Represents the original, unmodified `DERIVATIVE` block in the NMODL"
```
1uc, line 103: The comment is incorrect, it's not unmodified.

src/language/codegen.yaml
```diff
@@ -87,7 +87,30 @@
                                   type: StatementBlock
                               - finalize_block:
                                   brief: "Statement block to be executed after calling linear solver"
-                                  type: StatementBlock                                  
+                                  type: StatementBlock
+                        - DerivativeOriginalFunctionBlock:
```
1uc, line 91: Why the `Function` in `DerivativeOriginalFunctionBlock`?

1uc (commented):
Two comments:
* The `DerivativeOriginalBlock` is advertised as an unmodified copy. However, it's immediately during the process it's being modified.
* There's no code demonstrating how this works, i.e. no tests or anything similar.

src/visitors/derivative_original_visitor.hpp
JCGoran: Since there's two blocks that will be added: one for `Dx = f(x)` (the `_function` one), and the other one for `Dx = Dx / (1 - dt * J(f))` (the `_jacobian`, will be added in next breakdown of the original PR). I've debated about only adding one block, but the codegen would then require parsing the AST and _modifying_ it, which I feel shouldn't be done at the codegen stage, hence the "two blocks, one visitor" design (the one visitor is here only because it felt pointless to make another one which does almost the same thing).

src/visitors/derivative_original_visitor.hpp
JCGoran: Oops, yes!

src/language/codegen.yaml
JCGoran: NMODL doesn't allow two blocks with the same type (doesn't matter that they may not have the same name), and we will need another block for the transformation involving the Jacobian, so it's basically with that in mind that there's the additional `Function`. Open to suggestions on how to make it shorter though!

src/visitors/derivative_original_visitor.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace nmodl {
+namespace visitor {
+
+/**
+ * \addtogroup visitor_classes
+ * \{
+ */
+
+/**
+ * \class DerivativeOriginalVisitor
```
1uc, line 28: There's still a couple of places where the old term `DerivativeOriginal` exists (including filenames).

src/visitors/derivative_original_visitor.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * `DERIVATIVE_ORIGINAL_FUNCTION` block.
+ *
+ * If \ref SympySolverVisitor runs successfully, it replaces the original
+ * solution. This block is inserted before that to prevent losing access to
+ * information about the block.
+ */
+class CvodeVisitor: public AstVisitor {
+  private:
+    /// The copy of the derivative block we are solving
+    std::shared_ptr<ast::DerivativeBlock> der_block = nullptr;
```
1uc, line 39: The abbreviation is slightly jarring.

src/visitors/derivative_original_visitor.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    /// true while visiting differential equation
+    bool differential_equation = false;
```
1uc, line 42: Are the names for `differential_equation` and `derivative_block` idiomatic? Could they have a better name? Something the hints better at what they are: `inside_*`, `beneath_*`, `is_*`, `printing_*`

test/unit/visitor/derivative_original.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+            DERIVATIVE equation {
+                x' = -x + z * z
+                z' = z * x
+            }
+)";
+        auto ast = run_cvode_visitor(nmodl_text);
+        THEN("CVODE block is added") {
+            auto block = collect_nodes(*ast, {ast::AstNodeType::CVODE_BLOCK});
+            REQUIRE(!block.empty());
```
1uc, line 47: It has too be exactly one I think. At least it looks like that's the assumption further down.

test/unit/visitor/derivative_original.cpp
```diff
@@ -0,0 +1,54 @@
...
+            STATE {x z}
+
+            DERIVATIVE equation {
+                x' = -x + z * z
+                z' = z * x
+            }
+)";
+        auto ast = run_cvode_visitor(nmodl_text);
+        THEN("CVODE block is added") {
+            auto block = collect_nodes(*ast, {ast::AstNodeType::CVODE_BLOCK});
```
1uc, line 46: It's a vector of all `CVODE_BLOCK`s. Hence plural, `blocks` (or `cvode_blocks`); then later you'd be free to use `block = block[0]` (not much use).

src/visitors/derivative_original_visitor.cpp
```diff
@@ -0,0 +1,83 @@
...
+    node.visit_children(*this);
+    if (der_block) {
+        auto der_node = new ast::CvodeBlock(der_block->get_name(),
+                                            der_block->get_statement_block(),
+                                            der_block->get_statement_block());
+        node.emplace_back_node(der_node);
+    }
+
+    // re-visit the AST since we now inserted the DERIVATIVE_ORIGINAL block
+    node.visit_children(*this);
```
1uc, line 79: Why do we need to revisit them again?

src/visitors/derivative_original_visitor.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+void CvodeVisitor::visit_program(ast::Program& node) {
+    program_symtab = node.get_symbol_table();
+    node.visit_children(*this);
+    if (der_block) {
+        auto der_node = new ast::CvodeBlock(der_block->get_name(),
+                                            der_block->get_statement_block(),
+                                            der_block->get_statement_block());
+        node.emplace_back_node(der_node);
+    }
```
1uc, line 76: If you don't like the pattern that you need to toggle two booleans there's a way out. The code I marked essentially does a:

            auto blocks = collect_nodes(*ast, {ast::AstNodeType::DERIVATIVE_BLOCK});

Then all you need to do is run the `{}' -> d{}` replacement visitor (which consists of the `visit_binary_expression`) on each block and put them into `CVodeBlock`s and add them to the program.

Then there's no more boolean toggling; and it'll be more explicit about handling multiple DERIVATIVE blocks. BTW, the name for toggling being inside a cvode block is now rather misleading.

src/visitors/derivative_original_visitor.cpp
JCGoran: To modify the CVODE block.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    // re-visit the AST since we now inserted the CVODE block
+    node.visit_children(*this);
+    if (!conserve_equations.empty()) {
+        auto blocks = collect_nodes(node, {ast::AstNodeType::CVODE_BLOCK});
+        auto block = std::dynamic_pointer_cast<ast::CvodeBlock>(blocks[0]);
+        block->get_function_block()->erase_statement(conserve_equations);
+        block->get_diagonal_jacobian_block()->erase_statement(conserve_equations);
+    }
+}
```
1uc, line 151: This visitor does:
1. Collect CONSERVE statements (for later deletion).
2. Recurses once to create an invalid CVODE_BLOCK.
3. Keeps track of which "branch" of the invalid `CVODE_BLOCK` we're in via a counter `block_index`.
4. Keeps track of two additional layers booleans: `in_cvode_block`  and `in_differential_equation`.

When doing a:
``` 
node.visit_children(*this);
// changes to AST.
node.visit_children(*this);
```
it's not obvious which functionality is desired during the first pass; and what's desired during the second pass. It could be doing everything again, or one of the boolean flags turn off some functionality of the first/second pass.

I can't quickly tell if it's supposed to handle multiple DERIVATIVE blocks. ("No." is a good answer, see #1501). Note, it doesn't matter that it can't handle both, the issue is that it's hard to tell if it tries to.

I think there's room to make the intent/algorithm more visible. To my understanding the following is happening:
1. Find the one unique DERIVATIVE block.
2. Create a CVODE_BLOCK.
3. Insert the CVODE_BLOCK into the program.

The key is that you create a valid `CVODE_BLOCK` from the DERIVATIVE block. In pseudo code, I think you can do:
```
auto derivative_block = collect_nodes(node, "DERIVATIVE");
auto cvode_block = make_cvode_block(derivative_block);
node.emplace_back_node(derivative_block);
```
This change means, we can see the high-level idea in `CvodeVisitor`. The next problem is how to structure `make_cvode_block`.

I would propose something like:
```
CvodeBlock make_cvode_block(Derivative& derivative) {

    auto Dx_block = derivative.clone();
    Dx_block.accpet(CvodeDBlockVisitor());
    Dx_block.accept(RemoveConserveVisitor());

    auto J_block = derivative.clone();
    J_block.accept(CvodeJBlockVisitor());
    J_block.accept(RemoveConserveVisitor());

    return CVodeBlock(block_name, Dx_block, J_block);
}
```

Now it's clear that `make_cvode_block` simply creates the CVODE_BLOCK by transforming two copies of the derivative block. The transformation for each block can be written down in a visitor each, those are the poorly named `CvodeDxBlockVisitor` and `CvodeJBlockVisitor` (I suspect there's better names for both).

To get rid of CONSERVE blocks, you can (implement and) run a `RemoveConserveVisitor` on whatever node needs to be stripped.

1uc (commented):
Nice progress, good high-level documentation. It looks like the functionality has been implemented. Personally, I restructure the code to create better separation of concerns.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <optional>
+#include <utility>
+
+namespace pywrap = nmodl::pybind_wrappers;
+
+namespace nmodl {
+namespace visitor {
+
+static int get_index(const ast::IndexedName& node) {
+    return std::stoi(to_nmodl(node.get_length()));
```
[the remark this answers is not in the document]
JCGoran: `get_length` returns an `Expression` instance, which AFAIK has no "easy" way of outputting the actual index as an int or a string.

src/pybind/wrapper.cpp
```diff
@@ -186,6 +186,39 @@ except Exception as e:
...
 }
 
+std::tuple<std::string, std::string> call_diff2c(
+    const std::string& expression,
+    const std::string& variable,
+    const std::unordered_map<std::string, int>& indexed_vars) {
+    std::string statements;
+    // only indexed variables require special treatment
+    for (const auto& [var, prop]: indexed_vars) {
+        statements += fmt::format("_allvars.append(sp.IndexedBase('{}', shape=[1]))\n", var);
```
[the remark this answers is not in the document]
JCGoran: From my testing, it seems the actual value passed to `shape` is not that important since these are not the symbols we are diffing against, but just the "other" vars we need to provide so SymPy handles them properly.

src/main.cpp
```diff
@@ -494,10 +495,18 @@ int run_nmodl(int argc, const char* argv[]) {
         const bool sympy_sparse = solver_exists(*ast, "sparse");
 
         if (sympy_conductance || sympy_analytic || sympy_sparse || sympy_derivimplicit ||
-            sympy_linear) {
+            sympy_linear || neuron_code) {
```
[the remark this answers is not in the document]
JCGoran: Updated so it only activates it if there's a `DERIVATIVE` block found.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+            create_statement(statement));
+        const auto bin_expr = std::dynamic_pointer_cast<const ast::BinaryExpression>(
+            expr_statement->get_expression());
+        node.set_rhs(std::shared_ptr<ast::Expression>(bin_expr->get_rhs()->clone()));
+    }
+};
+
+
+void CvodeVisitor::visit_program(ast::Program& node) {
+    auto der_blocks = collect_nodes(node, {ast::AstNodeType::DERIVATIVE_BLOCK});
```
1uc, line 178: Please find a different name for `der_block`.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if (!der_blocks.empty()) {
```
1uc, line 179: There's three cases here:
1. No derivative blocks: just return.
2. More than one block: crash loudly.
3. Continue to create the cvode block.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    auto variable = std::make_pair(node->get_node_name(), std::optional<int>());
+    if (node->is_indexed_name()) {
+        variable.second = std::optional<int>(
+            get_index(*std::dynamic_pointer_cast<const ast::IndexedName>(node)));
+    }
+    return variable;
+}
+
+static std::unordered_map<std::string, int> get_name_map(const ast::Expression& node,
+                                                         const std::string& name) {
```
1uc, line 49: What is a 'name_map'? The second argument is called `name`, but I can't figure out what's it's the name of. From the code it's an exception, i.e. if the indexed name is called `name` then don't do something. Is there a better name for this one?

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    std::unordered_map<std::string, int> name_map;
+    // all of the "reserved" symbols
+    auto reserved_symbols = get_external_functions();
+    // all indexed vars
+    auto indexed_vars = collect_nodes(node, {ast::AstNodeType::INDEXED_NAME});
+    for (const auto& var: indexed_vars) {
+        if (!name_map.count(var->get_node_name()) && var->get_node_name() != name &&
+            std::none_of(reserved_symbols.begin(), reserved_symbols.end(), [&var](const auto item) {
+                return var->get_node_name() == item;
+            })) {
```
1uc, line 59: Please be explicit about converting integers to booleans: e.g. `count > 0`. I think this line is long enough to allow breaking up, e.g. give the third clause a name please.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+            logger->debug(
+                "CvodeVisitor :: adding INDEXED_VARIABLE {} to "
+                "node_map",
+                var->get_node_name());
+            name_map[var->get_node_name()] = get_index(
```
1uc, line 64: This is the fifth time we're typing `var->get_node_name()` maybe it should be given a name, e.g. `var_name`.

1uc (commented):
Looks quite good, a couple of last minor comments.

src/visitors/cvode_visitor.cpp
JCGoran: This also fails for the steady state usecase, since one of the intermediate mod files has:

```plaintext
DERIVATIVE state {
    LOCAL kf0_, kb0_
    kf0_ = kf
    kb0_ = kb
    CONSERVE Z = (8.0*volB+1.0*volC-(volB*1*Y))/(volC*1)
    X' = ((-1*(kf0_*X*Y-kb0_*Z)))/(volA)
    Y' = ((-1*(kf0_*X*Y-kb0_*Z)))/(volB)
    Z' = ((1*(kf0_*X*Y-kb0_*Z)))/(volC)
}

DERIVATIVE state_steadystate {
    dt = 1000000000
    LOCAL kf0_, kb0_
    kf0_ = kf
    kb0_ = kb
    CONSERVE Z = (8.0*volB+1.0*volC-(volB*1*Y))/(volC*1)
    X' = ((-1*(kf0_*X*Y-kb0_*Z)))/(volA)
    Y' = ((-1*(kf0_*X*Y-kb0_*Z)))/(volB)
    Z' = ((1*(kf0_*X*Y-kb0_*Z)))/(volC)
}
```

I could add a `_steadystate` detector, but is it really necessary?

1uc: It feels like an easy improvement. Now you know that you're making the assumption right here. In the future a different dev might be tasked with allowing multiple blocks, they'll remove the first check; and then need to search for hours why there's somethings just not showing up. If you add the check it'll immediately pop out to that future dev. It also serves as a reminder that this doesn't work (which is perfectly fine) and isn't intended to work for multiple derivative block, therefore the reader can put themselves into the same mindset.

If a dev dares add a derivative block, this part might start to fail spuriously, depending on in which order the blocks are 'collect'ed.

src/visitors/cvode_visitor.cpp
JCGoran: Turns out we don't need the index at all, so I changed this to a more appropriate name `get_indexed_variables`.

src/visitors/cvode_visitor.cpp
JCGoran: Done.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+        if (program_symtab->lookup(varname) == nullptr) {
+            auto symbol = std::make_shared<symtab::Symbol>(varname, ModToken());
+            symbol->set_original_name(name->get_node_name());
+            program_symtab->insert(symbol);
+        }
+
+        auto rhs = node.get_rhs();
+        // all indexed variables (need special treatment in SymPy)
+        auto name_map = get_indexed_variables(*rhs, name->get_node_name());
```
1uc, line 156: ?

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,218 @@
...
+    if (node->is_indexed_name()) {
+        variable.second = std::optional<int>(
+            get_index(*std::dynamic_pointer_cast<const ast::IndexedName>(node)));
+    }
+    return variable;
+}
+
+/// set of all indexed variables not equal to ``name``
+static std::unordered_set<std::string> get_indexed_variables(const ast::Expression& node,
+                                                             const std::string& name) {
```
1uc, line 50: the name still hasn't improved. this can easily be made trivial to understand by using a better name.

src/visitors/cvode_visitor.cpp
```diff
@@ -0,0 +1,218 @@
...
+                 std::back_inserter(derivative_blocks_copy),
+                 not_steadystate);
+    if (derivative_blocks_copy.size() > 1) {
+        auto message = "CvodeVisitor :: cannot have multiple DERIVATIVE blocks";
+        logger->error(message);
+        throw std::runtime_error(message);
+    }
+
+    auto derivative_block = std::dynamic_pointer_cast<ast::DerivativeBlock>(
+        derivative_blocks_copy[0]);
```
1uc, line 199:
```suggestion
    auto derivative_block = get_derivative_block();
```

src/visitors/cvode_visitor.cpp
JCGoran: Not really sure what's a better name for getting indexed variables than `get_indexed_variables`, but I am open to suggestions.

JCGoran: Then I think I misunderstood the original comment 😅 Just to clarify: the variable `name` should be `ignored_name`, correct?

PR #1473, opened 2024-09-27
Use forward declaration for nmodl lexer.

1uc:
We identified `nmodl_lexer.hpp` and `nmodl_driver.hpp` to be expensive to parse.

    **** Expensive headers:

    37155 ms: src/lexer/nmodl_lexer.hpp (included 47 times, avg 790 ms), included via:
      41x: nmodl_driver.hpp
      6x: <direct include>

    34970 ms: src/parser/nmodl_driver.hpp (included 47 times, avg 744 ms), included via:
      47x: <direct include>

The problem is easily avoided by using a forward declaration.

### DIFF
--- a/src/parser/nmodl_driver.hpp
+++ b/src/parser/nmodl_driver.hpp
@@ -16,7 +16,7 @@
 #include <string>
 #include <unordered_map>
 
-#include "ast/ast.hpp"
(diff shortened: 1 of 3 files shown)


pramodk (approved):
👌

PR #1478, opened 2024-09-30
Run SonarSource on demand.

1uc:


1uc:
Shaves off 2 minutes of build time.

pramodk (commented):
LGTM. Mentioning this in `CONTRIBUTING.rst` might be helpful?

1uc:
BTW, it's now faster than the Windows build; and about as fast as the documentation CI. Might be worth just running it to ensure it doesn't rot away.

1uc:
Let's leave it in place for now. When it's the last thing to go green, we can turn it off.

PR #1484, opened 2024-09-30
Use finite differences in `differentiate2c` for implicit functions

JCGoran:


test/unit/ode/test_ode.py
```diff
@@ -100,6 +100,16 @@ def test_differentiate2c():
         "g",
     )
 
+    assert _equivalent(
+        differentiate2c(
+            "-f(x)",
+            "x",
+            {},
+        ),
+        "1000.0*f(x - 0.00050000000000000001) - 1000.0*f(x + 0.00050000000000000001)",
```
JCGoran, line 109: Is there a more robust way of comparing the equality here? The SymPy codgen can change at some version and the order of the operands could change, making this test rather brittle.

1uc: Yes, you make it generate a function that you can run.

JCGoran: Feeding the output back to SymPy seems to work.

test/unit/ode/test_ode.py
```diff
@@ -100,6 +101,30 @@ def test_differentiate2c():
...
 
+    result = differentiate2c(
+        "-f(x)",
+        "x",
+        {},
+    )
+    # instead of comparing the expression as a string, we convert the string
+    # back to an expression and insert various functions
+    for function in [sp.sin, sp.exp, sp.tanh]:
+        for value in np.linspace(-5, 5, 100):
```
1uc, line 112: Not sure this tests anything new. One of them and 10 points is very similar.

1uc (approved):

python/nmodl/ode.py
```diff
@@ -643,15 +643,27 @@ def differentiate2c(expression, dependent_var, vars, prev_expressions=None):
     # differentiate w.r.t. x
     diff = expr.diff(x).simplify()
 
+    # could be something generic like f'(x), in which case we use finite differences
+    if needs_finite_differences(diff):
+        diff = (
+            transform_expression(diff, discretize_derivative)
+            .subs({finite_difference_step_variable(x): 1e-3})
```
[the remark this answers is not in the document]
1uc: That looks very specific to CVODE, maybe one can delay picking the step size. Okay, if it's a one-sided difference double check that it's not because of side-effects. I've seen the pattern:
```
i1 = f(v + dv);
i0 = f(v);
didv = (i1 - i0)/dv;
```
This is done to get `f(v)` after `f(v + dv)` because `f(v)` sets state (like `dX` variables) and it looked like that state should be for `v` not `v + dv`.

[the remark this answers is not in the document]
JCGoran: Added `stepsize` param (with a default of `1e-3`) to keep it general.

.github/workflows/nmodl-ci.yml
```diff
@@ -16,7 +16,7 @@ on:
 
 env:
   CTEST_PARALLEL_LEVEL: 1
-  PYTHON_VERSION: 3.8
+  PYTHON_VERSION: 3.9
```
JCGoran, line 19:
```suggestion
  PYTHON_VERSION: 3.8
```

Forgot to change this back.

PR #1492, opened 2024-10-02
Remove ARM-based MacOS from CI

JCGoran:
It keeps failing with complaints of wrong arch being used (see for instance https://github.com/BlueBrain/nmodl/actions/runs/11142396798/job/30965250232?pr=1484). I don't have MacOS 14 yet so I can't debug it properly, hence I would like to leave it out for now.


JCGoran:
Usage of numpy was avoided so this is not needed.

PR #1498, opened 2024-10-03
Improve COMPARTMENT names & documentation.

1uc:


alkino (approved):
So usefull, thanks for that!

PR #1502, opened 2024-10-07
Refactor printing setup code in entrypoints.

1uc:
The entrypoints into the MOD file from NEURON typically require setting up some variable, e.g. `_lmr`, `inst`, `node_data`, `_thread_vars`, etc.

The code has roughly two favours:

* From fully initialized NEURON, i.e. a Memb_list exists and the model is sorted. All the compute function have this type of entrypoint.

* From a `Prop`. Usually this happens when we don't know that the model is sorted, or that initialization has happened. The HOC/Python bindings of FUNCTION/PROCEDUREs use this type of entrypoint.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.hpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.hpp
@@ -507,8 +507,8 @@ class CodegenNeuronCppVisitor: public CodegenCppVisitor {
     /**
      * Prints setup code for entrypoints NEURON.
      *
-     * See `print_entrypoint_setup_code`. This variation should be used when one only has
-     * access to a `Prop`, but not the full `Memb_list`.
+     * See `print_entrypoint_setup_code_from_memb_list`. This variation should be used when one only
+     * has access to a `Prop`, but not the full `Memb_list`.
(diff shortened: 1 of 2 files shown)


alkino (approved):

src/codegen/codegen_neuron_cpp_visitor.hpp
```diff
@@ -492,6 +492,26 @@ class CodegenNeuronCppVisitor: public CodegenCppVisitor {
...
+     *
+     * This variation prints the fast entrypoint, where NEURON is fully initialized and setup.
+     */
+    void print_entrypoint_setup_code_from_memb_list();
+
+
+    /**
+     * Prints setup code for entrypoints NEURON.
+     *
+     * See `print_entrypoint_setup_code`. This variation should be used when one only has
```
JCGoran, line 510: Should this be `print_entrypoint_setup_code_from_memb_list`? I don't find any instances of `print_entrypoint_setup_code`...

1uc: ty, that was it's old name.

PR #1513, opened 2024-10-11
Fix FUNCTION_TABLE with POINT_PROCESS/ARTIFICIAL_CELL.

1uc:
There's three bugs that are fixed:
* make printing the function signature more consistent. This then results in the correct signature for POINT_PROCESSES.
* while we did register the "HOC" variation, i.e. `h.tau2_{mech_name}` would work, we didn't register the Python variations meaning `s(0.5).{mech_name}.tau2` didn't work.
* the `use_tau2` test had an implicit dependency on a previous test to set the dependency. After the first test ran the last instance of the mechanism went out of scope and the mechanism itself got collected. Leading to invalid memory access.

### DIFF
--- a/test/usecases/function_table/test_function_table.py
+++ b/test/usecases/function_table/test_function_table.py
@@ -72,6 +72,11 @@ def check_2d(make_inst, mech_name):
 
     inst = make_inst(s)
     set_table, eval_table = make_callbacks(inst, "tau2", mech_name)
+    eval_use_table = make_callable(inst, "use_tau2", mech_name)
+
+    if inst is None:
+        setdata = getattr(h, f"setdata_{mech_name}")
+        setdata(s(0.5))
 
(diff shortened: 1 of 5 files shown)


test/usecases/function_table/test_function_table.py
```diff
@@ -4,43 +4,80 @@
 import scipy
 
 
-def test_constant_1d():
```
JCGoran, line 7: Could the `test_*` methods be put back? I would prefer if anything validating NMODL functionality/outputs be called `test_*`.

JCGoran (approved):

test/usecases/function_table/test_function_table.py
1uc: Only in the form `test_everything` (which can be done); or through really annoying repetition.

PR #1515, opened 2024-10-14
Add longitudinal diffusion usecase to suite

JCGoran:
I think we forgot it at some point.


test/usecases/CMakeLists.txt
```diff
@@ -28,7 +28,8 @@ set(NMODL_USECASE_DIRS
     steady_state
     suffix
     table
-    useion)
+    useion
+    longitudinal_diffusion)
```
1uc, line 32: Thank you, I recently sorted them :)

1uc (approved):

PR #1516, opened 2024-10-14
Eigen solvers in ARTIFICIAL_CELLS can't use voltage.

1uc:
For ARTIFICIAL_CELLs we don't make `v` a parameter of
the "functor" ctor.

### DIFF
--- a/test/usecases/nonlinear/test_nonlin.py
+++ b/test/usecases/nonlinear/test_nonlin.py
@@ -1,15 +1,24 @@
 from neuron import h, gui
 
-
-def test_nonlin():
+def check_solve(make_inst):
     s = h.Section()
     s.insert("nonlin")
-    inst = s(0.5).nonlin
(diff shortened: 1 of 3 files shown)


test/usecases/nonlinear/test_nonlin.py
```diff
@@ -1,15 +1,25 @@
...
+def check_solve(make_inst):
     s = h.Section()
     s.insert("nonlin")
-    inst = s(0.5).nonlin
+
+    inst = make_inst(s)
 
     x = inst.solve()
     error = inst.residual(x)
     assert abs(error) < 1e-9, f"{x = }, {error = }"
```
JCGoran, line 12: Since you're modifying this file, maybe consider changing this call to `numpy.testing.assert_allclose` or similar.

JCGoran (approved):

PR #1504, opened 2024-10-08
Implement LONGITUDINAL_DIFFUSION.

1uc:

PR #1517, opened 2024-10-14
Add support for diffing against `sympy.Indexed` instances

JCGoran:
If we have an expression (string) that contains an indexed variable (like `x[0]`), we should be able to differentiate w.r.t. it as well.
As explained in the updated docstring, the way to do it is to pass a [`sympy.Indexed`](https://docs.sympy.org/latest/modules/tensor/indexed.html#sympy.tensor.indexed.Indexed) instance as the independent variable (not to be confused with an [`sympy.IndexedBase`](https://docs.sympy.org/latest/modules/tensor/indexed.html#sympy.tensor.indexed.IndexedBase) instance!).
Note that we also need to pass an `sympy.IndexedBase` into the SymPy list of known symbols (i.e. as one of `vars`), otherwise the parser breaks.

For reference, we added mod files with indexed vars in #1504.

### DIFF
--- a/python/nmodl/ode.py
+++ b/python/nmodl/ode.py
@@ -595,6 +595,11 @@ def differentiate2c(
     If the result coincides with one of the vars, or the LHS of one of
     the prev_expressions, then it is simplified to this expression.
 
+    Note that, in order to differentiate against indexed variables (such as
+    ``x[0]``), you must pass an instance of ``sympy.Indexed`` to
+    ``dependent_var`` (_not_ an instance of ``sympy.IndexedBase``), as well as
+    an instance of ``sympy.IndexedBase`` to ``vars``.
+
     Some simple examples of use:
 
     - ``nmodl.ode.differentiate2c ("a*x", "x", {"a"}) == "a"``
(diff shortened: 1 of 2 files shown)


python/nmodl/ode.py
```diff
@@ -595,6 +595,11 @@ def differentiate2c(
     If the result coincides with one of the vars, or the LHS of one of
     the prev_expressions, then it is simplified to this expression.
 
+    Note that, in order to differentiate against indexed variables (such as
+    ``x[0]``), you must pass an instance of ``sympy.Indexed`` to
+    ``dependent_var`` (_not_ an instance of ``sympy.IndexedBase``), as well as
+    an instance of ``sympy.IndexedBase`` to ``vars``.
+
```
1uc, line 602: This feels like something that's easier to understand by example.

1uc (approved):

python/nmodl/ode.py
JCGoran: Added an example now.

1uc: nice

PR #1518, opened 2024-10-14
Run tests in parallel.

1uc:


1uc:
Tests take 2m16s instead of 3m42s, i.e. two cores.

JCGoran (approved):

PR #1519, opened 2024-10-14
Prevent inlining VERBATIM.

1uc:
We're banning all VERBATIM blocks because the current solution failed to detect:
```
BREAKPOINT {
   SOLVE getonset
}

PROCEDURE getonset() {
   VERBATIM
   return 0;
   ENDVERBATIM
}
```
must not be inlined as:
```
void nrn_state_synp(...) {
// ...
    for (int id = 0; id < nodecount; id++) {
        // VERBATIM
        return 0; 
        // ENDVERBATIM
    }
}
```

### DIFF
--- a/src/visitors/inline_visitor.cpp
+++ b/src/visitors/inline_visitor.cpp
@@ -30,18 +30,10 @@ bool InlineVisitor::can_inline_block(const StatementBlock& block) {
             to_inline = false;
             break;
         }
-        // verbatim blocks with return statement are not safe to inline
-        // especially for net_receive block
+
+        // verbatim blocks (especially with return statements) are not safe to inline
         if (statement->is_verbatim()) {
-            const auto node = dynamic_cast<const Verbatim*>(statement.get());
-            assert(node);
-            auto text = node->get_statement()->eval();
-            parser::CDriver driver;
-            driver.scan_string(text);
-            if (driver.has_token("return")) {
-                to_inline = false;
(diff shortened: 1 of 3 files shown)


src/visitors/inline_visitor.cpp
```diff
@@ -318,12 +310,33 @@ void InlineVisitor::visit_wrapped_expression(WrappedExpression& node) {
...
 
+class CanInlineVisitor: public ConstAstVisitor {
+  private:
+    void visit_verbatim(const ast::Verbatim&) override {
+        is_inlineable = false;
+    }
+
+  public:
+    bool is_inlineable = true;
+};
```
pramodk, line 333: Just want to point out that we also have https://github.com/BlueBrain/nmodl/blob/164ed7ebb15b281c2b1380d4b5fb906ea2cac346/src/visitors/inline_visitor.cpp#L24 

maybe good to clarify when one is supposed to be used over other?

1uc: I implemented this to block it from inlining `SOLVE proc`. However, that inlining happens very differently. Therefore, this is somewhat useless. I can't quite make up my mind if not allowing inlining anything with a VERBATIM block is a good idea; or if it's needless pessimization.

1uc: Until someone does:
```
#define return_ return
```
and it all breaks.

PR #1521, opened 2024-10-15
SOLVE a PROCEDURE.

1uc:
From inspecting NOCMODL generated code it seems that:
```
SOLVE proc
```
where `proc` is some procedure, simply calls the procedure and checks if it returned an error code. This PR only implements the call part and leaves the checking part for the future.

### DIFF
--- a/src/visitors/solve_block_visitor.cpp
+++ b/src/visitors/solve_block_visitor.cpp
@@ -9,6 +9,7 @@
 
 #include "utils/fmt.h"
 #include <cassert>
+#include <memory>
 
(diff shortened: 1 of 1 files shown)


1uc:
It should be checking the return code of the procedure. However, that requires some slightly lengthy changes, which I'd like to postpone. This allows us to compile mod files found in the nrn tests and should be correct for any correct runs.

pramodk (approved):
LGTM

1uc:
See #1526 for followup work.

PR #1522, opened 2024-10-18
Refactor HOC/Python wrapper printing.

1uc:
The method is first split into three parts:
* determining the signature,
* printing the setup code,
* calling the implementation.

Additionally we rename the function, because it doesn't print the function body, it prints the whole wrapper function.

### DIFF
--- a/src/codegen/codegen_neuron_cpp_visitor.cpp
+++ b/src/codegen/codegen_neuron_cpp_visitor.cpp
@@ -285,7 +285,6 @@ void CodegenNeuronCppVisitor::print_function_procedure_helper(const ast::Block& 
 void CodegenNeuronCppVisitor::print_hoc_py_wrapper_call_impl(
     const ast::Block* function_or_procedure_block,
     InterpreterWrapper wrapper_type) {
-
     const auto block_name = function_or_procedure_block->get_node_name();
 
     const auto get_func_call_str = [&]() {
@@ -298,7 +297,6 @@ void CodegenNeuronCppVisitor::print_hoc_py_wrapper_call_impl(
         func_call.append(")");
         return func_call;
     };
-
 
     printer->add_line("double _r = 0.0;");
     if (function_or_procedure_block->is_function_block()) {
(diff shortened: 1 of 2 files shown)


src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -282,21 +282,43 @@ void CodegenNeuronCppVisitor::print_function_procedure_helper(const ast::Block&
...
     const ast::Block* function_or_procedure_block,
     InterpreterWrapper wrapper_type) {
-    if (info.point_process && wrapper_type == InterpreterWrapper::Python) {
-        return;
-    }
     const auto block_name = function_or_procedure_block->get_node_name();
-    if (wrapper_type == InterpreterWrapper::HOC) {
-        printer->fmt_push_block("{}", hoc_function_signature(block_name));
+
+    const auto get_func_call_str = [&]() {
```
JCGoran, line 290: This could be its own function. A docstring would also be useful.

src/codegen/codegen_neuron_cpp_visitor.hpp
```diff
@@ -244,9 +244,32 @@ class CodegenNeuronCppVisitor: public CodegenCppVisitor {
...
+
 
+    /** Print the code that calls the impl from the HOC/Py wrapper.
+     */
+    void print_hoc_py_wrapper_call_impl(const ast::Block* function_or_procedure_block,
+                                        InterpreterWrapper wrapper_type);
+
+    /** Return the wrapper signature.
+     *
+     * Everything without the { or ;.
```
JCGoran, line 269: An example would be useful here.

src/codegen/codegen_neuron_cpp_visitor.cpp
```diff
@@ -282,21 +282,43 @@ void CodegenNeuronCppVisitor::print_function_procedure_helper(const ast::Block&
...
+            func_call.append(fmt::format(", *getarg({})", i + 1));
+        }
+        func_call.append(")");
+        return func_call;
+    };
+
+    printer->add_line("double _r = 0.0;");
+    if (function_or_procedure_block->is_function_block()) {
+        printer->add_indent();
+        printer->fmt_text("_r = {};", get_func_call_str());
```
JCGoran, line 304: Out of curiosity, why not use `fmt_line` here?

JCGoran (approved):

src/codegen/codegen_neuron_cpp_visitor.cpp
1uc: It feels like it's (unfortunately) tailor made for this situation. Notice how the arguments are not `a, b, c` but are grabbed from HOC, i.e. `getarg(1), getarg(2), getarg(3)`. That reduces the reusability quite a bit. We could try to make that part customizable, but we'll need to worry about `f(a + b, c)` which requires traversing the AST and printing as one goes.

src/codegen/codegen_neuron_cpp_visitor.hpp
1uc, line 269:
```suggestion
     * Everything without the { or ; as an example:
            return_type function_name(<internal_args>, <args>)
            
       were `<internal_args> is the list of arguments required by the
       code to be passed along, while <args> are the arguments in of
       function as they appear in the MOD file.
```

1uc: done

PR #1524, opened 2024-10-18
Implement `stringutils::join_arguments`.

1uc:


src/utils/string_utils.cpp
```diff
@@ -27,5 +27,16 @@ std::string to_string(double value, const std::string& format_spec) {
     return fmt::format(format_spec, value);
 }
 
+std::string join_arguments(const std::string& lhs, const std::string& rhs) {
+    if (lhs == "") {
```
alkino, line 31:
```suggestion
    if (lhs.empty()) {
```

src/utils/string_utils.cpp
```diff
@@ -27,5 +27,16 @@ std::string to_string(double value, const std::string& format_spec) {
...
+        return rhs;
+    } else if (rhs == "") {
```
alkino, line 33:
```suggestion
    } else if (rhs.empty()) {
```

src/utils/string_utils.cpp
1uc: right, a leftover from some desperation fueled debugging.

pramodk (approved):

PR #1529, opened 2024-10-21
Avoid premature segfaults.

1uc:
In NOCMODL functions that only use global data can be called without an instance.


1uc:
Sufficient to get us past some more NRN unit-tests.

1uc:
Link to generated code: https://github.com/BlueBrain/nmodl-references/pull/24

pramodk (approved):

PR #1531, opened 2024-10-23
Fix SymPy errors when using reserved symbols

JCGoran:


1uc (commented):
There already code for renaming, e.g. search for `forbidden_var`. How to the two fit together and why do we need the duplication?

PR #1532, opened 2024-10-25
Handle voltage via instance specific copy.

1uc:
In NOCMODL, the voltage is (when `vectorized`) handled via an instance specific copy. While in CoreNEURON, this is never done, which why the field is named `v_unused`.

This PR always uses `v_unused`. The reason we can't (easily) use the CoreNEURON variation is because sometimes, e.g. CVODE we need to register callbacks, those callbacks have a fixed signature, we can't simply pass through the voltage as an argument, instead we need to retrieve it, e.g. from the instance data.

### DIFF
--- a/test/usecases/voltage/test_voltage.py
+++ b/test/usecases/voltage/test_voltage.py
@@ -51,6 +51,6 @@ def test_state():
 
 
 if __name__ == "__main__":
-    # test_voltage_access()
-    # test_breakpoint()
+    test_voltage_access()
+    test_breakpoint()
     test_state()
--- a/test/usecases/voltage/test_voltage.py
+++ b/test/usecases/voltage/test_voltage.py
@@ -1,0 +1,56 @@
+from neuron import h, gui
+
+import numpy as np
+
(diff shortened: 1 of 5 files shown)


test/usecases/voltage/test_voltage.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_state():
+    # Effectively, the timing when states are computed results in backward Euler.
+    check_ode("state_ode", lambda v, c: (1.0 + c * h.dt) * v)
+
+
+if __name__ == "__main__":
+    # test_voltage_access()
+    # test_breakpoint()
```
pramodk, line 55: to clarify: these can't be tested together? (i.e. don't see a reason for commenting out)

pramodk (approved):

test/usecases/voltage/test_voltage.py
1uc: ty, that wasn't intended; just easier to debug that way.

PR #1533, opened 2024-10-25
Consider NONLINEAR vectorizable.

1uc:


JCGoran (commented):
Could you remind me what `vectorize` refers to again? Is it in the SIMD sense, or something else?

1uc:
VECTORIZE is an NRN term that describes if MOD files have their data arranged in such a manner that vectorization (e.g. SIMD) is (potentially) possible. It affects the API of the registered function, e.g. they don't have access to a lot of things, and IIUC always go through `_ppvar` for everything (even during the simulation phase). Use `nocmodl` to explore how the generated MOD files changes when you add/remove LINEAR from some of the MOD files slated to be removed.

The term is also used to mean "not thread-safe".

The way it's annoying to us is that: a) it removes what we call `v_unused` and uses a static double to cache `v` instead, b) it makes things LINEAR and NONLINEAR not thread-safe even though, when using Eigen there's no issue.

JCGoran (approved):
Makes sense, approved.

PR #1538, opened 2024-10-25
Don't have instance copies of POINTER.

1uc:
The NRN unit-tests fail with a segfault when we try to get cached copy of the dereferenced Datums.

Since we don't need access via the cache but only through `_ppvar` we don't need to make them part of the instance structure.


JCGoran:
Just for clarification, which test was failing here?

1uc:
This one for sure:
```
    82 - hoctests::test_mechfunc_py (Failed)
```
and likely these as well:
```
    91 - hoctests::test_setdata_py (Failed)
    145 - api::vclamp_cpp (Failed)
```

JCGoran (approved):
LGTM

PR #1546, opened 2024-10-30
Refactor variable name of random variables.

1uc:


src/codegen/codegen_neuron_cpp_visitor.hpp
```diff
@@ -130,7 +130,6 @@ class CodegenNeuronCppVisitor: public CodegenCppVisitor {
     /*                     Common helper routines accross codegen functions                 */
     /****************************************************************************************/
 
-
```
JCGoran, line 133: Nitpick: just whitespace change, can be reverted (unless formatter complains).

JCGoran (approved):

src/codegen/codegen_neuron_cpp_visitor.hpp
1uc: similarly, it can stay, because the formatter allows it :)

PR #1550, opened 2024-11-05
Hide `CodegenCppVisitor::setup`.

1uc:


test/unit/codegen/codegen_coreneuron_cpp_visitor.cpp
```diff
@@ -36,17 +36,18 @@ std::shared_ptr<CodegenCoreneuronCppVisitor> create_coreneuron_cpp_visitor(
     const std::shared_ptr<ast::Program>& ast,
     const std::string& /* text */,
     std::stringstream& ss) {
+    ImplicitArgumentVisitor().visit_program(*ast);
```
JCGoran, line 39: Why do we need this one and the `PerfVisitor` below (and why didn't we need it before)?

test/unit/codegen/codegen_coreneuron_cpp_visitor.cpp
```diff
@@ -343,17 +314,17 @@ SCENARIO("Check parameter constness with VERBATIM block",
...
                 struct IntervalFire_Instance  {
                     double* invl{};
                     const double* burst_start{};
                     double* v_unused{};
                     IntervalFire_Store* global{&IntervalFire_global};
-                };
-            )";
-            REQUIRE(reindent_text(generated) == reindent_text(expected_code));
+                };)";
+            expected_code = reindent_text(expected_code, 1);
```
JCGoran, line 326: It's slightly worrisome that we do this in the first place, i.e. using a string comparison to check the code...

test/unit/utils/test_utils.cpp
```diff
@@ -50,12 +50,13 @@ NEURON {
  * nmodl, the nmodl output is without "extra" whitespaces in the provided input.
  */
 
-std::string reindent_text(const std::string& text) {
+std::string reindent_text(const std::string& text, int indent_level) {
```
JCGoran, line 53: I take it this (and below) is part of https://github.com/BlueBrain/nmodl/pull/1549?

test/unit/codegen/codegen_coreneuron_cpp_visitor.cpp
1uc: off-topic.

test/unit/utils/test_utils.cpp
1uc: correct.

test/unit/codegen/codegen_coreneuron_cpp_visitor.cpp
1uc: We fused all the "create code" functions into a single one: run coreneuron codegen. Before there used to be variants, one for printing only the instance struct, etc. Those had different visitors applied to the ast. One of them was ImplicitArgumentVisitor.

The issue is that half the CodegenCoreneuronCppVisitor API became public for testing purposes only. To avoid that, we just always run the full thing and check if we find the substring. Since now the substring can be found indented, we also need the ability to shift snippets to the required indent level.

JCGoran: Understood, thanks for clarifying!

JCGoran (approved):

PR #1551, opened 2024-11-05
Create CVODE on demand.

1uc:
Use `codegen --cvode` to enable.

### DIFF
--- a/src/codegen/codegen_cpp_visitor.hpp
+++ b/src/codegen/codegen_cpp_visitor.hpp
@@ -266,6 +266,20 @@ class CodegenCppVisitor: public visitor::ConstAstVisitor {
         , float_type(std::move(float_type))
         , optimize_ionvar_copies(optimize_ionvar_copies) {}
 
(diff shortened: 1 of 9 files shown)


src/codegen/codegen_cpp_visitor.hpp
```diff
@@ -1481,11 +1495,10 @@ class CodegenCppVisitor: public visitor::ConstAstVisitor {
     std::string compute_method_name(BlockType type) const;
 
     // Automatically called as part of `visit_program`.
-    void setup(const ast::Program& node);
+    virtual void setup(const ast::Program& node);
```
JCGoran, line 1498: Why is this change necessary?

JCGoran (approved):

src/codegen/codegen_cpp_visitor.hpp
JCGoran: Okay, makes sense.

PR #1554, opened 2024-11-06
Fix CMake integration of NMODL generation.

1uc:
The code for the code-generator NMODL is itself generated from Jinja templates. This requires CMake integration to ensure it happens automatically.

The following has been changed about the integration:

  * The CMake code needs a list of the generated files. This list is stored in a CMake file and included. This file must exist before running `code_generator.py` to generate the NMODL code, because the command to generate the files needs to know which file it will be generating. Therefore, we split the script into two phases: one to generate the list of generated files, and another to generate the files.

  * The list of generated files depends on the build flags. Therefore, this file (`code_generator.cmake`) can't be checked into the Git repo. The `code_generator.cmake` is created during the configure phase, and now lives in the build directory.

  * CMake code was lifted from `src/language/CMakeLists.txt` to the top-level `CMakeLists.txt` to avoid any scoping issues for the lists of generated files.

  * Removed pattern of yielding tasks while creating a list tasks in favour of returning the list of tasks directly.


1uc:
Second attempt at fixing the issue. The previous was a rather short-lived "solution".

pramodk (approved):
LGTM this approach as well.

PR #1555, opened 2024-11-07
Fix race condition in CMake code.

1uc:
The race is cause by the fact that two targets exist to copy the Python files. The first one because of `cpp_cc_build_time_copy` without `NO_TARGET`. The second because of the:

    add_custom_target(nmodl_copy_python_files ALL DEPENDS ...)

The solution is to not create the per file target by passing `NO_TARGET` to `cpp_cc_build_time_copy`.


1uc:
I suspect this has been the cause for the very infrequent build failure I see when using `--parallel`.

alkino (approved):

PR #1561, opened 2024-12-02
Remove units when parsing CVODE statements

JCGoran:


JCGoran:
While I feel like this PR provides more of a workaround than a solution, implementing it the "right" way (i.e. directly removing the nodes from the AST and making sure that the units are scaled correctly) would take much longer, so I'm inclined to merge this as-is since it properly handles the majority of the cases.

